PR #41, opened 2015-03-20
YAML-specified SBP deserialization unit tests and Python test generator.

mookerji:
Defines unit test coverage for SBP messages. The idea here is that
this single unit test specification can be shared by any language
implementing an SBP client.

Cleaning up a bunch of stuff, don't merge yet.

/cc @mfine @fnoble @imh @denniszollo


mookerji:
I got 99 problems (and Github private forks are one): #52 replaces this PR....

PR #43, opened 2015-03-23
Type updates to SBP spec.

mookerji:
Updates from piksi_firmware HITL integration with libsbp C headers:
0. Carrier phase struct is actually s32 and not u32.

/cc @mfine


mookerji:
Replaced by #47.

PR #46, opened 2015-03-24
Example - Send SBP messages over UDP

fnoble:
An example of how we may want to incorporate examples using the python client library.

Adds two examples:
 - Simple example of reading and printing `BASELINE_NED` messages
 - Sending SBP over UDP

In this case I needed something to send SBP over UDB so it can be fed into MAVProxy and eventually make its way into the telemetry stream up to a ArduPilot quadcopter (so we can use just one pair of radios for corrections and telemetry).

### DIFF
--- a/python/sbp/client/examples/udp.py
+++ b/python/sbp/client/examples/udp.py
@@ -1,0 +1,78 @@
+# Copyright (C) 2015 Swift Navigation Inc.
+# Contact: Fergus Noble <fergus@swiftnav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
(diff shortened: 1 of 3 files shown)


python/sbp/client/examples/udp.py
```diff
@@ -0,0 +1,78 @@
...
+    s += struct.pack("<H", msg.crc)
+    udp.sendto(s, (args.address[0], args.udp_port[0]))
+
+  return send_udp_callback
+
+def main():
+  args = get_args()
+
+  with PySerialDriver(args.serial_port[0], args.baud[0]) as driver:
+    with Handler(driver.read, driver.write, args.verbose) as handler:
```
njoubert: There is no args.verbose currently, so this crashes.

fnoble:
@mookerji @mark ready for review

fnoble:
wait, I'm not seeing my latest updates. I think its the same issue that @mookerji had.

fnoble:
Superseded by #53

PR #53, opened 2015-03-24
Examples for Python client library

fnoble:
An example of how we may want to incorporate examples using the python client library.

Adds two examples:

 - Simple example of reading and printing BASELINE_NED messages
 - Sending SBP over UDP

In this case I needed something to send SBP over UDB so it can be fed into MAVProxy and eventually make its way into the telemetry stream up to a ArduPilot quadcopter (so we can use just one pair of radios for corrections and telemetry).

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -16,6 +16,21 @@ pip install libsbp
 ```python
 ```
 
+## Examples
+
+### Sending SBP messages over UDP
+
+Receives SBP messages over a serial port and sends all incoming messages to a
+UDP socket.
+
+See the [`source code`](sbp/client/examples/udp.py)
+
+Run the example:
+
+```shell
+$ python -m sbp.client.examples.udp -s /path/to/serial/port
+```
+
 ## LICENSE
 
 Copyright © 2015 Swift Navigation
--- a/python/README.md
+++ b/python/README.md
@@ -18,6 +18,19 @@ pip install libsbp
 
 ## Examples
 
+### Simple example
+
+Receives SBP messages over a serial port, decodes `MSG_BASELINE` messages and
+prints them out.
+
+See the [`source code`](sbp/client/examples/simple.py)
+
+Run the example:
+
+```shell
+$ python -m sbp.client.examples.simple -p /path/to/serial/port
+```
+
 ### Sending SBP messages over UDP
 
 Receives SBP messages over a serial port and sends all incoming messages to a
--- a/python/README.md
+++ b/python/README.md
@@ -23,9 +23,9 @@ pip install libsbp
 Receives SBP messages over a serial port, decodes `MSG_BASELINE` messages and
 prints them out.
 
-See the [`source code`](sbp/client/examples/simple.py)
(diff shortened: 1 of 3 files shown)


fnoble:
@mookerji @mfine : ok, now we are good to go!

python/sbp/client/examples/udp.py
```diff
@@ -0,0 +1,78 @@
...
+    udp.sendto(s, (args.address[0], args.udp_port[0]))
+
+  return send_udp_callback
+
+def main():
+  args = get_args()
+
+  with PySerialDriver(args.serial_port[0], args.baud[0]) as driver:
+    with Handler(driver.read, driver.write) as handler:
+      udp = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
```
mfine: Maybe throw the socket in a with statement?

```python
  with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as udp
```

... but nevermind, because python sockets don't implement context manager apparently. Might be good to have a close on the socket just to be proper.

mfine:
Looks good to me!

python/sbp/client/examples/udp.py
mfine: [SO thread](http://stackoverflow.com/questions/16772465/how-to-use-socket-in-python-as-a-context-manager)

mookerji:
Looking.

python/sbp/client/examples/udp.py
fnoble: Hehe, I tried that first too. I like this method mentioned in the SO thread:

```
from contextlib import closing
import socket

with closing(socket.socket()) as s:
    print s
```

sound good?

python/sbp/client/examples/simple.py
```diff
@@ -0,0 +1,64 @@
...
+
+  import argparse
+  parser = argparse.ArgumentParser(description="Swift Navigation SBP Example.")
+  parser.add_argument("-p", "--port",
+                      default=['/dev/ttyUSB0'], nargs=1,
+                      help="specify the serial port to use.")
+  args = parser.parse_args()
+
+  # Open a connection to Piksi using the default baud rate (1Mbaud)
+  with PySerialDriver(args.port[0], 1000000) as driver:
```
mookerji: nit: `baud_rate = 1000000`.

python/README.md
```diff
@@ -16,8 +16,36 @@ pip install libsbp
...
 ```
 
+## Examples
+
+### Simple example
+
+Receives SBP messages over a serial port, decodes `MSG_BASELINE` messages and
+prints them out.
+
+See the [`source code`](sbp/client/examples/simple.py)
```
mookerji: nit: period missing.

python/README.md
```diff
@@ -16,8 +16,36 @@ pip install libsbp
...
+
+Run the example:
```
mookerji: nit: "Run this example with: " or "To run the example: "

mookerji:
just a few nits. otherwise fine to merge.

PR #54, opened 2015-03-24
setup.py requirements update.

mookerji:
/cc @mfine

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -28,11 +28,11 @@ PLATFORMS = [
   'win32',
 ]
 
-# TODO
-INSTALL_REQUIRES = []
-
 with open('README.md') as f:
   readme = f.read()
+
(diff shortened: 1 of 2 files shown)


python/setup.py
```diff
@@ -28,8 +28,7 @@
   'win32',
 ]
 
-# TODO
-INSTALL_REQUIRES = []
+INSTALL_REQUIRES = [i.strip() for i in open("./requirements.txt").readlines()]
```
mfine: Maybe mirror the readme?

```python
with open('requirements.txt) as f:
  INSTALL_REQUIRES = [i.strip() for i in f.readlines()]
```

mookerji:
PYTHON HOW DOES IT WORK????

mfine:
:shipit:

PR #63, opened 2015-03-30
`sbp.client` changes for piksi_firmware

mfine:
These are changes needed to support piksi_firmware for https://github.com/swift-nav/piksi_firmware/issues/295.

/cc @mookerji

### DIFF
--- a/python/sbp/client/handler.py
+++ b/python/sbp/client/handler.py
@@ -166,7 +166,7 @@ class Handler(object):
   def add_callback(self, callback, msg_type=None):
     """
     Add per message type or global callback.
-    
+
     Parameters
     ----------
     callback : fn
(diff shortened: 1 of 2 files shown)


mookerji:
.... so now might be a good time to figure out a better thread safety story with the context manager.

mfine:
Not sure if the problem is going to surface with the context managers, as I'm not super familiar with how contention around that will look like - assume it will be something similar - not sure how you could have multiple threads exiting the same context manager - but are you thinking something like this?

https://github.com/swift-nav/libsbp/compare/master...mfine:mfine-threading-contexts

I dug up a lot and couldn't find any great examples of chaining contexts - is there a decorator pattern that will do this or something?

mfine:
See #65.

PR #65, opened 2015-03-31
Add locks to the contexts around resources. Also flushes.

mfine:
Not sure what this is going to buy us - how do threads escape another thread's context? Flushes are good addition here, plus the flushing on `__enter__` for drivers. Is there a better pattern to accomplish this? Should we be using `acquire`/`release` instead of `__enter__`/`__exit__`? Really want to enter a lock's context management inside these contexts, but don't know how to do that.

/cc @mookerji

### DIFF
--- a/python/sbp/client/drivers/base_driver.py
+++ b/python/sbp/client/drivers/base_driver.py
@@ -8,6 +8,8 @@
 # EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
+import threading
+
 class BaseDriver(object):
   """
   BaseDriver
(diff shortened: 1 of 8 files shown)


mookerji:
I may very well be wrong about this. I thought the original case that came up was contention between the serial link listener thread and the main thread, where it seems that could be a problem depending on who's holding the file handle. At the very least, it seems like you'd want to have a thread lock a resource, since these handles aren't threadsafe by default.

mookerji:
Perhaps the best use case I've seen: https://hg.python.org/cpython/file/2.7/Lib/logging/__init__.py

mfine:
My thought was that in the old `serial_link.py` there was no context management around the log file resource or the driver resource - so there was no explicit cleanup around leaving those contexts, as there weren't any contexts. With the new context managers, that's hopefully less of an issue - as you leave the contexts, you clean up resources (there's still a race with the thread finishing and writing to a socket while the driver file context is closing). Thinking more about it, don't think this will do much. Maybe as a read lock? And the acquiring around resource access as well? (flush, write, read, etc.) That should keep the resource on to one thread.

What do you think?

mfine:
Added reading locks around handle accesses.

PR #72, opened 2015-04-07
libsbp C header updates.

mookerji:
Regenerates the libsbp C headers following update of spec for
bootload, file_io, flash, etc.

/cc @mfine @fnoble


mfine:
lgtm!

PR #73, opened 2015-04-07
Instructions for and updates to Sphinx docs.

mookerji:
Adds some instructions for generating documentation and getting
doc-specific prereqs. Magical import of logging apparently prevents
name clashing between cpython.logging and sbp.logging.

/cc @mfine


python/docs/requirements.txt
```diff
@@ -0,0 +1,4 @@
+sphinx
+sphinxcontrib
```
mfine: had an issue with this install - all other depends installed fine:

```
 pip install sphinxcontrib
Collecting sphinxcontrib
  Could not find any downloads that satisfy the requirement sphinxcontrib
  No distributions at all found for sphinxcontrib
```

mookerji: right. updated. can you let me know if that gets the docs building on your machine?

mfine:
Built for me and they look beautiful!

PR #80, opened 2015-04-09
New message MSG_EXT_EVENT for reporting timestamped external pin transitions

henryhallam:
Untested at the moment, since I can't get the generator to work.
@denniszollo @fnoble thoughts on the message structure appreciated.  Also, the message ID - does it make sense to make this independently maskable, or should it just be in the same bunch as e.g. the navigation messages?

What should the behavior be if we detect an event but don't yet have a navigation solution and hence can't convert to GPS time?  Perhaps output the time as all zeros, and/or use a flag bit?  Alternately, just don't output messages in that situation.

We should potentially also add an option to generate and output an interpolated RTK position, but I think that would be a separate message.

### DIFF
--- a/spec/yaml/swiftnav/sbp/ext_events.yaml
+++ b/spec/yaml/swiftnav/sbp/ext_events.yaml
@@ -43,8 +43,13 @@ definitions:
             type: u8
             desc: Flags
             fields:
-              - 1-7:
+              - 2-7:
                   desc: Reserved
+              - 1:
+                  desc: Time quality
+                  values:
+                      - 0: Unknown - don't have nav solution
+                      - 1: Good (< 1 microsecond)
               - 0:
                   desc: New level of pin
                   values:
--- a/spec/yaml/swiftnav/sbp/ext_events.yaml
+++ b/spec/yaml/swiftnav/sbp/ext_events.yaml
@@ -1,0 +1,48 @@
+# Copyright (C) 2015 Swift Navigation Inc.
+# Contact: Fergus Noble <fergus@swiftnav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+package: swiftnav.sbp.ext_events
+description: |
+  Messages reporting accurately-timestamped external events,
+  e.g. camera shutter time
+
+stable: False
+public: True
+include:
+  - types.yaml
+definitions:
+
+ - MSG_EXT_EVENT:
+    id: 0x0300
+    public: True
+    short_desc: External event
+    desc: |
+        Reports detection of an external event, the GPS time it occurred,
+        which pin it was and whether it was rising or falling.
+    fields:
+        - wn:
+            type: u16
+            units: weeks
+            desc: GPS week number
+        - tow:
+            type: u32
+            units: ms
+            desc: GPS Time of Week rounded to the nearest ms
+        - ns:
+            type: s32
+            units: ns
+            desc: Nanosecond remainder of rounded tow
+        - pin_edge:
+            type: u8
+            desc: |
+                Bit 7 (MSB) denotes the new level of the pin, i.e. 1 for
+                rising edge, 0 for falling edge.
+                Bits 6:4 reserved.
+                Bits 3:0 denote pin number, i.e. 0x0 for DEBUG0.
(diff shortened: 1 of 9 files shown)


spec/yaml/swiftnav/sbp/ext_events.yaml
```diff
@@ -0,0 +1,48 @@
...
+            desc: GPS week number
+        - tow:
+            type: u32
+            units: ms
+            desc: GPS Time of Week rounded to the nearest ms
+        - ns:
+            type: s32
+            units: ns
+            desc: Nanosecond remainder of rounded tow
+        - pin_edge:
```
fnoble: rename to `flags` as this now contains more than just the pin edge?

spec/yaml/swiftnav/sbp/ext_events.yaml
```diff
@@ -0,0 +1,48 @@
...
+            type: u8
```
fnoble: Consider making this a `u16` to give more room for future expansion?

spec/yaml/swiftnav/sbp/ext_events.yaml
```diff
@@ -0,0 +1,48 @@
+# Copyright (C) 2015 Swift Navigation Inc.
+# Contact: Fergus Noble <fergus@swiftnav.com>
```
fnoble: I didn't write this! :)

spec/yaml/swiftnav/sbp/ext_events.yaml
```diff
@@ -0,0 +1,48 @@
...
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+package: swiftnav.sbp.ext_events
```
fnoble: Perhaps we should make a package called `time`, and move `MSG_GPS_TIME` into there too?

fnoble:
When the change is ready to be merged you should make sure to commit the generated output too.

spec/yaml/swiftnav/sbp/ext_events.yaml
```diff
@@ -0,0 +1,48 @@
...
+            units: ms
+            desc: GPS Time of Week rounded to the nearest ms
+        - ns:
+            type: s32
+            units: ns
+            desc: Nanosecond remainder of rounded tow
+        - pin_edge:
+            type: u8
+            desc: |
+                Bit 7 (MSB) denotes the new level of the pin, i.e. 1 for
```
mookerji: There's a "syntax" for denoting flags/bitfields that might be useful here.

spec/yaml/swiftnav/sbp/ext_events.yaml
```diff
@@ -0,0 +1,48 @@
...
+stable: False
+public: True
+include:
+  - types.yaml
+definitions:
+
+ - MSG_EXT_EVENT:
+    id: 0x0300
+    public: True
+    short_desc: External event
```
mookerji: "Reports timestamped, external event"?

spec/yaml/swiftnav/sbp/ext_events.yaml
mookerji: That's a good idea, but we may want to discuss it briefly.

spec/yaml/swiftnav/sbp/ext_events.yaml
henryhallam: it's the pin and the edge, but sure.. although if there's bitmask syntax that would be better

henryhallam:
@fnoble Now with separate flags and pin number fields.  How do you feel about 0x0300 as message ID?

mookerji:
Now would be a good time to play test this new Makefile that got merged.

fnoble:
@mookerji : I'll let you manage merging this in when you are ready. Do you want to go ahead with changing the package name to `time`?

spec/yaml/swiftnav/sbp/ext_events.yaml
```diff
@@ -0,0 +1,61 @@
...
+  e.g. camera shutter time
+
+stable: False
+public: True
+include:
+  - types.yaml
+definitions:
+
+ - MSG_EXT_EVENT:
+    id: 0x0300
```
fnoble: Actually - how about making this part of the same grouping as the time message? e.g. `0x0101`? (`GPS_TIME` is `0x0100`)

henryhallam: I feel like the relationship between ext_events and time isn't really any stronger than e.g. the relationship between navigation and time.  But I don't really care or think it's terribly important.

henryhallam:
Added test, rebased on latest master.  Having trouble building the docs.

henryhallam:
retest this please

henryhallam:
rebased

mfine:
LGTM

henryhallam:
OK I'm going to merge this just for sanity... 'make test' still doesn't work, but I don't think it's related.

PR #82, opened 2015-04-09
Baby's first Makefile.

### LINKED ISSUE #77: Regenerate everything script
Would be nice if there was one script you could run at the top level after changing the YAML that built / generated all the things.

mookerji:
Closes #77.

/cc @gsmcmullin @mfine @denniszollo


Makefile
```diff
@@ -0,0 +1,34 @@
...
+all: c python docs
+
+c:
+	@echo "Generating C headers..."
+	$(SBP_GEN_BIN) -i $(SBP_SPEC_DIR) \
+		       -o $(SWIFTNAV_ROOT)/c/include/libsbp \
+	               --c;
+
+python:
+	@echo "Generating Python bindings..."
```
mfine: The first rule is the default target (if you just type make) - put the all rule first? The echo for all will be *after* the c and python and docs have been gen'd.

Add some [PHONY targets](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html)?

```make
.PHONY: all c python docs
```

Makefile
```diff
@@ -0,0 +1,34 @@
+# Convenience Makefile for generating and releasing SBP client
+# libraries.
+
+SWIFTNAV_ROOT := $(shell pwd)
+MAKEFLAGS += SWIFTNAV_ROOT=$(SWIFTNAV_ROOT)
+SBP_SPEC_DIR := $(SWIFTNAV_ROOT)/spec/yaml/swiftnav/sbp/
+SBP_GEN_BIN := python $(SWIFTNAV_ROOT)/sbpg/generator.py
+
+help:
```
mfine: this might be the first ever make `usage`

mookerji:
Note that this depends on #71 .

PR #93, opened 2015-04-11
README update.

mookerji:
/cc @mfine @henryhallam

### DIFF
--- a/python/docs/source/index.rst
+++ b/python/docs/source/index.rst
@@ -10,6 +10,11 @@ transmit solutions, observations, status and debugging messages, as
 well as receive messages from the host operating system, such as
 differential corrections and the almanac.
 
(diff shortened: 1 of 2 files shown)


python/docs/source/index.rst
```diff
@@ -10,6 +10,10 @@ transmit solutions, observations, status and debugging messages, as
 well as receive messages from the host operating system, such as
 differential corrections and the almanac.
 
+Please see [the message
```
henryhallam: Is this correct syntax for a hyperlink in RST?
http://sphinx-doc.org/rest.html#hyperlinks says not, but I guess what matters is how it comes out after compilation.

mookerji: Yep, i saw that earlier and just pushed an update.

PR #96, opened 2015-04-13
Python client updates: materialization and tests.

mookerji:
0. Updates serializations tests, which got dropped with the last
message migration.
1. Optionally, materialize SBP objects from fields. Previously an SBP
object was required, since we were only ever deserializing.
2. Docstring clarifications.

/cc @mfine


mfine:
Looks good.

PR #101, opened 2015-04-14
WIP JSON serialization

mfine:
Playing around with JSON serialization. This is pretty broken and not working:

+ the parent objects are getting lost or not able to be serialized (e.g., `payload`); what really needs to happen is that the json dict from the parent needs to be merged in with the child's json dict. Or some such.
+ the `from_json` approach here is no bueno with nested structures - `json.loads` returns dicts and lists, while the existing objects are built up from construct into containers. Need some kind of json-to-construct mapping function.

Thoughts and ideas on how to do a better job here very welcome!

DO NOT MERGE

/cc @mookerji

### DIFF
--- a/generator/sbpg/targets/resources/sbp_construct_template.py.j2
+++ b/generator/sbpg/targets/resources/sbp_construct_template.py.j2
@@ -15,8 +15,9 @@
 """
 
 from construct import *
+import json
 from sbp import SBP
-from sbp.utils import fmt_repr, exclude_fields
+from sbp.utils import fmt_repr, exclude_fields, walk_json_dict
 import six
 
 ((*- for i in include *))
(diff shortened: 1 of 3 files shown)


mfine:
Another idea here would be to jettison the parent fields (filter them all out on the way out) and just compose the object from its own relevant fields. Still would have the issue with building containers for nested structures.

mfine:
Just checked that this serializes ok - downside is that it won't serialize any of the parents stuff, or deserialize it back. Plus, the issue with the containers for nested objects is still there.

generator/sbpg/targets/resources/sbp_construct_template.py.j2
```diff
@@ -144,6 +145,19 @@ class ((( m.identifier | classnameify )))(SBP):
     c = Container(**exclude_fields(self))
     self.payload = ((( m.identifier | classnameify )))._parser.build(c)
     return self.pack()
+
+  def to_json(self):
```
mookerji: With the `__dict__`, this doesn't get the parent objects?

mfine: It does (I had filtered them originally with `exclude_fields`).

mfine:
Replaced by https://github.com/swift-nav/libsbp/pull/103.

PR #105, opened 2015-04-14
handle unpickling error more gracefully

denniszollo:
This should address issue https://github.com/swift-nav/sbp_log_analysis/issues/84 .


denniszollo:
Retest this please

denniszollo:
I'm not sure where to catch this error anymore.  I think that if there is a "SBP payload deserialization error" we shouldn't try to unpicke or do anything.

mfine:
@denniszollo since we're off pickle now, going to close this out (JSON is parsed per line and should have better handling here).

PR #114, opened 2015-04-15
Misc Fixes

fnoble:
- Changes id for `MSG_EPHEMERIS` to `0x46` (#98)
 - Removes annoying timestamps from generated files
 - Fixes `MSG_BOOTLOADER_HANDSHAKE` type spec to fix terminiation issue (#113)
 - Removes the unsupported `bytes` type from the specification

@mookerji @mfine

PR #117, opened 2015-04-16
Bootload test fixes (and more!)

fnoble:
- Some generalization of the generator to properly handle arrays of primitive types
 - Fix to bootloader test now we have a new payload spec (from #114)

### DIFF
--- a/spec/tests/yaml/swiftnav/sbp/test_bootload.yaml
+++ b/spec/tests/yaml/swiftnav/sbp/test_bootload.yaml
@@ -6,11 +6,20 @@ tests:
 
 - msg:
     fields:
-      handshake: 118
+      handshake: [118, 49, 46, 50] # 'v1.2'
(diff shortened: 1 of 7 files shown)


fnoble:
@mfine @mookerji

fnoble:
retest this please

spec/tests/yaml/swiftnav/sbp/test_bootload.yaml
```diff
@@ -6,11 +6,20 @@ tests:
 
 - msg:
     fields:
-      handshake: 118
+      handshake: [118, 49, 46, 50] # 'v1.2'
```
mookerji: Hmmm, I suppose this is fine for the time being. It's more a problem that we aren't being consistent with our schema (null terminated strings). I'll probably follow up at some point with making this parser conditional.

PR #122, opened 2015-04-16
Catch message dispatch KeyErrors and make them warnings.

mookerji:
In the event that we're missing a message in the dispatch table, raise
a warning instead of throwing an exception.

/cc @fnoble @mfine


mfine:
Oh wow, didn't realize it was an exception. Ship it.

PR #123, opened 2015-04-16
Expand JSON messages before they go out

mfine:
JSON messages need to be inflated before they go out. Do a table dispatch on the way out.

/cc @mookerji


fnoble:
lgtm

PR #124, opened 2015-04-17
Make ext_events messages public

henryhallam:


mfine:
@henryhallam there's an updated test case I just pulled in that will fix that failure on `test_pickle_log_missing`.

PR #126, opened 2015-04-17
SBP doc: Table 2.0.1. s/Payload/Packet/

henryhallam:
Chapter 2 "Message Structure",  Table 2.0.1: Swift Binary Protocol message structure

The bottom row says "Total Payload Length" but it should read "Total Packet Length" because this table is describing a whole packet including the header and footer.

All the other tables in the doc describe payloads, so they correctly say "Total Payload Length"


fnoble:
Perhaps "Message" instead of "Packet" for consistency? We don't tend to refer to packets much.

PR #127, opened 2015-04-17
Remove the main from client - it's in piksi tools

mfine:
This was trying to replace `serial_link.py`, but that right exists in `piksi_tools` now. This isn't being used anymore, and don't think anything ever took a dependency on it. So let's kill it.

/cc @fnoble @mookerji


fnoble:
lgtm

PR #129, opened 2015-04-21
Suppress NaN in python JSON dump

### LINKED ISSUE #128: Disallow NaN in JSON
JSON doesn't support NaN, but the python library will generate NaN values by default. See [here](http://stackoverflow.com/questions/6601812/sending-nan-in-json) for more context.

The NaN values currently kill log ingestion. Fix is to `allow_nan=False`.

mfine:
Set `allow_nan=False` in `json.dumps`. Fixes #128 

/cc @fnoble @mookerji

### DIFF
--- a/python/sbp/client/loggers/json_logger.py
+++ b/python/sbp/client/loggers/json_logger.py
@@ -67,13 +67,17 @@ class JSONLogIterator(LogIterator):
 
     """
     for line in self.handle:
-      data = json.loads(line)
-      delta = data['delta']
-      timestamp = data['timestamp']
-      item = SBP.from_json_dict(data['data'])
       try:
-        yield (delta, timestamp, self.dispatcher(item))
-      except KeyError:
-        yield (delta, timestamp, item)
+        data = json.loads(line)
+        delta = data['delta']
+        timestamp = data['timestamp']
+        item = SBP.from_json_dict(data['data'])
+        try:
+          yield (delta, timestamp, self.dispatcher(item))
+        except KeyError:
+          yield (delta, timestamp, item)
+      except ValueError:
+        warn = "Bad JSON decoding for line %s" % line
+        warnings.warn(warn, RuntimeWarning)
     self.handle.seek(0, 0)
     raise StopIteration
--- a/python/sbp/client/loggers/json_logger.py
+++ b/python/sbp/client/loggers/json_logger.py
@@ -11,6 +11,7 @@
 from ... import SBP
 from .base_logger import BaseLogger, LogIterator
 import json
+import warnings
 
 class JSONLogger(BaseLogger):
   """
@@ -31,8 +32,12 @@ class JSONLogger(BaseLogger):
             "data": data}
 
   def call(self, msg):
-    self.handle.write(json.dumps(self.fmt_msg(msg), allow_nan=False) + "\n")
-
+    try:
+      self.handle.write(json.dumps(self.fmt_msg(msg), allow_nan=False) + "\n")
+    except ValueError:
+      warn = "Bad values in JSON encoding for msg_type %d for msg %s" \
+             % (msg.msg_type, msg)
+      warnings.warn(warn, RuntimeWarning)
 
 class JSONLogIterator(LogIterator):
   """
--- a/python/sbp/client/loggers/json_logger.py
+++ b/python/sbp/client/loggers/json_logger.py
@@ -31,7 +31,7 @@ class JSONLogger(BaseLogger):
             "data": data}
 
   def call(self, msg):
-    self.handle.write(json.dumps(self.fmt_msg(msg)) + "\n")
+    self.handle.write(json.dumps(self.fmt_msg(msg), allow_nan=False) + "\n")
 
 
 class JSONLogIterator(LogIterator):
--- a/python/sbp/__init__.py
+++ b/python/sbp/__init__.py
@@ -117,7 +117,7 @@ class SBP(object):
 
     """
     d = self.to_json_dict()
-    return json.dumps(d, allow_nan=False)
+    return json.dumps(d)
 
   @staticmethod
   def from_json(s):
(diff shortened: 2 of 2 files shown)


fnoble:
Looks like this will cause a `ValueError` if we get an SBP message containing a NaN. Maybe there is a way to handle more gracefully?

mfine:
Good catch:

```python
>>> json.dumps({"wut": float('nan')})
'{"wut": NaN}'
>>> json.dumps({"wut": float('nan')}, allow_nan=False)
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 250, in dumps
    sort_keys=sort_keys, **kw).encode(obj)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
ValueError: Out of range float values are not JSON compliant
>>> 
```

Not sure what the graceful solution is here, though.

fnoble:
Dunno what the right answer is, perhaps just exclude it from the output?

mfine:
K, will swallow the `ValueError` and move on.

fnoble:
Maybe warn along the way?

mookerji:
How and why are we getting NaN's with the libsbp objects?

mfine:
@mookerji seeing it in tracking messages:

```json
{"timestamp": 1429577577, "data": {"sender": 863, "msg_type": 22, "states": [{"state": 1, "prn": 23, "cn0": 0.9774436354637146}, {"state": 1, "prn": 5, "cn0": 1.1590908765792847}, {"state": 1, "prn": 8, "cn0": 0.9618320465087891}, {"state": 1, "prn": 18, "cn0": 1.0597014427185059}, {"state": 1, "prn": 16, "cn0": 1.0}, {"state": 1, "prn": 19, "cn0": 0.7814569473266602}, {"state": 1, "prn": 29, "cn0": 1.7236841917037964}, {"state": 1, "prn": 6, "cn0": NaN}, {"state": 0, "prn": 0, "cn0": -1.0}, {"state": 0, "prn": 0, "cn0": -1.0}, {"state": 0, "prn": 0, "cn0": -1.0}], "crc": 31229, "length": 66, "preamble": 85, "payload": "ARe/OXo/AQUXXZQ/AQigOnY/ARJMpIc/ARAAAIA/AROQDUg/AR2vodw/AQYAAMB/AAAAAIC/AAAAAIC/AAAAAIC/"}, "delta": 203794}
```

fnoble:
NaN is a valid thing to have in a IEEE-754 float and hence is a valid thing to have in many SBP messages. Thats not to say that the firmware should be outputting NaNs, but it is valid on the wire.

mookerji:
Yeah, a `warn` or something would be good. I'm a bit worried about the firmware NaNs issue. Can you send me a log where this happens and maybe file an issue?

mfine:
Pull request updated with warning, issue opened on piksi_firmware https://github.com/swift-nav/piksi_firmware/issues/378.

mfine:
Catch bad ingestion of JSON, as well - https://github.com/swift-nav/sbp_log_analysis/issues/107.

python/sbp/client/loggers/json_logger.py
```diff
@@ -62,13 +67,17 @@ def next(self):
...
-        yield (delta, timestamp, item)
+        data = json.loads(line)
+        delta = data['delta']
+        timestamp = data['timestamp']
+        item = SBP.from_json_dict(data['data'])
+        try:
+          yield (delta, timestamp, self.dispatcher(item))
+        except KeyError:
+          yield (delta, timestamp, item)
+      except ValueError:
```
mookerji: Do we want to have different handling based on the kind of `ValueError` we have? Two cases it seems: 

```python
# 1. scan error implies end of file, so exit. 
ValueError: Unterminated string starting at:                                
# 2. value error implies bad entry, but the file may not be over yet? 
ValueError: Out of range float values are not JSON compliant   
```  

For 2, does the error file originating the issue continue past that point, or is it another manifestation of an unwanted EOF? There's a case to be made for exiting early, because it implies something serious about the data coming out. I suspect we'd want to be a bit more robust here. If we passed back an SBP object using just the SBP fields, a downstream analysis test could count these issues?

mfine: I think we want to issue the warn and move on to the next line. We're doing line-based processing, so it's feasible that one-line could go bad without impacting the other lines (for whatever reason) - we should be robust to that. If we get a `ValueError` here, we have nothing to yield back to iterator - there are no values we can draw from the line. I don't think we need to differentiate here between the errors unless further experience suggests otherwise - bad JSON value or incomplete line, the line output should give us enough information. My thoughts anyway.

mookerji: The thought as occurred to me that this is actually some contradictory feedback from earlier (sorry about that).

mookerji: Does it yield? After you catch the `ValueError`, I think it exits the `except` and then seeks back to the beginning of the file.

mfine: The indenting of the `ValueError` should be inside `for line in self.handle`:

```python
    for line in self.handle:
      try:
        data = json.loads(line)
        delta = data['delta']
        timestamp = data['timestamp']
        item = SBP.from_json_dict(data['data'])
        try:
          yield (delta, timestamp, self.dispatcher(item))
        except KeyError:
          yield (delta, timestamp, item)
      except ValueError:
        warn = "Bad JSON decoding for line %s" % line
        warnings.warn(warn, RuntimeWarning)
```

Should just go to the next line, right?

mookerji: I'm fucking retarded (I missed the second indent). Carry on.

PR #131, opened 2015-04-22
Removes unused message template.

mookerji:
/cc @mfine


mfine:
Template is now `sbp_construct_template.py.j2`. Ship it!

PR #132, opened 2015-04-23
Add file mode as a parameter - have log iterator take a filename

mfine:
Have the JSON logger take file mode as a parameter - we'd like to be able to append to a log file, as well. Some other changes here:

+ Reduce file permissions. From advice on the internet, got rid of the unnecessary `w+` and `r+` permissions. [1]
+ Have LogIterator take a filename that it opens. I don't know what I was thinking when I had this take a handle. This will break log analysis, but I'll clean those up. Think it's worth it to correct this.

/cc @mookerji 

[1] Here's a tip: you almost never want to use w+ or a+. The first clobbers your file before you can read it, and the second one is confusing because your read pointer can be anywhere, but on many systems, the writer always jumps to the end of the file.


mfine:
Added in support for iterating over JSON logs in multiple files in `MultiJSONLogIterator` - it parses and collects all the records in a list and then sorts the list on the timestamp (this requires the sort to be stable to preserve ordering, which according to documentation should be).

Also added `MultiJSONLogIterator.s3_handles(aws_access_key, aws_secret_key, bucket, keys)`, which returns string handles for uncompressed S3 objects - it collects all the objects matching bucket and list of key prefixes, uncompresses them, and returns filehandles. Useage looks like:

```python
handles = MultiJSONLogIterator.s3_handles('XXX', 'YYY', 'piksi-logs', ['devices/789/logs/2015042119', 'devices/789/logs/20150421'])
with MultiJSONLogIterator(handles) as log:
  for delta, timestamp, msg in log.next():
    print msg

```


No cool lazy reading and uncompressing and sorting of remote data from S3 here :(

python/requirements.txt
```diff
@@ -1,4 +1,5 @@
 six
+boto==2.38.0
```
mookerji: Errrr? You sure about that? 

We could also have a `requirements-dev.txt`, if we want to expand the scope of requirements.

mfine: ha, `noboto`?

python/sbp/client/loggers/json_logger.py
```diff
@@ -81,3 +94,89 @@ def next(self):
...
+      S3 bucket to read objects from.
+    keys : [string]
+      List of keys in bucket to iterate over.
+
+    """
+    handles = []
+    bucket = S3Connection(aws_access_key, aws_secret_key).get_bucket(bucket)
+    for key in keys:
+      for k in bucket.list(prefix=key):
+        s = k.get_contents_as_string()
```
mookerji: Any particular reason to read this into memory instead of something like `get_contents_to_file`?

python/sbp/client/loggers/json_logger.py
```diff
@@ -9,9 +9,14 @@
...
 
 from ... import SBP
+from ...table import dispatch
 from .base_logger import BaseLogger, LogIterator
 import json
 import warnings
+from boto.s3.connection import S3Connection
+from gzip import GzipFile
+from operator import itemgetter
+from StringIO import StringIO
```
mookerji: cStringIO?

python/tests/sbp/client/test_handler.py
```diff
@@ -69,6 +70,7 @@ def until(p, limit=1000):
   for i in itertools.count():
     if p():
       break
+    time.sleep(0)
```
mookerji: Cute. Very cute.

python/sbp/client/loggers/json_logger.py
mfine: Went with in memory and avoided disk - not sure how much would get out of using disk - still need to unzip the file. But maybe performs better? Wish could lazily read here because there's a big startup cost here to get everything downloaded and in order :(

mookerji: ... so I think you could probably refactor this to use `yield`, or use `concurrent.futures` to defer this to a thread pool.

I see the point about memory. The only thought that I had was that loading a (potentially) lot of matching keys into a `StringIO` could take a lot of memory and Python's not that memory efficient. `mmap`?

mfine: Oh yeah, this is far from ideal - more like a first pass at a reference implementation. There's lot of possible optimizations here - parallel downloads of keys, partial reading of files, mmap'ing in files... Not sure how applicable it will be initially with logs hopefully being in just a single key, but something to keep an eye on long term. But there are a lot of dragons around introducing optimizations here, and wanted to get something correct first.

mfine:
Going to bring this in. The approach around S3 objects is not ideal, but should be a correct starting point to improve on.

PR #133, opened 2015-04-28
Crc namespace collision

denniszollo:
The "crc" term in the ephemeris message was colliding with the crc value from the inherited class.  This change should avoid the namespace collision by adding an "_" character to each of the amplitude fields.  It will break the following:
 -- piksi_firmware (when updated to latest libsbp)
 -- generated documentation

### DIFF
--- a/c/include/libsbp/observation.h
+++ b/c/include/libsbp/observation.h
@@ -124,12 +124,12 @@ typedef struct __attribute__((packed)) {
 #define SBP_MSG_EPHEMERIS 0x0046
 typedef struct __attribute__((packed)) {
   double tgd;         /**< Group delay differential between L1 and L2 [s] */
-  double crs;         /**< Amplitude of the sine harmonic correction term to the orbit radius [m] */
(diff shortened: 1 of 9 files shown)


mookerji:
Ah, ok. Would you mind adding adding a few test cases for the ephemeris in the YAML test data? 

I can show you an example of how to do that later today: there's a utility in the Python tests directory. The command for doing that is run from the Python client:
```shell
PYTHONPATH=. python -m tests.sbp.build_test_data -l data/serial_link_log_20150402-100519.log --version 0.29
```

denniszollo:
I think we should update the tool to use JSON log iterator as well

denniszollo:
@mfine @mookerji @cbeighley 
This is ready for review if you anyone has a second.

mookerji:
Sweet!

mookerji:
One second for another look over and I'll go ahead and merge it.

fnoble:
lgtm, there may be some follow up needed in other modules (piksi_firmware, sbp_log_analysis) as a result of these changes however.

denniszollo:
I am creating an issue in piksi firmware

PR #137, opened 2015-04-29
Export PYTHONPATH from the Makefile.

gsmcmullin:
Fixes this non-descriptive problem when running make:
```
  File "sbpg/generator.py", line 20, in <module>
    import sbpg.specs.yaml2 as yaml
ImportError: No module named sbpg.specs.yaml2
```


fnoble:
I think there is a better way to fix this, I think its a symptom of not using packages correctly. Will take a look.

gsmcmullin:
I'm merging this since nobody's offered a better solution.

PR #139, opened 2015-04-29
Piksi: Add mask satellite message.

gsmcmullin:
I can't rebuild the latex documentation, but this fails on master too.

### DIFF
--- a/spec/yaml/swiftnav/sbp/piksi.yaml
+++ b/spec/yaml/swiftnav/sbp/piksi.yaml
@@ -221,16 +221,18 @@ definitions:
             type: u8
             desc: Mask of systems that should ignore this satellite.
             fields:
-                - 0-1:
-                    desc: Acquisition channel
-                    values:
-                        - 0: Enabled
-                        - 1: Skip this satellite on future acquisitions
-                - 1-2:
+                - 2-7:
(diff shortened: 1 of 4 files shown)


spec/yaml/swiftnav/sbp/piksi.yaml
```diff
@@ -209,3 +209,29 @@ definitions:
...
+ - MSG_MASK_SATELLITE:
+    id: 0x001B
+    short_desc: Mask a satellite from use in Piksi subsystems
+    desc: |
+      This message allows setting a mask to prevent a particular satellite
+      from being used in various Piksi subsystems.
+    fields:
+        - mask:
+            type: u8
+            desc: Mask of systems that should ignore this satellite.
```
mookerji: Wouldn't you actually want this to be something like:
```
        - mask:
            type: u8
            desc: Mask of systems that should ignore this satellite.
            fields:
                - 0:
                    desc: Acquisition channel
                    values:
                        - 0: Enabled
                        - 1: Skip this satellite on future acquisitions
                - 1:
                    desc: Tracking channels
                    values:
                        - 0: Enabled
                        - 1: Drop this PRN if currently tracking
```

cbeighley:
lgtm

PR #140, opened 2015-04-30
Add an example byte logger.

mfine:
Quick and dirty example byte logger.

/cc @henryhallam

### DIFF
--- a/python/sbp/client/examples/bytes.py
+++ b/python/sbp/client/examples/bytes.py
@@ -1,0 +1,53 @@
+# Copyright (C) 2015 Swift Navigation Inc.
+# Contact: Mark Fine <mark@swiftnav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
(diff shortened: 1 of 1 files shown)


python/sbp/client/examples/bytes.py
```diff
@@ -0,0 +1,53 @@
...
+DEFAULT_SERIAL_BAUD  = 1000000
+DEFAULT_LOG_FILENAME = time.strftime("sbp-%Y%m%d-%H%M%S.log")
+
+
+def get_args():
+  """
+  Get and parse arguments.
+  """
+  import argparse
+  parser = argparse.ArgumentParser(description="Swift Navigation SBP Bytes Example.")
```
henryhallam: Maybe clarify what the program does a little more: "Parses an SBP stream from serial and writes it to a file in the same binary format."

henryhallam:
looks great, thanks

PR #141, opened 2015-05-02
MAVLINK Bridge

njoubert:
This adds a UDP client that sends only observations and base position messages over UDP.

### DIFF
--- a/python/sbp/client/examples/mavlinkbridge.py
+++ b/python/sbp/client/examples/mavlinkbridge.py
@@ -1,0 +1,84 @@
+# Copyright (C) 2015 Swift Navigation Inc.
+# Contact: Fergus Noble <fergus@swiftnav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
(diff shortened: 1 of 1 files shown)


fnoble:
Looks good but I wonder if this is the best place for this to live. Not sure where would be better, maybe `piksi_tools`?

python/sbp/client/examples/mavlinkbridge.py
```diff
@@ -0,0 +1,84 @@
...
+                      help="specify the baud rate to use.")
+  parser.add_argument("-a", "--address",
+                      default=[DEFAULT_UDP_ADDRESS], nargs=1,
+                      help="specify the serial port to use.")
+  parser.add_argument("-p", "--udp-port",
+                      default=[DEFAULT_UDP_PORT], nargs=1,
+                      help="specify the baud rate to use.")
+  return parser.parse_args()
+
+def send_udp_callback_generator(udp, args):
```
mookerji: I believe there's is a packing function that does this for you in the library.

mookerji:
Sweet! I echo @fnoble 's suggestion. This should probably be in piksi_tools. If you could also add a test of some kind, that would be great.

python/sbp/client/examples/mavlinkbridge.py
njoubert: This is just a copy of Fergus' original code, can you link me to an example of what you're referring to?

mookerji: Yeah, see: 
https://github.com/swift-nav/libsbp/blob/master/python/sbp/__init__.py#L103

mookerji:
Closed in favor of https://github.com/swift-nav/piksi_tools/pull/74/.

PR #144, opened 2015-05-06
Iode to ephemeris

denniszollo:
We were missing a field in the SBP definition of the ephemeris


spec/tests/yaml/swiftnav/sbp/test_observation.yaml
```diff
@@ -349,305 +349,5 @@ tests:
     payload: KKNEGC4HEIRr2XkOlP//ASyvkwCBQsh9lB3+/5kpDrEDgLL4iI9C/v/eEriFDZ41soYqGwIA7R4hYBYmU1Z2qG///y040Fge
     preamble: '0x55'
     sender: '0x4c3'
```
mookerji: What happened to the ephemeris unit test case here?

mookerji:
Question about the tests. I would also check with @fnoble .

spec/tests/yaml/swiftnav/sbp/test_observation.yaml
denniszollo: oh weird.  the log I used to generate the unit test must have had only one type of ephemeris message in it.  I'll regenerate the unit tests with a log with both types of ephemeris.  Thanks for pointing that out.

fnoble:
lgtm once unit tests are restored

denniszollo:
Ephemeris tests back in with the correct field.  The old ephemeris message is checked in the new "deprecated" yaml per @mfine's updates.  I think this is ready to go.

PR #145, opened 2015-05-07
Allow the logger to inject tags into the logs

mfine:
Add a tags parameter to Base logger and use it in JSON logger. It's a string of tags separated by commas used to decorate the logs with. This results in something like:

```
tags=WUT=now,HI=hater
```

generates:

```json
{"timestamp": 1431022699, "WUT": "now", "HI": "hater", "data": ...
```

This will be used to inject session data into the logs, namely `session_id`. There's some questions here on whether to nest these tags and metadata - we can update that as necessary.

/cc @mookerji

### DIFF
--- a/python/sbp/client/loggers/json_logger.py
+++ b/python/sbp/client/loggers/json_logger.py
@@ -33,9 +33,10 @@ class JSONLogger(BaseLogger):
       data = self.dispatcher(msg).to_json_dict()
     except KeyError:
       data = msg.to_json_dict()
-    msg = {"delta": self.delta(), "timestamp": self.timestamp(), "data": data}
-    msg.update(self.tags)
-    return msg
+    return {"delta": self.delta(),
+            "timestamp": self.timestamp(),
+            "data": data,
+            "metadata": self.tags}
 
   def call(self, msg):
     try:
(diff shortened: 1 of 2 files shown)


python/sbp/client/loggers/json_logger.py
```diff
@@ -33,9 +33,9 @@ def fmt_msg(self, msg):
       data = self.dispatcher(msg).to_json_dict()
     except KeyError:
       data = msg.to_json_dict()
-    return {"delta": self.delta(),
-            "timestamp": self.timestamp(),
-            "data": data}
+    msg = {"delta": self.delta(), "timestamp": self.timestamp(), "data": data}
```
mookerji: Hmmm, I think there's a risk here of tags stamping over some existing keys (`delta`, `timestamp`, `data`).

mookerji: Probably having something like 'tags'?

mfine: Yeah, I actually reversed it - but then thought that might be desirable to overwrite. But probably not.

mfine: Added a `metadata` field, and always populate it, even if empty. `tags` is a little arbitrary - `metadata` fits nicer with `data`. Potentially `header` could be used here, too.

PR #146, opened 2015-05-08
Catch unhandled exceptions, log them, continue

mfine:
We crash hard when the thread goes down. Maybe this would help?

/cc @denniszollo @mookerji


mookerji:
Hmmm.... I'm going to give this a brief thought (seems like there are a few options) and get back to it tomorrow morning.

PR #151, opened 2015-05-26
Proposed new bootloader messages

mfine:
Capturing the changes described here:

+ Add new bootloader handshake message for Host -> Piksi (currently the same for Call/Response)
+ Change FLASH_PROGRAM (which is Host -> Piksi) to different ID (currently the same as FLASH_DONE (which is Piksi -> Host))
+ Add new FLASH_READ message for Host -> Piksi (currently the same for Call/Response)
+ Add new DEVICE_DNA message for Host -> Piksi (currently the same for Call/Response)
+ Add new STM_UNIQUE_ID message for Host -> Piksi (currently the same for Call/Response)

Simple pattern - leave old message in place and relabel with an appended `_OLD`; add new `_HOST` and `_DEVICE` appendages where applicable. `_DEVICE` and `_OLD` have same message id, where applicable.

Descriptions and other changes (tests, etc.) not updated yet. Trying to collect feedback about the approach here.

/cc @cbeighley


cbeighley:
lgtm

mfine:
@mookerji @fnoble any comments on the approach here? I'll update the documentation and specs and code if things look ok.

fnoble:
I need a bit longer to review if thats ok

mfine:
@fnoble definitely - didn't mean to imply any rush or haste.

fnoble:
Initial thoughts - how do we want to convey this in the documentation? Should we set `public: False`? Maybe we should add a note to the description about being deprecated?

Also, the new message descriptions should be updated to be more clear about how they are a request or a response, and if they are a request, which message to expect in reply.

Also do you think `_REQUEST` and `_RESPONSE` would make the semantics more clear than `_HOST` and `_DEVICE`?

mfine:
Added deprecation messages. Added updated host-, device- specific documentation. (Please review it - I might have gotten patterns wrong / potentially could use better descriptions).

I added request and response in the descriptions. Don't have much opinion about host/device vs. request/response. A possible thought I had is that if you would like to limit req/resp patterns in the future, might be easier not to use the names in the messages.

I also pulled out the payloads on the `_HOST` messages where applicable.

spec/yaml/swiftnav/sbp/bootload.yaml
```diff
@@ -22,15 +22,40 @@ include:
   - types.yaml
 definitions:
```
mookerji: I may be wrong, but we currently don't have an assert about message IDs not overlapping in the spec. Now that we're making these migrations, it may be some easy relief to add that.

spec/yaml/swiftnav/sbp/bootload.yaml
```diff
@@ -22,15 +22,40 @@ include:
...
+      - handshake:
+          type: array
+          fill: u8
+          desc: Version number string (not NULL terminated)
+
+ - MSG_BOOTLOADER_HANDSHAKE_OLD:
+    id: 0x00B0
+    public: False
+    short_desc: |
+      Deprecated, use MSG_BOOTLOADER_HANDSHAKE_HOST (0x00B3) and
```
mookerji: Maybe feature creep for the spec, but perhaps it might be useful to replace by:

```yaml
id: 0x00B0
public: False
deprecated: True
deprecated_in: 0.12
replaced_by:   
  - MSG_BOOTLOADER_HANDSHAKE_HOST
  - MSG_BOOTLOADER_HANDSHAKE_DEVICE
```

Also, I think the short description here might be too long for the latex columns in the datasheet. I may be wrong though.

mfine: Also related, what about a `deprecated` package to put these messages into? vs. cluttering up our existing packages. And maybe a better prefix than `_OLD` - like, `_DEPRECATED` instead?

spec/yaml/swiftnav/sbp/bootload.yaml
```diff
@@ -22,15 +22,40 @@ include:
...
     id: 0x00B0
     public: True
-    short_desc: Bootloading handshake (host <=> device)
+    short_desc: Bootloading handshake response (host <= device)
     desc: |
-      The handshake message establishes a handshake between the device
-      bootloader and the host.  The payload string contains the
-      bootloader version number, but returns an empty string for
-      earlier versions.
+      The handshake message response from the device establishes a
```
mookerji: Similarly, may also be feature creep, but having the response message be parseable:

```
short_desc: Bootloading handshake response (host <= device)
response: MSG_BOOTLOADER_HANDSHAKE_DEVICE
```

spec/yaml/swiftnav/sbp/flash.yaml
```diff
@@ -64,15 +64,51 @@ definitions:
...
+          desc: |
+            Length of set of addresses to program, counting up from
+            starting address
+      - data:
+          type: array
+          fill: u8
+          size_fn: addr_len
+          desc: Data to program addresses with, with length N=addr_len
+
  - MSG_FLASH_DONE:
```
mookerji: Should this be renamed to `MSG_FLASH_DONE` or `MSG_FLASH_DEVICE` or something to be consistent with the naming convention?

spec/yaml/swiftnav/sbp/bootload.yaml
mookerji: Yeah,  separating them is definitely an idea worth considering. I don't know how complicated it would be to move them to a `deprecated.yaml` or if that will ultimately get turned into a ghetto as well. My initial thought is that it'd be totally fine. That would make it easier to have a deprecated messages table in the datasheet.

mookerji: Regarding the name, it's not obvious to me that the name even has to change, given that the ID is what matters and we clearly indicate that it's been deprecated.

mookerji:
This overall plan is cool. @fnoble may have opinions about whether having a deprecated package specifically is reasonable.

Related to clarity, but unrelated to the bootloader, the settings package may also suffer from the same host/device/request/response issues that are fixed by the approach here.

@cbeighley ^^^

cbeighley:
Are there any other Call/Response messages in the protocol atm? Other than the settings stuff?

mookerji:
So there's also `file_io` and some request messages in `piksi` that don't have defined response types.

spec/yaml/swiftnav/sbp/bootload.yaml
mfine: If you want to use old and new names in the same space without package specifiers, you'd need different names, e.g., `MSG_EPHEMERIS` and `MSG_EPHEMERIS_OLD`. I'm going to change these all over to `MSG_EPHEMERIS_DEPRECATED` and create a deprecated package.

spec/yaml/swiftnav/sbp/bootload.yaml
mfine: Hmm, would that also suggest a paired `request:` field on the response message?

mfine:
Added a `deprecated` package. Added `replaced_by` fields. Changed names from `_OLD` to `_DEPRECATED`. Left out `request` and `response` fields.

mookerji:
This looks fine to me, pending whatever changes to the tests are needed. @cbeighley , you may want to look at it over? I'd double-check the diffs of the message IDs as a last sanity check.

mfine:
I grepped the message id space. I'll add a test that ensures they're all unique when I add / update tests (would fit in well with the static message count test).

mfine:
Added a unique message test (counts all messages, compares against total messages). I got rid of the "extra" deprecated messages, as they don't play well with our scheme of message lookup.

mfine:
Bringing this in.

PR #153, opened 2015-06-04
Surface the metadata

mfine:
Yield the metadata for the `MultiJSONLogIterator`. We should surface metadata to the other iterators, but for now I'm going to surface it here to avoid breaking consumers (I'm the only consumer of this iterator). Should be safe to bring in.

/cc @mookerji


mfine:
Bringing this in.

mookerji:
:+1:

PR #154, opened 2015-06-05
Deprecate bootloader handshake +  send SBP protocol version

mfine:
Setup a new bootloader message from the device, containing both the bootloader version and the SBP protocol version as strings. Also add SBP protocol version as a string to the heartbeat.

There's a lot more changes that need to happen on top of this - collecting feedback before embarking on those changes.

/cc @fnoble @cbeighley @mookerji

### DIFF
--- a/spec/yaml/swiftnav/sbp/deprecated.yaml
+++ b/spec/yaml/swiftnav/sbp/deprecated.yaml
@@ -11,7 +11,7 @@
 package: swiftnav.sbp.deprecated
 description: Deprecated SBP messages.
 stable: False
-public: False
+public: True
 include:
   - types.yaml
 definitions:
@@ -34,6 +34,7 @@ definitions:
 
  - MSG_EPHEMERIS_DEPRECATED:
     id: 0x001A
+    public: False
     short_desc: Deprecated
     desc: Deprecated.
     replaced_by:
--- a/spec/yaml/swiftnav/sbp/deprecated.yaml
+++ b/spec/yaml/swiftnav/sbp/deprecated.yaml
@@ -16,6 +16,22 @@ include:
   - types.yaml
 definitions:
 
+ - MSG_BOOTLOADER_HANDSHAKE_DEPRECATED:
+    id: 0x00B0
+    public: True
+    short_desc: Bootloading handshake response (host <= device)
+    desc: |
+      The handshake message response from the device establishes a
+      handshake between the device bootloader and the host. The
+      request from the host is MSG_BOOTLOADER_HANDSHAKE_HOST.  The
+      payload string contains the bootloader version number, but
+      returns an empty string for earlier versions.
+    fields:
+      - handshake:
+          type: array
+          fill: u8
+          desc: Version number string (not NULL terminated)
+
  - MSG_EPHEMERIS_DEPRECATED:
     id: 0x001A
     short_desc: Deprecated
--- a/spec/yaml/swiftnav/sbp/system.yaml
+++ b/spec/yaml/swiftnav/sbp/system.yaml
@@ -55,8 +55,10 @@ definitions:
                   values:
                       - 0: No external antenna detected
                       - 1: External antenna is present
-              - 3-30:
+              - 27-30:
                   desc: Reserved
+              - 3-26:
+                  desc: SBP protocol version number
               - 2:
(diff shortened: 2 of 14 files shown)


cbeighley:
lgtm

spec/yaml/swiftnav/sbp/system.yaml
```diff
@@ -66,3 +66,6 @@ definitions:
                   values:
                       - 0: System Healthy
                       - 1: An error has occurred
+        - protocol:
```
mookerji: uh, I believe that the intention for the version stuff would go into the reserved field of the heartbeat message?

mfine: @mookerji are we using a version string? If we we're packing the version elements (major, minor, sub-minor) into the reserved field, that might be possible - but if we're using an opaque string, that doesn't seem like it will work.

/cc @henryhallam

cbeighley: We've currently got a bit of a mix of encoding vs raw strings for versioning info throughout Piksi. It would probably be good to standardize this. Off the top of my head:

* STM FW Version
  * String
* NAP FW Version
  * String
  * Note: Older versions were a 4 byte identifier
* Bootloader Version
  * String
  * Note: Older versions were a 1 byte identifier
* HW Version
  * Currently a 4 byte identifier. Should we change this to a string?
* SBP Protocol Version
  * This PR would introduce it as a string

mfine:
:+1: for standardizing.

spec/yaml/swiftnav/sbp/system.yaml
mookerji: @mfine: Not a version string. In this case, this would put the SBP protocol minor version into the reserved field, like 42.

mookerji:
:+1: for standardizing

spec/yaml/swiftnav/sbp/system.yaml
fnoble: The NAP FW string has already proved to be troublesome to use as a string programmatically - the first 100 lines of main.c in piksi_firmware are a NAP version comparison function (https://github.com/swift-nav/piksi_firmware/blob/master/src/main.c#L57).

I'd vote for making the SBP protocol version a tuple of `u8`s or something else thats easy to compare programmatically and is more efficient to send in frequent message such as the `MSG_HEARTBEAT`.

Standardization would be good eventually but I don't think we should get too deep into that right now? I'd say we should make the best choice for the SBP version and use that going forward and we can improve the other versioning schemes later.

For the FW versions I think strings make more sense anyway as they are primarily just for displaying to the user.

mfine: So we'd have 24 bits in the reserved field of the heartbeat message? And use a similar fields scheme for the new bootloader handshake? Makes sense to be more formal and rigid with the SBP protocol version. Any objections?

cbeighley: Sounds good

mookerji: :+1: :100:

cbeighley:
Do we need 24 bits for the SBP protocol version? Seems like a lot. Would be nice if it was byte aligned as well for the heartbeat. What about 16 bits from [8:24)?

mfine:
@cbeighley k, had it in my head we wanted a 3 tuple - but yeah, just needs a `major.minor` for communication. Will update and make it byte aligned.

mfine:
bump on any reviews here.

spec/yaml/swiftnav/sbp/deprecated.yaml
```diff
@@ -16,6 +16,22 @@ include:
   - types.yaml
 definitions:
 
+ - MSG_BOOTLOADER_HANDSHAKE_DEPRECATED:
+    id: 0x00B0
+    public: True
```
fnoble: Do you think deprecated messages should be public? I can see arguments both ways...

mfine: Yeah, maybe they should inherit their previous state? Like if they were public, they stay public? And vice versa.

mookerji:
Looking at this just now.

mookerji:
Looks good. Merge at will.

PR #158, opened 2015-06-09
Version available in c

mfine:
This is what I got for getting a c header version. Not great (you have to choose the version ahead of time, which might change by the time you finally bring it in). But oh well.

/cc @mookerji @cbeighley


mookerji:
What do you mean by "change?"

mfine:
Since the version is generated, there can be races to get versions in (I do v16 you do v16, then we merge, etc.). Also we need to manually upkeep this as versions change. Not the end of the world, but something to keep an idea about. Maybe there's a better way to do this?

mookerji:
lgtm

PR #159, opened 2015-06-09
Use a zero length array instead of a pointer

mfine:
Represent strings a 0 length arrays to avoid problems calculating message sizes. (is there a better way to represent this or is this acceptable?)

/cc @mookerji @fnoble @cbeighley

### DIFF
--- a/c/include/libsbp/bootload.h
+++ b/c/include/libsbp/bootload.h
@@ -51,7 +51,7 @@
 #define SBP_MSG_BOOTLOADER_HANDSHAKE_DEVICE 0x00B4
 typedef struct __attribute__((packed)) {
   u32 flags;      /**< Bootloader flags */
-  char* version;    /**< Bootloader version number */
+  char[0] version;    /**< Bootloader version number */
 } msg_bootloader_handshake_device_t;
 
 
--- a/c/include/libsbp/logging.h
+++ b/c/include/libsbp/logging.h
@@ -35,7 +35,7 @@
  */
 #define SBP_MSG_PRINT     0x0010
 typedef struct __attribute__((packed)) {
-  char* text;    /**< Human-readable string */
+  char[0] text;    /**< Human-readable string */
 } msg_print_t;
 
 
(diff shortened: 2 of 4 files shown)


mfine:
Also, does gcc have better support for doing stuff like this?

mookerji:
@henryhallam ^^^

lgtm

fnoble:
lgtm, we shouldn't ever have a pointer in an SBP message so might be good to check if there are other occurrences of this (I think we fixed one other one before at least).

Re gcc, this is actually the gcc way of doing it. C99 you do [] instead of [0], the difference is C99 will give you an error if you try to do sizeof but the gcc way will assume the array is zero length.

https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html

mfine:
Checked for more pointer fields - none left:

```
➜  c/include/libsbp git:(master) ag '\*.*;'                   
edc.h
18:u16 crc16_ccitt(const u8 *buf, u32 len, u16 crc);

sbp.h
38:typedef void (*sbp_msg_callback_t)(u16 sender_id, u8 len, u8 msg[], void *context);
47:  void *context;                       /**< Pointer to a context */
48:  struct sbp_msg_callbacks_node *next; /**< Pointer to next node in list. */
67:  void* io_context;
68:  sbp_msg_callbacks_node_t* sbp_msg_callbacks_head;
74:                         sbp_msg_callbacks_node_t *node);
75:void sbp_clear_callbacks(sbp_state_t* s);
76:sbp_msg_callbacks_node_t* sbp_find_callback(sbp_state_t* s, u16 msg_type);
77:void sbp_state_init(sbp_state_t *s);
78:void sbp_state_set_io_context(sbp_state_t *s, void* context);
79:s8 sbp_process(sbp_state_t *s, u32 (*read)(u8 *buff, u32 n, void* context));
81:                    u32 (*write)(u8 *buff, u32 n, void* context));
➜  c/include/libsbp git:(master)
```

Whereas on the previous builds, the `char *`'s had show'd up.

PR #161, opened 2015-06-09
reorder major and minor fields

mfine:
I think this makes more sense, to have the minor number in the lower bits.

/cc @cbeighley


mfine:
(note, this doesn't change any generated code (except documentation)).

cbeighley:
Agreed, merge it

PR #168, opened 2015-06-16
Versioning Proposal

mfine:
Here's a first pass through versioning policy and generated artifacts in major version directories. Comments very welcome.

+ Updated the versioning policy to cover message stability.
+ Updated the generators for c, python, latex - this is kind of a mess and might need a refactor.
+ C headers generated; non-header files left as is (and common.h - not using a relative path!)
+ Python generated. Didn't update the client code here or the tables stuff.

Mainly looking to see what we think about this approach.

/cc @fnoble @mookerji @cbeighley


VERSIONING.md
```diff
@@ -0,0 +1,19 @@
+In order to support backwards and forwards compatibility, SBP messages
+are versioned and follow a small number of conventions, enabling
+devices to indicate understood message formats, capabilities and
+features. SBP uses a major, minor numbering scheme to indicate
+versions, where major and minor are represented by 8-bit values. The
+minor number is incremented by each message addition, deprecation. The
+major number is incremented when the format of a message is
```
mookerji: The clarifyL "message" here refers to the SBP packet structure or the change of a stable message?

mfine: Change of a stable message. There's no provision here for changing the SBP framing structure. Should we include some language do disallow that?

VERSIONING.md
```diff
@@ -0,0 +1,19 @@
...
+changed.
+
+Message ID's are guaranteed to be unique for all minor versions in a
```
mookerji: An alternative layout would make this easier to parse? Possibly something like ... 

* *Major version*. The major number is incremented when the format of a message is changed. Message ID's are not guaranteed to be unique between major versions. The addition or deprecation of stable messages results in the incrementing of the major version. Stable messages are guaranteed to be available for all minor versions in a major version

* *Minor version*.  The minor number is incremented by each non-stable message addition, deprecation.  Changes to a message in a major version are manifest by incrementing the minor version, adding a new message with the changes, and deprecating the old message.

c/include/libsbp/v0/acquisition.h
```diff
@@ -20,10 +20,10 @@
  * * Satellite acquisition messages from the device.
  * \{ */
 
-#ifndef LIBSBP_ACQUISITION_MESSAGES_H
-#define LIBSBP_ACQUISITION_MESSAGES_H
+#ifndef LIBSBP_V0_ACQUISITION_MESSAGES_H
+#define LIBSBP_V0_ACQUISITION_MESSAGES_H
 
-#include "common.h"
+#include "<libsbp/common.h>"
```
mookerji: 0_0 mixed quotation angle brackets?

generator/sbpg/targets/resources/sbp_messages_template.h
```diff
@@ -20,10 +20,10 @@
  * (((description|commentify)))
  * \{ */
 
-#ifndef LIBSBP_(((pkg_name|upper)))_MESSAGES_H
-#define LIBSBP_(((pkg_name|upper)))_MESSAGES_H
+#ifndef LIBSBP_V(((major)))_(((pkg_name|upper)))_MESSAGES_H
+#define LIBSBP_V(((major)))_(((pkg_name|upper)))_MESSAGES_H
 
-#include "common.h"
+#include "<libsbp/common.h>"
```
mookerji: ditto

VERSIONING.md
mookerji: Just wanted to make sure that's what it was referring to. 

A mention of stable vs unstable/draft, similar to what we have in the datasheet would be good: "By convention, we refer to messages as `stable` and `draft`. Stable messages are intended for broad use, like `MsgBaselineECEF`, while draft ones are intended for internal development.`.

generator/sbpg/targets/resources/sbp_messages_template.h
```diff
@@ -53,4 +53,4 @@ typedef struct __attribute__((packed)) {
 ((* endfor *))
 /** \} */
 
-#endif /* LIBSBP_(((pkg_name|upper)))_MESSAGES_H */
+#endif /* LIBSBP_V(((major)))_(((pkg_name|upper)))_MESSAGES_H */
```
fnoble: Could we add a newline at the end of the file?

VERSIONING.md
fnoble: Agree, I think the content is spot on but could be restructured to be easier to read.

mfine:
Completion criteria: completed `versioning.md` and version containers.

PR #170, opened 2015-06-17
Start the handler inside the context manager

mfine:
Start the handler inside context.

/cc @cbeighley


mfine:
Bringing this is in!

Issue #166, opened 2015-06-16
Promote Settings to Stable


mfine:
To reach SBP version 1.0, we should promote the settings messages to stable:

MSG_SETTINGS
MSG_SETTINGS_SAVE
MSG_SETTINGS_READ_BY_INDEX

Insure their form and fitness for stability.

/cc @fnoble @mookerji @denniszollo

mfine:
Ok, these messages are funky. They've got some weird edge cases and their bi-directionality is problematic. My proposal is that we:

1. Remove MSG_SETTINGS and replace it with a one-way MSG_SETTINGS_WRITE. This is always from the host to the device and has a payload of values to write.
2. MSG_SETTINGS_SAVE is unchanged
3. MSG_SETTINGS_READ_BY_INDEX gets replaced with MSG_SETTINGS_READ_HOST and MSG_SETTINGS_READ_DEVICE (the READ_DEVICE message takes READ_BY_INDEX's message id). READ_DEVICE always sends a string (no NULL).

Will put together some changes in sbp and firmware for this. Hope this will get us towards making these stable.

Thoughts?

mookerji:
:+1: Consistent with the other request/response changes we're making. I'm checking out the ref'd PR now.

mfine:
Ongoing PR's:

https://github.com/swift-nav/libsbp/pull/172
https://github.com/swift-nav/piksi_tools/pull/82
https://github.com/swift-nav/piksi_firmware/pull/449

PR #172, opened 2015-06-18
Improved Settings

mfine:
Improve the settings messages and promote them to stable. https://github.com/swift-nav/libsbp/issues/166.

/cc @mookerji @fnoble @cbeighley


mfine:
Have some questions here about compatibility. Looking in the firmware code, there's definitely some questions about how to move forward wrt the old messages. Do we want to preserve the ability to read a single settings value? Or is the read by index good enough?

mfine:
After playing around with reading settings on the device, I may have misunderstood this API a bit. Thinking there should be the following messages:

```
MSG_SETTINGS_READ_DEVICE
MSG_SETTINGS_READ_HOST

MSG_SETTINGS_READ_BY_INDEX_DEVICE
MSG_SETTINGS_READ_BY_INDEX_HOST

MSG_SETTINGS_WRITE
MSG_SETTINGS_SAVE
```

Not sure how best to indicate not found setting for `MSG_SETTINGS_READ`. Should not use NULL for payload. Maybe empty string for setting?

Not sure how best to indicate `DONE` for `MSG_SETTINGS_READ_BY_INDEX`. Should not use NULL for payload. Maybe -1 for index or empty string for setting?

mookerji:
* Not found setting `MSG_SETTINGS_READ`: `NOT FOUND`?
* Done: `END`?

c/include/libsbp/settings.h
```diff
@@ -51,22 +50,40 @@ such strings on reads.
...
+/** Read setting by direct index (host => device)
  *
  * The settings message for iterating through the settings
  * values. It will read the setting at an index, returning a
  * NULL-terminated and delimited string with contents
- * [SECTION_SETTING, SETTING, VALUE].
+ * [SECTION_SETTING, SETTING, VALUE] in MSG_SETTINGS_READ_DEVICE.
  */
-#define SBP_MSG_SETTINGS_READ_BY_INDEX 0x00A2
+#define SBP_MSG_SETTINGS_READ_HOST   0x00A2
```
fnoble: What about REQUEST and RESPONSE instead of HOST and DEVICE? I could imagine a potential world in which the notion of host and device are less clear than today.

mookerji: Alas, I think we already chose HOST/DEVICE with the messages we changed this way  : /.

mfine: I went with HOST/DEVICE here because we kind of went with that paradigm with the bootloader messages. Would prefer consistency here and be illustrative in this case about who's sending what... but I do long term like the more generic terms REQUEST, RESPONSE.

gsmcmullin: I prefer REQUEST/RESPONSE.  HOST/DEVICE does not give me any idea of who sends what.

fnoble: Did we ship anything with HOST/DEVICE yet? Could just change the names of the other ones if everyone prefers REQUEST/RESPONSE and save us some debt down the line.

mfine: They should just be names so far.

mfine: We can pivot and change everything back over to REQUEST/RESPONSE. Seems like a better scheme - I think the concern was we didn't want to proliferate REQ/RESP pattern in piksi communication. For these cases, though, I think Request (host) / Response (device) is apt and trying to avoid a REQ/RESP pattern has instead caused more problems than its solved. (I too have found it confusing figuring out the direction of HOST/DEVICE despite coming up with the naming :).

spec/yaml/swiftnav/sbp/settings.yaml
```diff
@@ -17,39 +17,50 @@ description: |
...
+    fields:
+      - index:
+          type: u16
+          desc: |
+            An index into the device settings, with values ranging from
+            0 to length(settings)
+
+ - MSG_SETTINGS_READ_DEVICE:
+    id: 0x00A3
+    short_desc: Read setting by direct index (host <= device)
```
denniszollo: I don't think we want "VALUE" in this field.  The read should only need the group and the setting name

mfine:
Completion criteria: new settings messages installed and promoted to stable.

mfine:
Respun this change with all of the feedback.

denniszollo:
So from this I expect that we plan to add the complexity into the console to support both types of settings messages?  Or do we think we don't need that complexity before the 1.0?

mfine:
Corresponding pull request for piksi_tools is https://github.com/swift-nav/piksi_tools/pull/82. It supports both new and old messages.

mfine:
I eased up on the deprecation. Ready for review.

spec/yaml/swiftnav/sbp/settings.yaml
```diff
@@ -17,39 +17,55 @@ description: |
...
-      R/W device configuration settings (host <=> device)
+ - MSG_SETTINGS_SAVE:
+    id: 0x00A1
+    short_desc: Save settings to flash (host => device)
     desc: |
-      The setting message reads and writes the device's configuration.
+      The save settings message persists the device's current settings
+      configuration to its onboard flash memory file system.
+
+ - MSG_SETTINGS_WRITE:
```
mookerji: Does MSG_SETTINGS_WRITE need a accompanying response message or are we assuming that the caller will have to read back settings to make sure it was successful?

mfine: Oh, hm, that' a good question. The preceding behavior was echoing back the write (it was multipurpose and also being used for reads). But then the usage of it in the console was just super weird and bad - I just removed the return message and things work fine. It's just super weird around this call now. What do you think?

mfine:
Going to bring this in if there are no objections.

spec/yaml/swiftnav/sbp/settings.yaml
denniszollo: what if we send a read response with the new value upon successful write?

PR #173, opened 2015-06-19
UDP logger for SBP

mfine:
A simple UDP logger. Sends messages over a UDP address:port pair. There should be some library help around framing out the message instead of doing it explicitly here. Haven't tried this out yet.

/cc @fnoble @denniszollo

### DIFF
--- a/python/sbp/client/examples/udp.py
+++ b/python/sbp/client/examples/udp.py
@@ -15,9 +15,8 @@ messages from a serial port and sending them to a UDP socket.
 
 from sbp.client.drivers.pyserial_driver import PySerialDriver
 from sbp.client.handler import Handler
+from sbp.client.loggers import UdpLogger
 
-from contextlib import closing
-import socket
 import struct
 import time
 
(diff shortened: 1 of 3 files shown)


fnoble:
lgtm, @mookerji suggested adding a unit test before like this: https://github.com/swift-nav/pySTR4500/blob/master/test/test_client.py.

python/sbp/client/examples/udp.py
```diff
@@ -15,9 +15,8 @@
 
 from sbp.client.drivers.pyserial_driver import PySerialDriver
 from sbp.client.handler import Handler
+from sbp.client.loggers import UdpLogger
```
denniszollo: Is this going to work?  The file is named udp_logger...

mfine: Oh yea, should be

```python
from sbp.client.loggers.udp_logger import UdpLogger
```

denniszollo: Also, we should make the file executable.

mfine:
Going to bring this in for integration. Will finish the test in a later PR.

PR #174, opened 2015-06-19
Make add and remove callbacks take lists of message for convenience

mfine:
Check to see if the msg_type is a list, and then apply the callbacks to all the message types. Better way to do something like this?

/cc @fnoble @denniszollo

### DIFF
--- a/python/sbp/client/handler.py
+++ b/python/sbp/client/handler.py
@@ -185,7 +185,7 @@ class Handler(object):
       List type adds the callback to all the message types.
     """
     try:
-      for mt in msg_type:
+      for mt in iter(msg_type):
         self.callbacks[mt].add(callback)
     except TypeError:
       self.callbacks[msg_type].add(callback)
@@ -203,7 +203,7 @@ class Handler(object):
       List type removes the callback from all the message types.
     """
     try:
-      for mt in msg_type:
+      for mt in iter(msg_type):
         self.callbacks[mt].remove(callback)
     except TypeError:
       self.callbacks[msg_type].remove(callback)
--- a/python/sbp/client/handler.py
+++ b/python/sbp/client/handler.py
@@ -180,15 +180,15 @@ class Handler(object):
     ----------
     callback : fn
       Callback function
-    msg_type : int | list
+    msg_type : int | iterable
       Message type to register callback against. Default `None` means global callback.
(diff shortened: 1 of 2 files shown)


python/sbp/client/handler.py
```diff
@@ -180,10 +180,15 @@ def add_callback(self, callback, msg_type=None):
     ----------
     callback : fn
       Callback function
-    msg_type : int
+    msg_type : int | list
       Message type to register callback against. Default `None` means global callback.
+      List type adds the callback to all the message types.
     """
-    self.callbacks[msg_type].add(callback)
+    if not isinstance(msg_type, list):
```
fnoble: This could be generalized to all iterables. The most "pythonic" way to check if something is an iterable is just to assume it is and catch failure:

```python
try:
  for mt in msg_type:
    self.callbacks[mt].add(callback)
catch TypeError:
  self.callbacks[msg_type].add(callback)
```

http://stackoverflow.com/questions/1952464/in-python-how-do-i-determine-if-an-object-is-iterable

python/sbp/client/handler.py
```diff
@@ -180,10 +180,15 @@ def add_callback(self, callback, msg_type=None):
     ----------
     callback : fn
       Callback function
-    msg_type : int
+    msg_type : int | list
```
fnoble: If you generalize to iterables you should also s/list/iterable in the docstring

mfine: Done.

mfine:
Moved to iterable and added some tests.

python/sbp/client/handler.py
```diff
@@ -180,10 +180,15 @@ def add_callback(self, callback, msg_type=None):
     ----------
     callback : fn
       Callback function
-    msg_type : int
+    msg_type : int | iterable
       Message type to register callback against. Default `None` means global callback.
+      List type adds the callback to all the message types.
```
fnoble: Still a "List" here

denniszollo:
Love it

python/sbp/client/handler.py
fnoble: But maybe that more instructive?

fnoble:
lgtm

mfine:
Fixed the docs. Bringing in.

PR #176, opened 2015-06-19
Step1: Fix Piksi's lack of social networking features.

mookerji:
Step3: ....

/cc @fnoble @mfine @henryhallam


cbeighley:
We have a prospective logo that would go along really well with this.

PR #177, opened 2015-06-19
UDP logger test

mfine:
Added a spec for UDP logging.


mookerji:
Great use of closures/factory whatever.

PR #182, opened 2015-06-23
Fix wait, and expose is_alive

mfine:
Fix wait so we can use it. Expose is_alive.

/cc @denniszollo @gsmcmullin @cbeighley


cbeighley:
Looks good. Is Handler.is_alive() going to be called by serial_link or something?

mfine:
Yeah, we're using it right of receiver_thread right now to check thread death. We'll move over to this call.

PR #183, opened 2015-06-24
New file_io messages

mfine:
Remove bi-directionality. Request, response all the things. No deprecation necessary. Were there more changes to these messages we wanted? A `dir_done` message vs. the 0xff thing might be nice. Thoughts?

### DIFF
--- a/spec/yaml/swiftnav/sbp/file_io.yaml
+++ b/spec/yaml/swiftnav/sbp/file_io.yaml
@@ -71,6 +71,10 @@ definitions:
           type: string
           size: 20
           desc: Name of the file read from (NULL padded)
+      - contents:
+          type: array
+          fill: u8
+          desc: Contents of read file
 
  - MSG_FILEIO_READ_DIR_REQUEST:
     id: 0x00A9
@@ -114,6 +118,10 @@ definitions:
           type: string
           size: 20
           desc: Name of the directory to list (NULL padded)
+      - contents:
+          type: array
+          fill: u8
+          desc: Contents of read directory
 
  - MSG_FILEIO_REMOVE:
     id: 0x00AC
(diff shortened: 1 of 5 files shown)


spec/yaml/swiftnav/sbp/file_io.yaml
```diff
@@ -51,18 +50,39 @@ definitions:
...
+    fields:
+      - offset:
+          type: u32
+          units: bytes
+          desc: File offset
+      - chunk_size:
+          type: u8
+          units: bytes
+          desc: Chunk size read
+      - filename:
```
denniszollo: This is missing the actual byte stream from the file.  In keeping with the current design, you would get these fields + the rest of the 256 bytes with a portion of the file.

spec/yaml/swiftnav/sbp/file_io.yaml
```diff
@@ -75,9 +95,28 @@ definitions:
...
+    fields:
+      - offset:
+          type: u32
+          desc: |
+            The offset to skip the first n elements of the file list
+      - dirname:
+          type: string
+          size: 20
+          desc: Name of the directory to list (NULL padded)
+
```
denniszollo: Same here - where is the field for the actual data (i.e the result of the "dir" command)  Are people supposed to parse that out of the payload themselves?

mfine:
Contents fields added for both responses.

mookerji:
Instead of having a special message indicating that a response is done, what about using an empty response to indicate that a request is finished?

mfine:
Pending the firmware changes and a rebase, this should be ready to go. The comment around the contents has been addressed. There's already an existing signal to finish dir reading - 0xff, we'll just leave that.

mfine:
Rebased.

mfine:
Going to bring this in if there are no objections.

mookerji:
:+1:

Issue #184, opened 2015-06-24
Remove the deprecated packages


mfine:
It makes deprecating messages with substructures unfeasible.

mfine:
Fixed by #185.

PR #185, opened 2015-06-24
Widen PRNs

mfine:
Widen the PRN field and add some missing fields to the ephemeris (IODC).  Also, a `reserved` field is being added to the ephemeris to capture future small fields we might want to capture (AODO, URA values, anti-spoof, alert flag).

Note a couple of things about deprecated messages: they're not in the deprecated package because sub structures make it non-feasible to move around (see #184). Also, since there can be multiple message deprecations, there's now a naming scheme for deprecations: `DEPRECATED_<GREEK_LETTER>`.

This just captures message definition changes. If things look alright, the rest of the changes will be worked in.

/cc @fnoble @henryhallam @mookerji @gsmcmullin


spec/yaml/swiftnav/sbp/acquisition.yaml
```diff
@@ -17,8 +17,37 @@ include:
...
+      This message describes the results from an attempted GPS signal
+      acquisition search for a satellite PRN over a code phase/carrier
+      frequency range. It contains the parameters of the point in the
+      acquisition search space with the best signal-to-noise (SNR)
+      ratio.
+    fields:
+        - snr:
+            type: float
+            desc: |
+              SNR of best point. Currently dimensonless, but will have
```
mookerji: @fnoble This is now dB-Hz, right?

spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -140,8 +136,129 @@ definitions:
...
+          desc: Satellite is healthy?
+      - sid:
+          type: u32
+          desc: |
+            Signal identifier being tracked - if the 24 MSB are not set
+            itrepresents a PRN-1 identifier
+      - iode:
+          type: u8
+          desc: Issue of ephemeris data
+      - iodc:
```
mookerji: @gsmcmullin @henryhallam opinions?

spec/yaml/swiftnav/sbp/acquisition.yaml
henryhallam: I don't think acquisition results are presently reported as dbHz, only tracking C/No.  We could change that.

spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -253,3 +370,52 @@ definitions:
       - iode:
           type: u8
           desc: Issue of ephemeris data
+
+ - PackedObsContentDeprecatedAlpha:
```
mookerji: What happened with the deprecated package?

spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -140,8 +136,129 @@ definitions:
             desc: Height
 
  - MSG_EPHEMERIS:
+    id: 0x0047
```
mookerji: oh boy, yet another ephemeris message....

spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -140,8 +136,129 @@ definitions:
...
+    short_desc: WGS84 satellite ephemeris parameters
```
henryhallam: s/WGS84 satellite ephemeris parameters/Satellite broadcast ephemeris/
and delete reference to WGS84 from long description

spec/yaml/swiftnav/sbp/tracking.yaml
```diff
@@ -18,7 +18,46 @@ include:
 definitions:
 
  - TrackingChannelState:
-    public: True
+    short_desc: Satellite tracking channel state
```
mookerji: Also, there was some notion of making `MsgObs`, `MsgEphemeris`, and `MsgBasePos` public, right? I think we're going to have to support stability on a message-by-message basis, which I think means fucking with the libsbp generator again.

spec/yaml/swiftnav/sbp/observation.yaml
henryhallam: One reasonable approach would be to include all the fields that are in the [RINEX ephemeris record](https://igscb.jpl.nasa.gov/igscb/data/format/rinex211.txt) (Table A4).
We're missing the "codes on L2 channel" (u8) and "L2 P data flag" (bool) though I could be convinced to omit these.
Also missing SV accuracy, fit interval and transmission time fields.
SV health field is currently a bool but should contain 6 bits.

spec/yaml/swiftnav/sbp/acquisition.yaml
mookerji: That's probably worth doing, if only because it (i) seems like a low-cost/penalty change and (ii) is probably good to be consistent tracking/obs.

spec/yaml/swiftnav/sbp/observation.yaml
mfine: The thinking was to track some of these smaller items in the `reserved` field below (SV accuracy, health, L2 P data, etc.) without defining them up front. A run through the fields suggested the 32-bits in the reserved field would give us enough room to cover them when we're interested in bringing them in. What do you think?

spec/yaml/swiftnav/sbp/observation.yaml
mookerji: Derp. Reading commit message.

mookerji:
lol @ deprecating message scheme.

spec/yaml/swiftnav/sbp/observation.yaml
mfine: From the PR description:

Note a couple of things about deprecated messages: they're not in the deprecated package because sub structures make it non-feasible to move around (see #184). Also, since there can be multiple message deprecations, there's now a naming scheme for deprecations: `DEPRECATED_<GREEK_LETTER>.`

What do you think?

spec/yaml/swiftnav/sbp/tracking.yaml
mfine: We're making everything public. We're making some packages stable, other not stable. We don't have a per-message stability, but we might need to.

mookerji: Ack, I meant "stable".

spec/yaml/swiftnav/sbp/acquisition.yaml
```diff
@@ -17,8 +17,37 @@ include:
...
+            units: chips
+            desc: Code phase of best point
+        - cf:
+            type: float
+            units: hz
+            desc: Carrier frequency of best point
+        - sid:
+            type: u32
+            desc: |
+              Signal identifier of the satellite signal for which acquisition
```
mookerji: This sid descriptor seems a bit ambiguous. It may just be better to say something like: '24 MSB are empty. 8 LSB represent an L1 PRN - 1 identifier.' Otherwise, someone will ask, `What if it's 24 MSB are not empty?' and be confused.

henryhallam: How about something like
"Values 0x00000000 through 0x0000001F represent GPS PRNs 1 through 32 respectively (PRN-1 notation).  Other values reserved for future use."

spec/yaml/swiftnav/sbp/acquisition.yaml
mfine: @mookerji @henryhallam resolution here? This wasn't part of the proposed change - should it be added in or is it out of scope?

mookerji: I think it's out of scope for the time being...

spec/yaml/swiftnav/sbp/observation.yaml
mookerji: @henryhallam Do you think the rest of that could be fit into a u32. My thought is "yes".

mfine: @fnoble and I did a rough bit count up and thought it would be able to fit.

spec/yaml/swiftnav/sbp/acquisition.yaml
mfine: @mookerji @henryhallam how about:

```
Signal identifier being tracked - values 0x00 through 0x1F represent GPS PRNs
1 through 32 respectively (PRN-1 notation); other values reserved for future use.
```

mookerji: :+1:

henryhallam: yep

mfine:
Rebased and picked up Henry's comments. Going to now start integrating into the rest of the code base.

spec/yaml/swiftnav/sbp/acquisition.yaml
henryhallam: I can put out a PR to fix this in the firmware real quick.  I'm pretty certain it isn't used anywhere either in firmware or host-side so no plausible detrimental effects.

fnoble: This is used in the automated testing / bring up script. I don't think it would be a big deal but we might have to adjust some threshold or tell Andrew that there has been a change. @cbeighley ?

cbeighley: Yes, will require changing the threshold. I've made an issue https://github.com/swift-nav/piksi_bringup/issues/7

henryhallam: https://github.com/swift-nav/piksi_firmware/pull/484

mfine:
Generated files, condensed naming.

mookerji:
I think you're going to have to rebase on the slots PR. It's going to conflict on the generated code, so after rebasing, regenerating the code should fix that issue.

mfine:
@mookerji rebased.

mfine:
The changes emanating from this PR and widening the PRN is captured in this PR and in the following other places:

https://github.com/swift-nav/piksi_tools/pull/112
https://github.com/swift-nav/piksi_firmware/pull/486
https://github.com/swift-nav/sbp_log_analysis/pull/229

Because the nature of this change, the following sequencing needs to happen:

1. Bring in the `libsbp` changes and publish the new library
2. Update the `piksi_tools` and `sbp_log_analysis` consumers of `libsbp` (these should be backwards compatible and able to support old and new message formats)
3. evaluate the `piksi_firmware` changes in HITL
4. bring in the `piksi_firmware` changes once the HITL results look good

How do people feel about these steps and changes?

mookerji:
So given @mfine 's local integration testing for the downstream dependencies, I'm pretty happy with the current change and the list of dependencies. The piksi_firmware change is going to require some thorough HITL testing.

PR #187, opened 2015-06-26
doc revisions

mfine:
Little tweaks to make the doc work. Find the generated doc here:

https://github.com/mfine/libsbp/blob/mfine-doc-tweaks/docs/sbp.pdf

/cc @cbeighley @denniszollo @mookerji


mookerji:
Man, I could swear we did something to break up that message ID table across two pages.

mfine:
yeah, looks like we've got too many messages now

mookerji:
We should limit our message table to only those that can fit on a single A4 sheet of paper.

mookerji:
Single page is fine, we can fuck with it later.

mookerji:
Add the semantic versioning policy as well? 
https://github.com/mfine/libsbp/blob/mfine-versioning/VERSIONING.md

mfine:
Doc driven development

mfine:
Yeah, we'll add the versioning policy in when its ready - we're just at point release here.

mookerji:
:+1: merge at will.

mfine:
Also, that `cut` line I put in there is super non-portable.. I tried sed, but it puked - will need to revisit.

PR #191, opened 2015-06-30
tracking: Add message definition for I/Q output.

gsmcmullin:
/cc @henryhallam @fnoble

### DIFF
--- a/c/include/libsbp/tracking.h
+++ b/c/include/libsbp/tracking.h
@@ -55,8 +55,8 @@ typedef struct __attribute__((packed)) {
  * Structure containing in-phase and quadrature correlation components.
  */
 typedef struct __attribute__((packed)) {
-  u32 I;    /**< In-phase correlation */
-  u32 Q;    /**< Quadrature correlation */
+  s32 I;    /**< In-phase correlation */
+  s32 Q;    /**< Quadrature correlation */
 } tracking_channel_correlation_t;
 
 
@@ -68,7 +68,7 @@ typedef struct __attribute__((packed)) {
 #define SBP_MSG_TRACKING_IQ    0x001C
 typedef struct __attribute__((packed)) {
   u8 channel;    /**< Tracking channel of origin */
-  u8 prn;        /**< PRN-1 being tracked */
+  u32 sid;        /**< Identifier of satellite signal being tracked */
   tracking_channel_correlation_t corrs[3];   /**< Early, Prompt and Late correlations */
 } msg_tracking_iq_t;
 
--- a/c/include/libsbp/tracking.h
+++ b/c/include/libsbp/tracking.h
@@ -50,6 +50,29 @@ typedef struct __attribute__((packed)) {
 } msg_tracking_state_t;
 
 
+/** Complex correlation structure
+ *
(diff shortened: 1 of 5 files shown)


c/include/libsbp/tracking.h
```diff
@@ -50,6 +50,29 @@ typedef struct __attribute__((packed)) {
 } msg_tracking_state_t;
 
 
+/** Complex correlation structure
+ *
+ * Structure containing in-phase and quadrature correlation components.
+ */
+typedef struct __attribute__((packed)) {
+  u32 I;    /**< In-phase correlation */
```
henryhallam: should be signed?

c/include/libsbp/tracking.h
```diff
@@ -50,6 +50,29 @@ typedef struct __attribute__((packed)) {
...
+
+/** Tracking channel correlations
+ *
+ * When enabled, a tracking channel can output the correlations at each
+ * update interval.
+ */
+#define SBP_MSG_TRACKING_IQ    0x001C
+typedef struct __attribute__((packed)) {
+  u8 channel;    /**< Tracking channel of origin */
+  u8 prn;        /**< PRN-1 being tracked */
```
henryhallam: Use new signal identifier format? (I understand space is at a premium here, 8-bit is ok by me if you prefer)

c/include/libsbp/tracking.h
gsmcmullin: Correct, I'll fix it.

c/include/libsbp/tracking.h
gsmcmullin: The upstream hasn't changed this yet in other places.  I thought it made sense to change it everywhere all in one go.

fnoble: I think its fine to change it here now, @mfine has PRs that update the other messages by now

fnoble:
looks good, did you think about buffering up a number of samples and then sending them in one SBP message? Might cut down on overhead, which is quite high now. If it works well as is though I don't want to make it more complex than it needs to be.

c/include/libsbp/tracking.h
gsmcmullin: Ok, will correct it.

gsmcmullin:
I thought about buffering, but it will tie up a lot of ram for statically allocated buffers that we usually won't care about.  This will choke if we turn on all channels, but I think that's unlikely to be needed.

fnoble:
Ok, sounds good to me

gsmcmullin:
I already merged this, but do we want to add the carrier frequency to the message?  I think this would be useful.  If so, I'll create a new PR.

henryhallam:
There are a handful of tracking loop variables that have the potential to be useful for certain kinds of debugging.  I think in most situations you only need to be able to get out one channel's worth at a max of 1 kHz, so after overhead and leaving some room for other msgs there are perhaps ~50 bytes available per update.  
 - Prompt I and Q: 8 bytes
 - Early and late I and Q?  16 bytes
 - Carrier freq 4 bytes
 - Sample count 4 bytes
 - Code phase 4 bytes
 - Carrier phase 8 bytes
 - Discriminator outputs? more of a pain to get to

PR #193, opened 2015-07-02
Added libsbp C example.

ungureanuvladvictor:
Serial port library used is: http://sigrok.org/wiki/Libserialport. The example listens for heartbeat messages and prints a confirmation when one is received. It is built separately from libsbp not to introduce any dependency with libserialport.

Signed-off-by: Vlad Ungureanu <vvu@vdev.ro>

/cc: @fnoble @gsmcmullin

### DIFF
--- a/c/example/README.md
+++ b/c/example/README.md
@@ -1,0 +1,48 @@
+## C example for libsbp
+
+An example C program, intended to be run on a host computer connected to
+the Piksi, that parses incoming SBP messages from the Piksi and prints some of
+them to the screen.
+
+## Installation
+
+On Debian-based systems (including Ubuntu 12.10 or later) you can get all
+the requirements with:
+
+```shell
+sudo apt-get install build-essential pkg-config cmake
+```
+
+On other systems, you can obtain CMake from your operating system
+package manager or from http://www.cmake.org/.
+
+You should also install libsbp and libserialport. For instructions please refer to:
+* [libsbp](https://github.com/swift-nav/libsbp/tree/master/c)
+* [libserialport](http://sigrok.org/wiki/Libserialport)
+
+Once you have the dependencies installed, create a build directory where the example will be built:
+
+```shell
+mkdir build
+cd build
+```
+
+Then invoke CMake to configure the build, and then build,
+
+```shell
+cmake ../
+make
+```
+
+## Usage
+
+```shell
+./libsbp_example
+usage: ./libsbp_example [-p serial port]
+```
+
+## LICENSE
+
+Copyright © 2015 Swift Navigation
+
+Distributed under LGPLv3.0.
--- a/c/example/example.c
+++ b/c/example/example.c
@@ -1,0 +1,137 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include <libserialport.h>
+
+#include <libsbp/sbp.h>
+#include <libsbp/system.h>
+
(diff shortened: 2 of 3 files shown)


c/example/README.md
```diff
@@ -0,0 +1,42 @@
+## C example for libsbp
+
```
henryhallam: Describe more clearly what this is, i.e.
An example C program, intended to be run on a host computer connected to the Piksi, that parses incoming SBP messages from the Piksi and prints some of them to the screen.

henryhallam:
Maybe flesh it out a bit by having it print MSG_PRINTs and SPP or baseline solutions, and an example of sending something to the Piski - perhaps MSG_RESET if a certain cmd-line argument is provided, à la serial_link.py?

denniszollo:
I think a good example if bi-directional comms would be reading a setting

denniszollo:
Perhaps we can ask for the firmware version via the settings message and print it out?

ungureanuvladvictor:
+1 for the firmware version, seems more reasonable.

c/example/README.md
```diff
@@ -0,0 +1,42 @@
...
+## Installation
```
henryhallam: mention libserialport dependency

c/example/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+On Debian-based systems (including Ubuntu 12.10 or later) you can get all
+the requierments with:
```
henryhallam: typo: s/requierments/requirements/

henryhallam:
> firmware version 
sounds fine to me

c/example/example.c
```diff
@@ -0,0 +1,137 @@
...
+
+  if (argc <= 1) {
+    usage(argv[0]);
+    exit(EXIT_FAILURE);
+  }
+
+  while ((opt = getopt(argc, argv, "p:")) != -1) {
+    switch (opt) {
+      case 'p':
+        serial_port_name = (char *)calloc(strlen(optarg) + 1, sizeof(char));
```
gsmcmullin: Why not strdup(optarg)?

c/example/example.c
```diff
@@ -0,0 +1,137 @@
...
+  sbp_state_init(&s);
+
+  sbp_register_callback(&s, SBP_MSG_HEARTBEAT, &hearbeat_callback, NULL,
+                        &heartbeat_callback_node);
+
+  while(1) {
+    sbp_process(&s, &piksi_port_read);
+  }
+
+  result = sp_close(piksi_port);
```
gsmcmullin: This code is all dead.  How do we get out of the while(1) loop?

ungureanuvladvictor: We do not get out...read while CTRL+C or something like that happens. I can put a signal handler to behave nice in those cases.

c/example/example.c
ungureanuvladvictor: Did not think about it.

ungureanuvladvictor:
My OSX installation misbehaves with the piksi. If I add another callback entry all the usb bus freezes and leaves everything unresponsive. Code for reference is here: https://gist.github.com/ungureanuvladvictor/40f69e0c3b12320e71ba.

denniszollo:
I think this occurs when you use a device that is already being used (i.e
you run two instances of your code, or the former one did not exit yet or
close the serial port)

ungureanuvladvictor:
Happens even after a restart so it is not from there.

PR #194, opened 2015-07-05
Field constructor update for non-SBP structs.

### LINKED ISSUE #192: Materialize non-SBP structs used in nested messages
We need to be able to materialize structs nested in SBP messages, like MsgObs. Currently, we don't generate that code, enough though we could. See: 
https://github.com/swift-nav/libsbp/blob/master/python/sbp/observation.py#L27

mookerji:
Creating non-SBP message structs still required binary payloads to be
created. Updates code generation template so non-SBP constructor
pattern follows SBP constructor pattern.

Closes #192.

/cc @mfine


mfine:
Sweet!

PR #195, opened 2015-07-05
Make SBP and SBP-struct object attributes static.

### LINKED ISSUE #95: SBP object implementation can't use slots
http://bugs.python.org/issue15535

As far as I can tell, if you implement a class with slots, the python pickling library will have difficulties, looking for an implicit `__dict__` attribute when none exists. Incidentally, you can add a `__dict__` attribute try and fool python, but then it'll go ahead and call `dict()` anyway (which sucks). In any case, whenever we've moved off using picking serialization for anything, we ought to use `__slots__` for the memory help, since it may actually be useful here for reducing memory usage of serial_link and the testing stuff.

For an implementation example, see: https://github.com/construct/construct/blob/master/construct/core.py#L100

mookerji:
This defines a __slots__ attribute for each SBP message (and non-SBP
struct) object, which fixes memory for a fixed set of attributes
instead of allocating a dynamic __dict__. Reduces memory overhead on
using the libsbp's Python bindings when we're instantiating 1k's of
objects. Anecdotal 7x decrease in heap usage:

Partition of a set of 240058 objects. Total size = 117214528 bytes.
 Index  Count   %     Size   % Cumulative  % Kind (class / dict of class)
     0 100000  42 104800000  89 104800000  89 dict of sbp.navigation.MsgBaselineECEF
     1 100000  42  6400000   5 111200000  95 sbp.navigation.MsgBaselineECEF

Partition of a set of 140393 objects. Total size = 21265240 bytes.
 Index  Count   %     Size   % Cumulative  % Kind (class / dict of class)
     0 100000  71 15200000  71  15200000  71 sbp.navigation.MsgBaselineECEF

Closes #95.

/cc @mfine @fnoble

### DIFF
--- a/python/sbp/msg.py
+++ b/python/sbp/msg.py
@@ -11,6 +11,7 @@
 
 from construct import *
 import base64
+import copy
 import json
 import struct
 
@@ -74,6 +75,12 @@ class SBP(object):
                    ULInt8('length'),
                    Bytes("payload", lambda ctx: ctx.length),
                    ULInt16('crc'),)
+  __slots__ = ['preamble',
+               'msg_type',
+               'sender',
+               'length',
+               'payload',
+               'crc']
 
   def __init__(self, msg_type=None, sender=None,
                length=None, payload=None, crc=None):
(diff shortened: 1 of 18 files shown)


python/sbp/msg.py
```diff
@@ -85,7 +92,17 @@ def __init__(self, msg_type=None, sender=None,
...
+      if self is other:
+        return True
+      elif not isinstance(self, type(other)):
+        return False
+      elif self.__slots__ != other.__slots__:
+        return False
+      else:
+        return all(getattr(self, s) == getattr(other, s) for s in self.__slots__)
+    except AttributeError:
+      return False
```
mfine: wow, this is a pretty involved check!

mfine:
Really cool - seems like a great perf win! :shipit:

PR #196, opened 2015-07-06
Configurable, timed, rotating JSON/SBP logger.

mookerji:
A low overhead, cross-platform way of keeping rotating logs over a
recent time interval, which seems like a common diagnostic use case
for field testing or for customers sending diagnostic logs.

Requires some real integration testing.

/cc @mfine @fnoble @denniszollo

### DIFF
--- a/python/sbp/client/loggers/rotating_logger.py
+++ b/python/sbp/client/loggers/rotating_logger.py
@@ -1,0 +1,100 @@
+# Copyright (C) 2015 Swift Navigation Inc.
+# Contact: Bhaskar Mookerji <mookerji@swiftnav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+from logging.handlers import TimedRotatingFileHandler
+from sbp.client.loggers.json_logger import JSONLogger
+from sbp.table import dispatch
+import Queue
+import logging
+import threading
+import time
+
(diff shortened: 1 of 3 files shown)


mookerji:
Hold off on merging this. I'm going to make the handler async so it doesn't screw with the console.

cbeighley:
Awesome, this is going to be really useful.

python/sbp/client/loggers/rotating_logger.py
```diff
@@ -0,0 +1,77 @@
...
+  ----------
+  filename : string
+    Path to file to write SBP messages to.
+  when : str
+    Specifies a type of interval
+  interval : int
+    Specifies length of interval
+  backupCount : int
+    Number of backups to keep around
+  mode : str
```
mfine: This `mode` parameter is not being used and should be removed.

python/sbp/client/loggers/rotating_logger.py
```diff
@@ -0,0 +1,77 @@
...
+    File mode
+  tags : dict
```
mfine: The `tags` parameter is not being used and should be removed.

mfine: Ahh, the tags is getting used by the parent class and its `fmt_msg` call. Nm.

python/sbp/client/loggers/rotating_logger.py
```diff
@@ -0,0 +1,77 @@
...
+  def flush(self):
+    self.handler.flush()
+
+  def close(self):
+    self.handler.close()
+
+  def call(self, msg):
+    try:
+      self.logger.info(json.dumps(self.fmt_msg(msg), allow_nan=False))
+    except ValueError:
```
mfine: Man, maybe we should move this error handling logic into a dumps message on the base class? I think we're starting to repeat it in a couple of places. There's got to be a better way to just change the the implementation of how the string goes out here.

mfine:
Looks cool - just a couple of nits to improve things.

denniszollo:
It might also be cool to have this break the files due to some external event (like a thread event).  This way the console might be able to manually break files when needed.

python/sbp/client/loggers/rotating_logger.py
mookerji: Done

mfine:
Nice, lgtm

PR #198, opened 2015-07-06
python: Catch exception in each callback independently.

gsmcmullin:
Without this change, an exception in on handler could prevent others
from being called.

/cc @mfine


mfine:
We have another catch above this that also does a break out of the loop for `SystemExit` - if this would swallow that (which it seems like), we should consolidate the checks:

https://github.com/swift-nav/libsbp/blob/master/python/sbp/client/handler.py#L141-145

But yeah, sounds great that concern notwithstanding.

mfine:
Fixed in https://github.com/swift-nav/libsbp/pull/203.

PR #199, opened 2015-07-06
libsbp Python sender ID is 0x42.

mookerji:
/cc @mfine


mfine:
Looks good - do we want to call it just `SENDER_ID`?  We had a bad experience with `HOST` before.

mfine:
Also, let's add something for the c header file.

mookerji:
Ditto and ditto

PR #202, opened 2015-07-06
add note about what sender id of 0x42 represents

denniszollo:
@mfine  What do you think about adding this to the spec to note what 0x42 represents?

### DIFF
--- a/spec/yaml/swiftnav/sbp/settings.yaml
+++ b/spec/yaml/swiftnav/sbp/settings.yaml
@@ -39,7 +39,9 @@ definitions:
           type: string
           desc: |
             A NULL-terminated and delimited string with contents
-            [SECTION_SETTING, SETTING, VALUE].
+            [SECTION_SETTING, SETTING, VALUE]. A device will only
+            process to this message when it is received from
+            sender ID 0x42.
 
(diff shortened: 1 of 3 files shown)


mfine:
Maybe something more like ```By default, clients of `libsbp` use a sender id value of `0x42````

spec/yaml/swiftnav/sbp/settings.yaml
```diff
@@ -39,7 +39,9 @@ definitions:
           type: string
           desc: |
             A NULL-terminated and delimited string with contents
-            [SECTION_SETTING, SETTING, VALUE].
+            [SECTION_SETTING, SETTING, VALUE]. A device will only
+            react to this message when it is received from
```
mfine: /react to/process/

mfine:
Couple of nits. Also do the file_io messages?

mookerji:
@denniszollo can you rebase this so we can merge?

denniszollo:
@mookerji - should be rebased now.  There used to be notes about how you know when you get to the "end" of a list.  Since those were removed previously, I left them out, but would actually prefer if we kept them in as it is useful information.

PR #204, opened 2015-07-09
Unfixed strings, remove null termination

mfine:
Remove the fixed strings in fileio.  Remove the null termination description in settings.

/cc @fnoble @gsmcmullin @mookerji @denniszollo

### DIFF
--- a/python/sbp/file_io.py
+++ b/python/sbp/file_io.py
@@ -47,28 +47,33 @@ from a given offset into a file, and returns the data in a
 MSG_FILEIO_READ_RESPONSE message where the message length field
 indicates how many bytes were succesfully read. If the message is
 invalid, a followup MSG_PRINT message will print "Invalid fileio
-read message".
+read message". The sequence number in the request will be returned
+in the response.
 
 
   Parameters
   ----------
   sbp : SBP
     SBP parent object to inherit from.
+  sequence : int
+    Read sequence number
   offset : int
     File offset
   chunk_size : int
     Chunk size to read
   filename : string
-    Name of the file to read from (NULL padded)
+    Name of the file to read from
   sender : int
     Optional sender ID, defaults to SENDER_ID (see sbp/msg.py).
 
   """
   _parser = Struct("MsgFileioReadRequest",
+                   ULInt32('sequence'),
                    ULInt32('offset'),
                    ULInt8('chunk_size'),
-                   String('filename', 20),)
+                   greedy_string('filename'),)
   __slots__ = [
+               'sequence',
                'offset',
                'chunk_size',
                'filename',
@@ -84,6 +89,7 @@ read message".
       super( MsgFileioReadRequest, self).__init__()
       self.msg_type = SBP_MSG_FILEIO_READ_REQUEST
       self.sender = kwargs.pop('sender', SENDER_ID)
+      self.sequence = kwargs.pop('sequence')
       self.offset = kwargs.pop('offset')
       self.chunk_size = kwargs.pop('chunk_size')
       self.filename = kwargs.pop('filename')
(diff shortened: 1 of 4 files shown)


mfine:
This is missing the construct changes to support greedy strings. How do these C-structs looks WRT intended implementation?

mfine:
@fnoble what are your thoughts on removing the NULL-termination on settings and file_io messages? Right now, the code is checking everywhere for it and it will likely hurt a bit of backwards compat to drop it.

mfine:
@fnoble @gsmcmullin proposed new structure with a `sequence` field - comments, name field name suggestions welcome:

```diff
--- a/spec/yaml/swiftnav/sbp/file_io.yaml
+++ b/spec/yaml/swiftnav/sbp/file_io.yaml
@@ -37,6 +37,9 @@ definitions:
       invalid, a followup MSG_PRINT message will print "Invalid fileio
       read message".
     fields:
+      - sequence:
+          type: u32
+          desc: Sequence number for identifying request and response
       - offset:
           type: u32
           units: bytes
@@ -58,16 +61,12 @@ definitions:
       message where the message length field indicates how many bytes
       were succesfully read.
     fields:
-      - offset:
+      - sequence:
           type: u32
-          units: bytes
-          desc: File offset
-      - chunk_size:
-          type: u8
-          units: bytes
-          desc: Chunk size read
-      - filename_contents:
-          type: string
+          desc: Sequence number identifying request and response
+      - contents:
+          type: array
+          fill: u8
           desc: |
             Name of the file read from separated by NULL from its contents
```

mfine:
@fnoble @gsmcmullin as per discussion, the fileio messages have been updated to introduce a `sequence` number up front that gets echo'd back on responses. What do you guys think of these changes? No settings changes are needed!

python/sbp/file_io.py
```diff
@@ -47,28 +47,33 @@ class MsgFileioReadRequest(SBP):
...
   sender : int
     Optional sender ID, defaults to SENDER_ID (see sbp/msg.py).
 
   """
   _parser = Struct("MsgFileioReadRequest",
+                   ULInt32('sequence'),
                    ULInt32('offset'),
                    ULInt8('chunk_size'),
-                   String('filename', 20),)
+                   greedy_string('filename'),)
```
fnoble: Does this get changed to `GreedyString` when your Construct PR makes it into upstream?

spec/yaml/swiftnav/sbp/file_io.yaml
```diff
@@ -35,8 +35,12 @@ definitions:
       MSG_FILEIO_READ_RESPONSE message where the message length field
       indicates how many bytes were succesfully read. If the message is
       invalid, a followup MSG_PRINT message will print "Invalid fileio
```
fnoble: Do you think we should remove the documentation around the `MSG_PRINT` response? Even if its true right now we don't want people to rely on it as part of the spec.

fnoble: Ditto for the other instances

fnoble:
Like the sequence number! This looks good, seems to be missing updated C headers (and PDF, but I'm not sure if we have been generating that every time?)

spec/yaml/swiftnav/sbp/file_io.yaml
gsmcmullin: I agree, it shouldn't be part of the spec.

python/sbp/file_io.py
mfine: Yeah, if/when it gets pulled in to upstream, we can merge pull `greedy_string`

spec/yaml/swiftnav/sbp/file_io.yaml
mfine: Definitely. Removing.

mfine:
Woops, missed the C generation. Added. Addressed other comments as well. Haven't been generating the doc every run. We'll regenerate around the release.

fnoble:
lgtm!

PR #205, opened 2015-07-09
Greedy strings

mfine:
Replace newline-terminated strings with greedy strings.

/cc @mookerji @fnoble


mfine:
Going to generate code for this now.

PR #207, opened 2015-07-10
Remove the janky 0xFF terminator on read dir requests

mfine:
Pull out the 0xFF terminator. End of dir entries will come via an an empty response.

/cc @gsmcmullin


gsmcmullin:
It's probably worth mentioning in the packet description that if `offset` is greater or equal to the number of files in directory and empty response will be generated.

Issue #167, opened 2015-06-16
Promote Logging to Stable


mfine:
To reach SBP version 1.0, we should promote the logging to stable. Namely, some logging that support debug, info, error, etc. Should deprecate MSG_PRINT and remove MSG_DEBUG_VAR.

/cc @mookerji @fnoble

mfine:
What do we think about deprecating MSG_PRINT (which uses that beastly string with newline terminator) and move to MSG_LOG which includes a level?

mookerji:
+1 so hard

denniszollo:
Would this be treating the first CHAR in the string as the log level and just replacing the macro in the firmware, or something more complete?

Also, we should definitely also add the console filter for log level if we do this.

mfine:
The macro is in libswfitnav, and it's putting a string up front. How about we move from printf to a log function which then takes a level (byte, short, or int) and encodes a message with a string and a int. Unfortunately, this would be pretty piksi_firmware oriented - not sure how to incoporate that with libswiftnav. @fnoble thoughts?

denniszollo:
How about we use the preprocessor with a define ARM or something through the makefile? (makefile should now when it's compiling libswiftnav for a piksi target, right?)

mfine:
We could do something like that if we thought it was ok and not too obscure - point `log_warn(str` etc, at `log(WARN, str` on piksi builds, and the string prepending elsewhere.

mfine:
@mookerji For backwards compatibility, I'm thinking of adding a newlinestring to the YAML. Just for these print messages.

mookerji:
I'm in favor of having something like 

```yaml
 - MSG_LOG:
    id: 0x0203
    public: True
    short_desc: Bar
    desc: Foo
    fields:
          - level:
              type: u8
              desc: Logging level flag
              fields: 
                - 4-7:
                    desc: Reserved
                - 0-3:
                    desc: Logging level
                    values:
                        - 0: INFO
                        - 1: DEBUG
                        - 2: TRACE
          - text:
              type: string
              desc: Human readable string
```

Also, "newlinestring to the yaml"?

fnoble:
[...]

PR #208, opened 2015-07-10
MSG_LOG

mfine:
Introduce MSG_LOG, per #167. Note the values, which match the current logging levels in [libswiftnav] (https://github.com/swift-nav/libswiftnav/blob/master/include/libswiftnav/logging.h#L37-52). Also note the maskable id - is this where we want log messages, or do we want to give them their own group (0x0500?).

/cc @fnoble @denniszollo @mookerji @gsmcmullin

### DIFF
--- a/spec/yaml/swiftnav/sbp/logging.yaml
+++ b/spec/yaml/swiftnav/sbp/logging.yaml
@@ -19,7 +19,7 @@ include:
 definitions:
 
  - MSG_LOG:
-    id: 0x0102
+    id: 0x0401
     short_desc: Plaintext logging messages with levels
     desc: |
       This message contains a human-readable payload string from the
--- a/spec/yaml/swiftnav/sbp/logging.yaml
+++ b/spec/yaml/swiftnav/sbp/logging.yaml
@@ -18,6 +18,31 @@ include:
   - types.yaml
 definitions:
 
+ - MSG_LOG:
+    id: 0x0102
+    short_desc: Plaintext logging messages with levels
+    desc: |
+      This message contains a human-readable payload string from the
+      device containing errors, warnings and informational messages at
+      ERROR, WARNING, DEBUG, INFO logging levels.
+    fields:
+      - level:
+          type: u8
+          desc: Logging level
+          fields:
+              - 3-7:
+                    desc: Reserved
+              - 0-2:
+                    desc: Logging level
+                    values:
+                        - 0: DEBUG
+                        - 1: INFO
+                        - 2: WARN
+                        - 3: ERROR
+      - text:
+          type: string
+          desc: Human-readable string
+
  - MSG_PRINT:
     id: 0x0010
     short_desc: Plaintext logging messages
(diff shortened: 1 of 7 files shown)


mookerji:
lgtm!

fnoble:
awesome, lgtm!

mookerji:
To clarify, I don't have a preference much for the message masking for the non-obs type stuff. It's not obvious to me that they need to be treated any differently than the other things that don't go over the radio.

fnoble:
oh, didn't consider the masking - let me have a think about that first. I think 0x01xx was already taken for a particular group.

mfine:
Anyone else have preferences on the logging message mask stuff? @henryhallam? @gsmcmullin? @cbeighley? Reproducing Dennis' concern here:

*So I think that the MSG_LOG should have it's own group in the messages (i.e away from MSG_UART_STATE, MSG_THREAD_STATE). I sort of see that the product will only use this message if something pretty bad is happening for which we might need user awareness and user interaction.*

*I'm thinking a message ID of 0x110, 0x510 might work well, that way it comes with basic system event info (i.e. the external event trigger), the old logging group, and a new msg_log only bit.*

*This way you could get the warnings and errors on the device without all the thread state overhead.*

fnoble:
So far we have the following groups, roughly speaking:
```
0000 00GG xGxx xxxx
       ||  | over the radio (0x40)
       ||--- time / event related messages
       |---- position / velocity related messages
```
G - group indicator bit
0 - unallocated group bit
x - sub-group indicator bit

We should probably establish more of a policy around this soon.

fnoble:
With that in mind though, as seeing as this is being stabilized, I'd propose using 0x0401 and allocating the 11th bit of the ID for such informational kinds of messages.

spec/yaml/swiftnav/sbp/logging.yaml
```diff
@@ -12,20 +12,33 @@ package: swiftnav.sbp.logging
 description: |
   Logging and debugging messages from the device. These are in the
   implementation-defined range (0x0000-0x00FF).
-stable: False
+stable: True
```
fnoble: Should stable be specified for the whole package? Will this make the deprecated message stable?

mfine: The stability will not apply to deprecated messages. I need to update that here and in a couple of other places. But, yes, `MSG_TWEET` will be stable.

fnoble: Is stability only indicated on a per package basis?

mfine:
Yeah, it's only been indicated on a per package basis so far. I added per message type for public, we could do that for stability as well.

henryhallam:
Stability should be per message.
0x0401 sounds good to me; we should move to a whitelist system rather than the message mask in the not too distant future.

denniszollo:
I like 0x410, but think there might be more accidental implicit groups than the one's Fergus outlined.  Last time I looked it looked like the first few nibbles could be used and have some patterns.

mfine:
0x401 or 0x410? Going to change and bring in

mfine:
retest this please

PR #209, opened 2015-07-10
Add a wait_callback

mfine:
Like wait, but takes a callback.

/cc @cbeighley

### DIFF
--- a/python/sbp/client/handler.py
+++ b/python/sbp/client/handler.py
@@ -280,7 +280,7 @@ class Handler(object):
     """
     self.framer.write(msg.to_binary())
 
-  def wait(self, msg_type, timeout):
+  def wait(self, msg_type, timeout=1.0):
     """
     Wait for a SBP message.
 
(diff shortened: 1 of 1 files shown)


cbeighley:
What do you think about having an optional callback arg for `wait` instead?

mfine:
Not sure how that would work or look? The one returns a value, the other doesn't - how would the composite one work? All optional arguments and return the payload if there was a result? Something like:

```python
  def wait_callback(self, callback=None, msg_type=None, timeout=1.0):
    """
    Wait for a SBP message with a callback.

    Parameters
    ----------
    callback : fn
      Callback function
    msg_type : int | iterable
      Message type to register callback against. Default `None` means global callback.
      Iterable type adds the callback to all the message types.
    timeout : float
      Waiting period
    """
    event = threading.Event()
    payload = {'data': None}
    def cb(sbp_msg):
      payload['data'] = sbp_msg
      if callback is not None:
        callback(sbp_msg)
      event.set()
    self.add_callback(cb, msg_type)
    event.wait(timeout)
    return payload['data']
```

PR #213, opened 2015-07-13
SBP misc. cleanup.

mookerji:
When materializing SBP messages without fields, the parent payload
field was uninitialized, which threw an exception while packing binary
for the serial buffer.

/cc @fnoble @mfine


mfine:
Nice! LGTM

PR #217, opened 2015-07-15
Refactor ideas [DO NOT MERGE]

mfine:
!!!DO NOT MERGE!!!

There's a better way to structure our client code to make it more composeable. Right now, so many things are getting baked in together. But I think if break out the important pieces, we can make it a lot easier to extend things. It seems like we have the notion of sources (read) and sinks (writes) (some things are both!), and then wrappers around sources and sinks for doing specialized operations, like json encoding and decoding / pickle encoding and decoding, etc. What follows is some loose sketching of code to start heading that way - thoughts?

/cc @mookerji

### DIFF
--- a/python/sbp/client/pyserial_handle.py
+++ b/python/sbp/client/pyserial_handle.py
@@ -1,0 +1,29 @@
+# Copyright (C) 2015 Swift Navigation Inc.
+# Contact: Mark Fine <mark@swiftnav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+import serial
+
+def PySerialHandle(object):
+  """
+  PySerialSource
+
+  The :class:`PySerialSource` class reads SBP messages from a serial port
+  using the pyserial source.
+
+  Parameters
+  ----------
+  port : string
+    Path to port to read SBP messages from.
+  baud : int
+    Baud rate of serial port.
+  timeout : int | float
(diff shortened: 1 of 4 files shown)


python/sbp/client/source.py
```diff
@@ -0,0 +1,60 @@
...
+# Contact: Mark Fine <mark@swiftnav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+class Source(object):
```
mookerji: How do we handle IO sources that don't necessarily have file-like APIs? Are we going to have to define our own? I'm thinking of sources and sinks that might have a `flush()` method.

python/sbp/client/pyserial_handle.py
```diff
@@ -0,0 +1,29 @@
...
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+import serial
+
+def PySerialHandle(object):
```
mookerji: Is this intended to be a subclass of Source/Sink?

python/sbp/client/source.py
mfine: Yeah, good question - I'm not even sure we should have these source/sink classes. Ideally, there would be an interface that each thing requires and then it would consume objects that met those interfaces... I'm going to try and catalog all the things we're doing in SBP now to figure out what we can reduce down to.

mfine:
Closing this out in favor of an iterator approach.

PR #219, opened 2015-07-15
Don't pull CRC field from JSON.

gsmcmullin:
The CRC field is missing from the Android JSONLogger and it's easier to fix it here.

/cc @mookerji @mfine

### DIFF
--- a/python/sbp/msg.py
+++ b/python/sbp/msg.py
@@ -169,7 +169,6 @@ class SBP(object):
     sbp.sender = d.pop('sender')
     sbp.length = d.pop('length')
     sbp.payload = base64.standard_b64decode(d.pop('payload'))
-    sbp.crc = d.pop('crc')
     return sbp
 
   def to_json_dict(self):


python/sbp/msg.py
```diff
@@ -169,7 +169,6 @@ def from_json_dict(d):
     sbp.sender = d.pop('sender')
     sbp.length = d.pop('length')
     sbp.payload = base64.standard_b64decode(d.pop('payload'))
-    sbp.crc = d.pop('crc')
```
mfine: ???

mfine: If you want to ignore a `crc` that's not there:

```python
     sbp.crc = d.pop('crc', None)
```

mookerji:
Can we close this PR?

gsmcmullin:
I don't really care about getting it merged, but I don't think the crc should be recorded in the json log, and the log iterator shouldn't barf it it's not present.

PR #226, opened 2015-07-20
Preserve the logs when there are nans

mfine:
When the firmware generates bad data that makes it to the wire, preserve the logs.

/cc @mookerji


mfine:
retest this please

PR #227, opened 2015-07-21
added raim flags

kovach:


mfine:
Out of curiousity and ignorance, is this a redundant place to put this kind of status? (seems like the convention for it - just curious)

mfine:
(looking good otherwise)

henryhallam:
Looks good.  Since this is the first use of flag bits in this field beyond the two LSbits, we'd better go over anything in piksi_tools, log analysis etc that checks the flags field and make sure it's masking appropriately, not comparing the whole field to 0, 1 or 2.

Off-topic: Why do we have 3 bits for fix type?

@mfine not sure what you mean?

cbeighley:
Do we need a RAIM enabled/disabled flag in the message? I think its probably good enough to just have it in the settings.

henryhallam:
The availability flag doesn't just show whether RAIM is enabled/disabled, because RAIM can be enabled but unavailable due to number of satellites.  It's traditional to have an availability indicator separate from the repair flag, so as to convey the three integrity states "solution was OK, solution was not OK, solution integrity unknown."

kovach:
yes, currently I am conflating 'disabled' with 'unavailable because of satellite count.' is this ok? it could be partially disambiguated with the n_sats field.

henryhallam:
IMO it's fine. I like the flags as you have them.

mookerji:
lgtm! Thanks for doing this.

PR #228, opened 2015-07-23
python: Iterator interface implementation.

gsmcmullin:
- All sources of SBP messages are iterators of tuple(delta, time, msg).
- All sinks of messages are callable with args (delta, time, msg).
- Send and receive paths decoupled.
- SBP binary streams use `FrameReceiver` and `FrameSender` as source and 
  sink, respectively.
- `ReceiveHandler` provides plumbing on the receive path, connects a single
  source to many sinks with queueing and filtering.  It also provides
  the old-syle `add_callback` interface
- A `Forwarder` class is provided to connect a source to a sink for logging
  or forwarding.  This should be used rather than `ReceiveHandler.add_callback()`
  in the case where the sink may block or take time to process a message.

/cc @mfine @mookerji @denniszollo


mookerji:
*Great* I'll have a bit of time to review this later today.

python/sbp/client/examples/pusher_client.py
```diff
@@ -65,13 +68,12 @@ def connect_it(data):
       push_client.connection.bind('pusher:connection_established', connect_it)
       push_client.connect()
 
-      handler.add_callback(push_it, [SBP_MSG_OBS, SBP_MSG_BASE_POS, SBP_MSG_OBS_DEP_A])
-
       try:
-        while True:
-          time.sleep(0.1)
+        for delta, time, msg in handler.filter(OBS_MSG_LIST):
```
mfine: Should we kill the tuple usage here and above? Passing in 0's is super janky and using the tuples ends up being super inflexible. A newer iterator introduced a `metadata` [column](https://github.com/swift-nav/libsbp/blob/master/python/sbp/client/loggers/json_logger.py#L143), which then becomes a nightmare to thread around everywhere - potentially an envelope around the msg makes more sense to pass around? @mookerji ?

gsmcmullin: I'm going to change the convention to be a 2-tuple of the message and a dictionary of metadata that may contain delta and timestamp if these are available, as well as other fields that we may add in future.  I'll also change the call signature to take the metadata as **kwargs.

mookerji: Moving a dict instead of a tuple is a good idea. +1

python/sbp/client/handler.py
```diff
@@ -293,7 +209,7 @@ def wait(self, msg_type, timeout=1.0):
     """
     event = threading.Event()
     payload = {'data': None}
-    def cb(sbp_msg):
+    def cb(d, t, sbp_msg):
```
mfine: This is still taking delta, timestamp.

mfine:
One nit, but looks good!

python/sbp/client/handler.py
```diff
@@ -316,54 +232,37 @@ def wait_callback(self, callback, msg_type=None, timeout=1.0):
...
-      Number of bytes to read.
+    Class for upstream iterators.  Implements callable interface for adding
+    messages into the queue, and iterable interface for getting them out.
     """
-    data = self.driver.read(size)
-    if data is not None:
-      self.logger.write(data)
-    self.write()
-    return data
+    def __init__(self, maxsize):
```
mookerji: Can we have a default configuration for the Queue? Like 0 or -1 or something.

python/sbp/client/forwarder.py
```diff
@@ -8,25 +8,25 @@
...
 
-import struct
+from threading import Thread
 
-from .base_logger import BaseLogger
+class Forwarder(Thread):
+  def __init__(self, source, sink):
+    Thread.__init__(self, name="SBP Forwarder")
+    self.daemon = True
+    self._broken = False
```
mookerji: nit: doc string for source and sink within Forwarder?

python/sbp/client/forwarder.py
```diff
@@ -8,25 +8,25 @@
...
 
-  The :class:`ByteLogger` logs SBP messages to bytes.
-  """
-  def __call__(self, msg):
-    self.call(msg)
+  def stop():
+    self._broken = True
+    try:
+      self._source.breakiter()
+    except: pass
```
mookerji: nit: newline

python/sbp/client/framer.py
```diff
@@ -0,0 +1,135 @@
...
+    # crc
+    crc = self._readall(2)
+    crc, = struct.unpack("<H", crc)
+    if crc != msg_crc:
+      print "crc mismatch: 0x%04X 0x%04X" % (msg_crc, crc)
+      return None
+    msg = SBP(msg_type, sender, msg_len, data, crc)
+    try:
+      msg = self._dispatch(msg)
+    except: pass
```
mookerji: nit: newline

python/sbp/client/handler.py
```diff
@@ -167,6 +60,38 @@ def __enter__(self):
   def __exit__(self, *args):
     self.stop()
 
+  # This exception is raised when a message is dispatched to a garbage
+  # collected upstream iterator.
+  class _DeadCallbackException(Exception): pass
```
mookerji: nit: newline

python/sbp/client/loggers/json_logger.py
```diff
@@ -76,11 +72,9 @@ def next(self):
     for line in self.handle:
       try:
         data = json.loads(line)
-        delta = data['delta']
-        timestamp = data['timestamp']
-        item = SBP.from_json_dict(data['data'])
+        item = SBP.from_json_dict(data.pop('data'))
         msg = self.dispatch(item, line)
-        yield (delta, timestamp, msg)
+        yield (msg, data)
```
mookerji: Do we want to replace the tuple here with a dict?

python/sbp/client/handler.py
mookerji: Why is this defined as a subclass?

python/sbp/client/loggers/json_logger.py
```diff
@@ -135,12 +129,9 @@ def next(self):
...
 
     for data in sorted(datas, key=itemgetter('timestamp')):
-      delta = data['delta']
-      timestamp = data['timestamp']
-      metadata = data['metadata']
-      item = SBP.from_json_dict(data['data'])
+      item = SBP.from_json_dict(data.pop('data'))
       msg = self.dispatch(item)
-      yield (delta, timestamp, metadata, msg)
+      yield (msg, data)
```
mookerji: Same: dict instead of tuple?

python/sbp/client/loggers/rotating_logger.py
```diff
@@ -16,41 +16,8 @@
 import threading
 import time
 
-# TODO (Buro): Replace _mk_async_emit with a function using from
-# logutils.queue: QueueHandler, QueueListener.
 # TODO (Buro): Add a real benchmark after integration testing.
```
mookerji: nit: remove this comment.

mookerji: nit: remove this comment.

python/sbp/client/handler.py
gsmcmullin: Thanks.  I'll fix it.

python/sbp/client/handler.py
gsmcmullin: This class is not intended to be public.  The default value of an unbounded queue is in Handler.filter()

python/sbp/client/loggers/json_logger.py
gsmcmullin: It's a tuple of a message and a dictionary of metadata.  The same as produced by Framer and Handler iterators.

python/sbp/client/handler.py
gsmcmullin: I'm not sure what you're asking.  It's an internal exception that is an implementation detail to allow dead iterators to be garbage collected.  It's not intended for public use.

python/sbp/client/drivers/network_drivers.py
```diff
@@ -16,8 +16,6 @@
 # TODO (Buro): Consider making a basic, standard library async tcp
 # client: http://pymotw.com/2/asyncore/.
 
-DEFAULT_RECV_SIZE = 1024
-
 class TCPDriver(BaseDriver):
   """TCPDriver
```
mookerji: Have you tested to see if this TCP driver and serial link can work together for exchanging observations?

python/sbp/client/forwarder.py
```diff
@@ -0,0 +1,51 @@
...
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+from threading import Thread
+
+class Forwarder(Thread):
```
mookerji: Can a Forwarder be composed with its reverse to duplex observations? It doesn't seem so. I want to be able to preserve a capability for reading/writing from the serial port and a network connection (TCP/HTTP, etc.).

python/sbp/client/handler.py
mookerji: Sorry, I meant to say "nested class". The alternative would be to have it defined, not nested, at the top of the file with an underscored name. I don't think nested classes are that common. I think any caller outside this function will have to reference both the module path and the class to get at the exception.

python/sbp/client/handler.py
```diff
@@ -14,151 +14,44 @@
...
   """
-  def __init__(self, read, write, verbose=False):
-    self.callbacks = collections.defaultdict(set)
-    self.framer = Framer(read, write, verbose)
-    self.receive_thread = ReceiveThread(self.framer.receive, self.call)
+  def __init__(self, source):
+    self._source = source
+    self._callbacks = collections.defaultdict(set)
+    self._receive_thread = threading.Thread(target=self._recv_thread, name="Handler")
+    self._receive_thread.daemon = True
```
mookerji: Can we have transmitting thread too?

python/sbp/client/handler.py
```diff
@@ -167,6 +60,39 @@ def __enter__(self):
...
+    # We use a weakref so that the iterator may be garbage collected if it's
+    # consumer no longer has a reference.
+    ref = weakref.ref(iterator)
+    self._sinks.append(ref)
+    def feediter(msg, **metadata):
+      i = ref()
+      if i is not None:
+        i(msg, **metadata)
+      else:
+        raise Handler._DeadCallbackException
```
mookerji: `Handler._DeadCallbackException()`?

python/tests/sbp/client/test_handler.py
```diff
@@ -105,26 +84,71 @@ def test_handler_callbacks():
...
+    handler.start()
+  thread = threading.Thread(target=latestart)
+  thread.start()
+  iter(handler) # Dead iterator, should be gc'd
+  xx = iter(handler)
+  thread.join()
+  handler._receive_thread.join()
+  gc.collect()
+  assert [sink() for sink in handler._sinks] == [xx]
+
```
mookerji: You probably want to have some test coverage for this dead callback exception stuff.

python/sbp/client/loggers/rotating_logger.py
```diff
@@ -77,18 +42,18 @@ class RotatingFileLogger(JSONLogger):
   """
 
   def __init__(self, filename, when='M', interval=30, backupCount=3,
-               tags={}, dispatcher=dispatch):
+               tags={}, dispatcher=None):
```
mookerji: Why is dispatcher here now defaulting to `None`? I think this needs to be defined for the parent class, which will swallow an exception and only log undispatched SBP because the message dispatch table is `None`.

python/sbp/client/forwarder.py
gsmcmullin: You can make two forwarders, one for each direction:
```
  with Serial(args.port[0], 1000000) as driver:
    # Create a handler to connect our Piksi driver to our callbacks
    with Handler(Framer(driver.read, None, verbose=True)) as handler:

      tcpdriver = TCPDriver('localhost', 2000)
      tcpframer = Framer(tcpdriver.read, tcpdriver.write, verbose=True)
      Forwarder(handler, tcpframer).start()
      Forwarder(tcpframer, handler).start()
```

python/sbp/client/loggers/base_logger.py
```diff
@@ -25,7 +25,7 @@ class BaseLogger(object):
   filename : string
     File to log to.
   """
-  def __init__(self, filename, mode="w", tags={}, dispatcher=dispatch):
+  def __init__(self, filename, mode="w", tags={}, dispatcher=None):
     self.handle = open(filename, mode)
     self.dispatcher = dispatcher
```
mookerji: Yeah, what's the motivation for moving the message dispatch default to `None`?

It's just occurred to me that message dispatch could actually just another thing we compose in the current calling chain instead of passing as an argument. Something like: `Handler(Dispatcher(Framer(driver.read, driver.write)))`. @mfine, thoughts?

python/tests/sbp/client/test_handler.py
gsmcmullin: This `test_dead_gc` does cover the dead callback exception.  The exception is an internal implementation detail in the removing of the weakref to the gc'd iterator.

python/sbp/client/forwarder.py
mookerji: Have you tested this? Does that mean that the handlers need mutexes ("muteces"?) on the resources they manage?

python/sbp/client/handler.py
```diff
@@ -14,151 +14,44 @@
...
+from Queue import Queue
 
 class Handler(object):
   """
   Handler
 
   The :class:`Handler` class provides an interface for connecting handlers
-  to a driver providing SBP messages.
+  to a driver providing SBP messages.  Also provides queued and filtered
+  iterators for synchronous, blocking use in other threads.
```
mookerji: Can we also have a asynchronous, nonblocking handling of messages? If the handler throws them on the queue, having an async callback that returns immediately would be useful in a few cases.

gsmcmullin: The old callback mechanism is still available.  I'll correct the comment.

python/sbp/client/loggers/base_logger.py
```diff
@@ -25,7 +25,7 @@ class BaseLogger(object):
   filename : string
     File to log to.
   """
-  def __init__(self, filename, mode="w", tags={}, dispatcher=dispatch):
+  def __init__(self, filename, mode="w", tags={}, dispatcher=None):
     self.handle = open(filename, mode)
```
mookerji: I thought there was some thought to replacing filename arguments in the loggers with handler (file, network, etc.) references. @mfine ?

python/sbp/client/forwarder.py
gsmcmullin: It works for me.  To what resources are you referring? What is the use case for using the driver other than through its handler?  If it really makes sense to do this then you may need a mutex, but it'd make more sense to include the mutex in the driver implementation.

python/sbp/client/loggers/base_logger.py
mfine: Yeah, that's a nice idea - a dispatcher thing that produces a rich message if dispatch succeeds.

PR #231, opened 2015-07-24
added include guard to common.h

kovach:
cc @mfine @mookerji


mfine:
:+1:

kovach:
ecstatic sees this as a redefinition without the #ifndef

PR #234, opened 2015-08-06
Fix module path includes.

mookerji:
Structs defined in other files break the python path, since swiftnav
top-level belongs to libswiftnav.

/cc @ungureanuvladvictor


ungureanuvladvictor:
tested, lgtm

PR #238, opened 2015-08-11
A Skylark Driver

mfine:
An initial skylark driver. Takes a `device-uid` to pull observations for. 

/cc @mookerji @fnoble

### DIFF
--- a/python/sbp/client/drivers/skylark_driver.py
+++ b/python/sbp/client/drivers/skylark_driver.py
@@ -1,0 +1,22 @@
+# Copyright (C) 2015 Swift Navigation Inc.
+# Contact: Bhaskar Mookerji <mookerji@swiftnav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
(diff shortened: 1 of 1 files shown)


mookerji:
That was fast.

python/sbp/client/drivers/skylark_driver.py
```diff
@@ -0,0 +1,33 @@
...
+
+  Parameters
+  ----------
+  device_uid : uid
+    Device unique id.
+
+  """
+  def __init__(self, device_uid):
+    url = 'http://skylark.swiftnav.com'
+    headers = {'Device-Uid': device_uid, 'Accept': 'application/sbp'}
```
mookerji: POST? Error handling and retries?

mfine: This is the client, receive side. Shoudl we setup the server, send side in the same driver?

mookerji: This is vaguely related to the refactoring question. We still need a client POST to the endpoint. We could expose a file handle interface for HTTP and stream through that.

PR #240, opened 2015-08-17
Remove watchdog

mfine:
Remove the software-based watchdog.

/cc @mookerji


mookerji:
I'm good for taking this out of `piksi_tools`. If you don't think this might be useful eventually, I'm good for removing it.

PR #235, opened 2015-08-07
Added signal_t struct and modified messages to use it.

ungureanuvladvictor:
\cc: @gsmcmullin @mookerji @mfine 

Signed-off-by: Vlad Ungureanu <vvu@vdev.ro>

PR #241, opened 2015-08-19
Split ephemeris_t into kepler/xyz counterparts.

ungureanuvladvictor:
\cc: @fnoble @mfine @gsmcmullin 

Signed-off-by: Vlad Ungureanu <vvu@vdev.ro>

### DIFF
--- a/c/include/libsbp/acquisition.h
+++ b/c/include/libsbp/acquisition.h
@@ -24,6 +24,7 @@
 #define LIBSBP_ACQUISITION_MESSAGES_H
 
 #include "common.h"
+#include "signal.h"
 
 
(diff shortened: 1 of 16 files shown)


mfine:
The split looks ok to me, but not saying much if it makes sense :) Nit is that this has #235 in it. Also, still not sure about the `sbp_` prefixing here - though I don't have a better alternative. Feels like if we do `sbp_`, we should consider `sbp_msg_` as well (as well as adding `sbp_` prefix for the other SBP substructs). /bikeshed

ungureanuvladvictor:
retest this please

JoshuaGross:
Assuming we can safely close this?

PR #254, opened 2015-09-10
JSON encodings and accompanying tests.

### LINKED ISSUE #253: JSON serialization for libsbp Haskell

mookerji:
/cc @mfine @kovach @fnoble

Closes #253.

### DIFF
--- a/haskell/src/SwiftNav/SBP/Bootload.hs
+++ b/haskell/src/SwiftNav/SBP/Bootload.hs
@@ -13,8 +13,10 @@
 
 module SwiftNav.SBP.Bootload where
 
+import BasicPrelude
 import Control.Monad
 import Control.Monad.Loops
+import Data.Aeson.TH (deriveJSON, defaultOptions, fieldLabelModifier)
 import Data.Binary
 import Data.Binary.Get
 import Data.Binary.IEEE754
@@ -23,6 +25,7 @@ import Data.ByteString
 import Data.ByteString.Lazy hiding ( ByteString )
 import Data.Int
 import Data.Word
+import SwiftNav.SBP.Encoding
 
 msgBootloaderHandshakeReq :: Word16
 msgBootloaderHandshakeReq = 0x00B3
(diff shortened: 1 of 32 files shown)


haskell/src/SwiftNav/SBP/Bootload.hs
```diff
@@ -11,10 +11,13 @@
 -- use only. Note that some of these messages share the same message type ID
 -- for both the host request and the device response.
 
+{-# LANGUAGE TemplateHaskell #-}
```
mfine: could move this into cabal file

haskell/src/SwiftNav/SBP/Encoding.hs
```diff
@@ -0,0 +1,33 @@
...
+-- Module:      SwiftNav.SBP.Encoding
+-- Copyright:   Copyright (C) 2015 Swift Navigation, Inc.
+-- License:     LGPL-3
+-- Maintainer:  Mark Fine <dev@swiftnav.com>
+-- Stability:   experimental
+-- Portability: portable
+--
+-- Encoding utilities.
+
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
mfine: can move this into the cabal file

haskell/src/SwiftNav/SBP/Bootload.hs
```diff
@@ -68,6 +72,8 @@ instance Binary MsgBootloaderHandshakeResp where
     putWord32le msgBootloaderHandshakeRespFlags
     putByteString msgBootloaderHandshakeRespVersion
 
+$(deriveJSON defaultOptions ''MsgBootloaderHandshakeResp)
```
mfine: out of curiosity, how does this formulate the field names - is it compatible with what we have in python? is it configurable?

haskell/src/SwiftNav/SBP/Bootload.hs
mookerji: Hmmmm, yes. Good idea.

haskell/src/SwiftNav/SBP/Encoding.hs
mookerji: Not sure. Perhaps we don't want warnings like this disabled across the entire project?

mfine: Are they per file options? I assumed it gets turned on during the compile if this file gets brought in for everyone, so I usually just go stick it in the cabal file.

mfine:
```NOTE: the contents of OPTIONS_GHC are prepended to the command-line options, so you do have the ability to override OPTIONS_GHC settings via the command line.

It is not recommended to move all the contents of your Makefiles into your source files, but in some circumstances, the OPTIONS_GHC pragma is the Right Thing. (If you use -keep-hc-file-too and have OPTION flags in your module, the OPTIONS_GHC will get put into the generated .hc file).```

haskell/src/SwiftNav/SBP/Bootload.hs
mookerji: Pretty sure it's just the identity (`'id`), but you can pass in some other function. See: https://hackage.haskell.org/package/aeson-0.9.0.1/docs/Data-Aeson-TH.html

PR #255, opened 2015-09-12
C: include files with libsbp in path

flixr:
In order to make it easier to directly use the C files in other projects, don't make it necessary to add all header files in include/libsbp to the include dirs....


mookerji:
This looks good to me. Thoughts @gsmcmullin @fnoble ?

ungureanuvladvictor:
Looks ok, keeps same style as libswiftnav of includes.

PR #257, opened 2015-09-16
Bring in the CRC16 module

mfine:
This is a crc16 module akin to what we have under the python and c modules.

/cc @mookerji


mfine:
Not ready yet - want to try and add some QC properties for this.

mfine:
Ok, I'm not sure how to setup qc properties for crc16 without a lot of work :( I don't know what I was thinking. Made some simple tests instead.

PR #258, opened 2015-09-16
Aeson specialization

mookerji:
/cc @mfine


mookerji:
ugh

mookerji:
This would be so much easier to do in anything but Haskell.

generator/sbpg/targets/resources/SbpTemplate.hs
```diff
@@ -5,35 +5,50 @@
 -- Maintainer:  Mark Fine <dev@swiftnav.com>
 -- Stability:   experimental
 -- Portability: portable
+--
+--
+
+{-# LANGUAGE OverloadedStrings #-}
```
mfine: what about `OverloadedStrings` in the cabal library section's `default-extensions`?

PR #262, opened 2015-09-17
Lens

mfine:
What this library needs is more lens.

/cc @mookerji

### DIFF
--- a/generator/sbpg/targets/resources/SbpTemplate.hs
+++ b/generator/sbpg/targets/resources/SbpTemplate.hs
@@ -134,7 +134,9 @@ instance FromJSON SBPMsg where
 
 merge :: Value -> Value -> Value
 merge (Object one) (Object two) = Object (one <> two)
-merge _ _ = Null
+merge _ (Object two) = Object two
+merge (Object one) _ = Object one
+merge _ v = v
 
 instance ToJSON SBPMsg where
 ((*- for m in msgs *))
--- a/generator/sbpg/targets/resources/SbpTemplate.hs
+++ b/generator/sbpg/targets/resources/SbpTemplate.hs
@@ -11,6 +11,7 @@
 module SwiftNav.SBP where
 
 import BasicPrelude hiding (lookup)
+import Control.Lens hiding ((.=))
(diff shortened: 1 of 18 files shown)


mookerji:
Oh man, another rev of the library, another >1k code review.

generator/sbpg/targets/resources/SbpTemplate.hs
```diff
@@ -121,25 +122,23 @@ instance Binary SBPMsg where
...
+        | msgType == (((m | to_global))) = SBP(((m))) <$> parseJSON obj <*> parseJSON obj
+((*- endfor *))
+        | otherwise = SBPMsgUnknown <$> parseJSON obj
+  parseJSON _ = mzero
 
 merge :: Value -> Value -> Value
-merge (Object one) (Object two) = Object $ one <> two
-merge _ _ = undefined
+merge (Object one) (Object two) = Object (one <> two)
+merge _ _ = Null
```
mookerji: Hmmmm, the problem here is that if we ever have something that isn't an `Object` map, it'll fail silently: https://hackage.haskell.org/package/aeson-0.9.0.1/docs/Data-Aeson.html#g:9 . I don't have strong opinions either way.

mookerji:
lgtm

generator/sbpg/targets/resources/SbpTemplate.hs
mfine: Hmm, yeah - was wondering about that.... maybe something like this???

```haskell
merge :: Value -> Value -> Value
merge (Object one) (Object two) = Object (one <> two)
merge obj@(Object _) _ = obj
merge _ obj@(Object _) = obj
merge obj Null = obj
merge _ obj = obj
```

Definitely think we should yank the `undefined` either way.

mookerji: Yes and yes.

mookerji: Ideally, this would be an Either, but I'm happy not complicating the generated dispatch code.

PR #263, opened 2015-09-17
Start to improve the module imports/exports

mfine:
We should start to improve the imports situation so that the library is more pleasant to use. Here's one stab that re-exports all the module plus `Msg` and `SBPMsg` - just import `SwiftNav.SBP` and go.

/cc @mookerji


mookerji:
Good idea.

PR #271, opened 2015-10-09
Heading message

benjamin0:
@fnoble please review, thanks!


fnoble:
I'd suggest adding `flags` and `n_sats` to the heading message (duplicate of `BASELINE_NED`)

mookerji:
This looks good. 

I recommend filling in your Github handle for future PRs you submit, just to keep track of work.

mfine:
This is great - when it comes in, we should rev the minor version of the library (`v0.51.0`), so that it'll serve as a conditional version point for consumers.

PR #276, opened 2015-10-21
Exception and session handling.

mookerji:
Adds configurable session handling (exponential backoff, retries,
etc.) and less cryptic exceptions.

/cc @mfine @fnoble @denniszollo

### DIFF
--- a/python/sbp/client/drivers/network_drivers.py
+++ b/python/sbp/client/drivers/network_drivers.py
@@ -8,15 +8,16 @@
 # EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
+"""TCP and HTTP networking client components.
+
+"""
+
 from .base_driver import BaseDriver
+from requests.adapters import DEFAULT_POOLBLOCK, DEFAULT_POOLSIZE, HTTPAdapter
+from requests.packages.urllib3.util import Retry
 import requests
 import socket
-import sys
 import threading
-import warnings
-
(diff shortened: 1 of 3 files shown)


python/sbp/client/drivers/network_drivers.py
```diff
@@ -97,11 +111,41 @@ class HTTPDriver(BaseDriver):
...
+               url="http://broker.testing.skylark.swiftnav.com",
+               timeout=DEFAULT_TIMEOUT):
+    self.session = requests.Session()
+    retry = Retry(connect=MAX_CONNECT_RETRIES,
+                  read=MAX_READ_RETRIES,
+                  redirect=MAX_REDIRECTS,
+                  status_forcelist=[500],
+                  backoff_factor=DEFAULT_BACKOFF_FACTOR)
+    self.session.mount("http://",
+                       HTTPAdapter(pool_connections=DEFAULT_POOLSIZE,
```
mfine: oh nice... I kind of question needing a conn pool, but why not

mfine: oh is that to get the retries?

mfine:
Looks good. Will this retry on those exceptions?

python/sbp/client/drivers/network_drivers.py
mookerji: Yep

mookerji:
I think what's going to happen here is that it'll retry as configured and then fall through to one of those exceptions. There's a possibility it'll throw a RetryException actually, so I'll add a test covering that case?

python/sbp/client/drivers/network_drivers.py
mookerji: I will note, the HTTPAdapter that configures retries actually already has a connection pool. My motivation here was to make that explicit in case we wanted to change it.

PR #278, opened 2015-10-27
Java client updates and HTTP client driver.

gsmcmullin:
Updates to the Java client to implement iterator interfaces similar to what we have in Python.  This also adds `SBPDriverHTTP` and `SBPForwarder` classes for connection of Piksi to Skylark.

/cc @mookerji

### DIFF
--- a/java/src/com/swiftnav/sbp/client/SBPForwarder.java
+++ b/java/src/com/swiftnav/sbp/client/SBPForwarder.java
@@ -1,0 +1,34 @@
+package com.swiftnav.sbp.client;
+
+import com.swiftnav.sbp.SBPMessage;
+import com.swiftnav.sbp.observation.MsgEphemeris;
+
+import java.io.IOException;
+
+public class SBPForwarder extends Thread {
+    private Iterable<SBPMessage> source;
+    private SBPSender sink;
+
+    public SBPForwarder(Iterable<SBPMessage> source_, SBPSender sink_) {
+        source = source_;
+        sink = sink_;
+    }
+
+    @Override
+    public void run() {
+        for (SBPMessage msg : source) {
+            System.out.println(msg.toString());
+            try {
+                if (msg.type == MsgEphemeris.TYPE) {
+                    System.out.println(msg.toJSON());
+                    System.out.println(msg.getPayload().toString());
+                }
+                sink.sendMessage(msg);
+            } catch (IOException e) {
+                e.printStackTrace();
+                break;
+            }
+        }
+        System.out.println("Forwarder thread exiting");
+    }
+}
--- a/java/src/com/swiftnav/sbp/client/SBPForwarder.java
+++ b/java/src/com/swiftnav/sbp/client/SBPForwarder.java
@@ -19,10 +19,6 @@ public class SBPForwarder extends Thread {
         for (SBPMessage msg : source) {
             System.out.println(msg.toString());
             try {
-                if (msg.type == MsgEphemeris.TYPE) {
-                    System.out.println(msg.toJSON());
-                    System.out.println(msg.getPayload().toString());
-                }
                 sink.sendMessage(msg);
             } catch (IOException e) {
                 e.printStackTrace();
--- a/java/src/com/swiftnav/sbp/client/SBPFramer.java
+++ b/java/src/com/swiftnav/sbp/client/SBPFramer.java
@@ -11,7 +11,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 
-public class SBPFramer extends SBPIterable {
+public class SBPFramer extends SBPIterable implements SBPSender {
     private static final byte PREAMBLE = 0x55;
 
     private static final int PREAMBLE_SIZE = 1;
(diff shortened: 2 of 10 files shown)


java/src/com/swiftnav/sbp/drivers/SBPDriverHTTP.java
```diff
@@ -0,0 +1,44 @@
...
+        url = new URL(url_);
+        uid = uid_;
+    }
+
+    @Override
+    protected void openStream() throws IOException {
+
+        System.out.println("Connecting to " + url.toString());
+
+        HttpURLConnection http = (HttpURLConnection)url.openConnection();
```
mookerji: You may want to set up some connection and read timeouts here as well.

http://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html#setReadTimeout(int)

What happens if the connection fails in the app? Can you reinitiate the connection manually? I was looking to see if the JDK had a way of doing retries, but apparently not.

java/src/com/swiftnav/sbp/client/SBPForwarder.java
```diff
@@ -0,0 +1,34 @@
...
+        source = source_;
+        sink = sink_;
+    }
+
+    @Override
+    public void run() {
+        for (SBPMessage msg : source) {
+            System.out.println(msg.toString());
+            try {
+                if (msg.type == MsgEphemeris.TYPE) {
```
mookerji: This is only forwarding the ephemeris and not the observations?

mookerji: Also, where are these messages being forwarded to? I seem to remember that the Python client forwarded all messages to the logger and that was it.

java/src/com/swiftnav/sbp/drivers/SBPDriverHTTP.java
```diff
@@ -0,0 +1,44 @@
...
+        //http.setRequestProperty("Content-Type", "application/sbp");
+
+        http.connect();
+        System.out.println("" + http.getResponseCode() + " " + http.getResponseMessage());
+
+        if (http.getDoOutput()) {
+            out = new BufferedOutputStream(http.getOutputStream());
+            System.out.println(out.toString());
+        }
+        if (http.getDoInput()) {
```
mookerji: I might be missing something, but where is `SBPDriverHTTP` actually being invoked?

java/src/com/swiftnav/sbp/client/SBPForwarder.java
gsmcmullin: It's forwarding everything the source produces.  This junk was to debug a crash when handling the Ephemeris message.  I'll remove this.

java/src/com/swiftnav/sbp/drivers/SBPDriverHTTP.java
gsmcmullin: It isn't.  It's for use in the Android app.  I hacked it into `SerialLink` for testing, but didn't commit that.  I'll add an updated `SerialLink` that instantiates `SBPDriverHTTP` and `SBPForwarder`.

java/src/com/swiftnav/sbp/client/SBPFramer.java
```diff
@@ -0,0 +1,89 @@
...
+        int sender = bb.getShort() & 0xffff;
+        int len = bb.get() & 0xff;
+
+        byte[] payload = driver.read(len);
+        calccrc = CRC16.crc16(payload, calccrc);
+
+        int crc = ByteBuffer.wrap(driver.read(CRC_SIZE))
+                .order(ByteOrder.LITTLE_ENDIAN).getShort() & 0xffff;
+        if (crc != calccrc) {
+            System.err.println("CRC error in received SBPMessage");
```
JoshuaGross: Should we be passing these errors along instead of swallowing them here?

java/src/com/swiftnav/sbp/drivers/SBPDriverHTTP.java
gsmcmullin: As it is now, if the connection breaks, the iterator will stop iteration and the forwarder thread will exit.  I guess we should probably handle timeouts and retries here, and require the application to interrupt the forwarder thread if it's no longer needed.

java/src/com/swiftnav/sbp/client/SBPFramer.java
gsmcmullin: I don't know of a convenient way to do that.  They are not passed up in the C or Python clients.

java/src/com/swiftnav/sbp/client/SBPForwarder.java
gsmcmullin: The messages are forwarded to an implementation of the `SBPSender` interface, as passed to the constructor.  `SBPFramer` implements this interface.  I'll update `JSONLogger` to do the same.

mookerji:
This is great. Merge it whenever.

gsmcmullin:
Merging as is.  Timeout/reconnect logic in the HTTP client can come in as another PR.

PR #281, opened 2015-11-02
[superseded] Added signal_t struct and modified messages to use it.

gsmcmullin:
Rebase of #235


jacobmcnamee:
Looks fine, but we should add test coverage for the modified messages.

jacobmcnamee:
If the updated message is intended to be binary compatible, then u16 prn needs to occupy the LSBs of the sbp_signal_t struct.

It's also a bit misleading to name the field PRN when we are actually storing SID = PRN-1.

jacobmcnamee:
Rebased / updated in #304

PR #283, opened 2015-11-04
Initial Javascript bindings work

JoshuaGross:
Javascript bindings including unit tests, streaming interface (the `dispatch` method). I can provide some demo tomorrow and get the tests running on Travis.

### DIFF
--- a/generator/sbpg/targets/javascript.py
+++ b/generator/sbpg/targets/javascript.py
@@ -1,0 +1,182 @@
+#!/usr/bin/env python
+# Copyright (C) 2015 Swift Navigation Inc.
+# Contact: Joshua Gross <josh@swiftnav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+"""Generator for JavaScript target.
+
+This module consumes the YAML spec and generates some message classes
+files.
+
+"""
+
+from sbpg.targets.templating import JENV
+import copy
+import textwrap
+
+from construct import *
+
+TEMPLATE_NAME = "sbp_js.js.j2"
+
+CONSTRUCT_CODE = {
+  'u8': 'uint8',
+  'u16': 'uint16',
+  'u32': 'uint32',
+  'u64': 'uint64',
+  's8': 'int8',
+  's16': 'int16',
+  's32': 'int32',
+  's64': 'int64',
+  'float': 'floatle',
+  'double': 'doublele',
+}
+
+WRITE_BUFFER_CODE = {
+  'u8': 'writeUInt8',
+  'u16': 'writeUInt16LE',
+  'u32': 'writeUInt32LE',
+  'u64': 'writeUIntLE', # only supports up to 48 bits of accuracy
+  's8': 'writeInt8',
+  's16': 'writeInt16LE',
+  's32': 'writeInt32LE',
+  's64': 'writeIntLE', # only supports up to 48 bits of accuracy
+  'float': 'writeFloatLE',
+  'double': 'writeDoubleLE',
+}
+
+BUFFER_SIZE_MAP = {
+  'u8': 1,
+  'u16': 2,
+  'u32': 4,
+  'u64': 8,
+  's8': 1,
+  's16': 2,
+  's32': 4,
+  's64': 8,
+  'float': 4,
+  'double': 8
+}
+
+# lol
+JSDOC_CODE = {
+  'u8': 'number (unsigned 8-bit int, 1 byte)',
+  'u16': 'number (unsigned 16-bit int, 2 bytes)',
+  'u32': 'number (unsigned 32-bit int, 4 bytes)',
+  'u64': 'number (unsigned 64-bit int, 8 bytes)',
+  's8': 'number (signed 8-bit int, 1 byte)',
+  's16': 'number (signed 16-bit int, 2 bytes)',
+  's32': 'number (signed 32-bit int, 4 bytes)',
+  's64': 'number (signed 64-bit int, 4 bytes)',
+  'float': 'number (float, 4 bytes)',
+  'double': 'number (float, 8 bytes)',
+}
+
+def is_array():
+  return False
+
+
+def builtin_type(type_id, type_map=WRITE_BUFFER_CODE):
+  return type_map.get(type_id, None) is not None
+
+
+def write_buffer(type_id, type_map=WRITE_BUFFER_CODE):
+  return type_map.get(type_id, None)
+
+
+def buffer_size(type_id, type_map=BUFFER_SIZE_MAP):
+  if not type_map.get(type_id, None) is None:
+    return type_map.get(type_id, None)
+  elif type(type_id) == str:
+    return "this.fields.%s.length" % type_id
+  else:
+    return "this.fields.%s.length" % type_id.identifier
+
+def jsdoc_format(type_id, jsdoc=JSDOC_CODE):
+  """
+  Formats type for jsdoc.
+  """
+  return jsdoc.get(type_id, type_id)
+
+
+def construct_format(f, type_map=CONSTRUCT_CODE):
+  """
+  Formats for binary-parser library.
+  """
+  formatted = ""
+  if type_map.get(f.type_id, None):
+    return "%s('%s')" % (type_map.get(f.type_id), f.identifier)
+  elif f.type_id == 'string' and f.options.get('size', None):
+    return "string('%s', { length: %d })" % (f.identifier, f.options['size'].value)
+  elif f.type_id == 'string':
+    return "string('%s', { greedy: true })" % (f.identifier)
+  elif f.type_id == 'array':
+    fill = f.options['fill'].value
+    f_ = copy.copy(f)
+    f_.type_id = fill
+    size = f.options.get('size', None)
+    size_fn = f.options.get('size_fn', None)
+    field_type = type_map.get(f_.type_id, None)
+    if field_type is None:
+      field_type = "%s.prototype.parser" % f_.type_id
+    else:
+      field_type = "'%s'" % field_type
+    if size is not None:
+      return "array('%s', { length: %d, type: %s })" % (f.identifier, size.value, field_type)
+    elif f.options.get('size_fn') is not None:
+      return "array('%s', { type: %s, length: '%s' })" % (f_.identifier, field_type, size_fn.value)
+    else:
+      return "array('%s', { type: %s, readUntil: 'eof' })" % (f_.identifier, field_type)
+  else:
+    return "nest('%s', { type: %s.prototype.parser })" % (f.identifier, f.type_id)
+  return formatted
+
+
+def js_classnameify(s):
+  """
+  Makes a classname.
+  """
+  if not '_' in s:
+    return s
+  return ''.join(w[0].upper() + w[1:].lower() for w in s.split('_'))
+
+def star_wordwrap(s):
+  return "\n * ".join(textwrap.wrap(' '.join(s.split('\n')), 80))
+
+def star_wordwrap_indent(s):
+  return "\n *   ".join(textwrap.wrap(' '.join(s.split('\n')), 80))
+
+JENV.tests['builtinType'] = builtin_type
+
+JENV.filters['js_classnameify'] = js_classnameify
+JENV.filters['writeBuffer'] = write_buffer
+JENV.filters['bufferSize'] = buffer_size
+JENV.filters['construct_js'] = construct_format
+JENV.filters['jsdoc'] = jsdoc_format
+JENV.filters['starWordWrap'] = star_wordwrap
+JENV.filters['starWordWrapIndent'] = star_wordwrap_indent
+
+
+def render_source(output_dir, package_spec, jenv=JENV):
+  """
+  Render and output
+  """
+  path, name = package_spec.filepath
+  directory = output_dir
+  destination_filename = "%s/%s.js" % (directory, name)
+  py_template = jenv.get_template(TEMPLATE_NAME)
+  module_path = ".".join(package_spec.identifier.split(".")[1:-1])
+  includes = [".".join(i.split(".")[:-1]) for i in package_spec.includes]
+  includes = [i for i in includes if i != "types"]
+  with open(destination_filename, 'w') as f:
+    f.write(py_template.render(msgs=package_spec.definitions,
+                               filepath="/".join(package_spec.filepath) + ".yaml",
+                               module_path=module_path,
+                               include=includes,
+                               timestamp=package_spec.creation_timestamp,
+                               description=package_spec.description))
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -2,6 +2,8 @@
 
 From the parent directory, run `make javascript` to regenerate bindings.
 
+This library supports and is tested with NodeJS v0.10.x, v0.12.x, v4.x, and v5.x.
+
 # Using the JavaScript libsbp bindings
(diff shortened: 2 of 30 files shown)


mookerji:
!!!

gsmcmullin:
Can we please break the anti-pattern of keeping the generated files in version control?

JoshuaGross:
@gsmcmullin Why don't you like checking in generated code? In my experience it makes debugging (especially things like `bisect`) significantly easier as you don't need to consider if someone had a build environment set up differently or go through any extra steps besides checking out the code. Are you concerned about history being "messy"?

mfine:
Checking in the code is ok. This repo serves both the purpose of generating code and hosting code artifacts (with some non generated artifacts as well). It's a common pattern in generated libraries. See:

https://github.com/brendanhay/amazonka
https://github.com/brendanhay/gogol
https://github.com/aws/aws-sdk-go/tree/master/service
https://github.com/boto/boto/tree/develop/boto

gsmcmullin:
It bothers me because I can't review a diff that's 90% generated crap.  Because others do it doesn't make it a good idea.

JoshuaGross:
If that's the only concern we can easily split the generated-code-only changes into their own commits so that they can be isolated. That said, that code needs to be reviewed anyway.

javascript/sbp/acquisition.js
```diff
@@ -0,0 +1,177 @@
...
+
+  
+  This message describes the results from an attempted GPS signal
+acquisition search for a satellite PRN over a code phase/carrier
+frequency range. It contains the parameters of the point in the
+acquisition search space with the best signal-to-noise (SNR)
+ratio.
+
+
+  Parameters
```
mookerji: The comment format here is for numpy's Doxygen comments. If there's some kind of autogenerated documentation for javascript using embedded comment annotations, I recommend using that.

jacobmcnamee:
Can we move the generated files to a separate repo? Having everything in one place makes code review tedious, and it's very easy for generated files to get out of sync.

I don't see the use case for needing to download both the generated files and the build system, let alone in the same repo.

mfine:
From a customer getting tripped up on missing generated Java files...

```Hello! 

I'm trying to send Piksi data to an Android device via Bluetooth. I'm using your libsbp library to decode the messages. 

Is it possible that there are some files missing in you github repository? The example uses the class MsgTrackingState but I can't find this or any other message classes in the java folder. However, this classes exist in the python folder. 

I tried to implement the class MsgPosLLH myself, but I can't get it to work properly. 

Kind regards, 
David```

**Comment by mookerji** [comment]
@gsmcmullin @jacobmcnamee, so in every project of this kind I've seen, it's been typical to keep generated code and the compiler under version control in the same repository. I was going to list examples but it seems @mfine has already done that. I disagree pretty strongly about splitting this repository up or keeping the source code out of version control. If this is an antipattern, I'd prefer to see some more substantive arguments describing why its necessary rather than the statement that others are Wrong.

Some points:
* Having multiple repositories or separate release artifacts introduces more management overhead to keep track off than we currently have. If you think it's tedious now to code review or manage multi-language clients across directories, just wait until they're split into multiple repositories. If we generated the source code as release artifacts, most users and developers would likely download and review those separately. At this point, you'd go through a more roundabout approach to performing what we currently do. The easy thing with the current repository structure is a very flat information hierarchy: you can read through source, review it, and deploy it all from a single place. It's tedious, yes, but typically for large commits and less arguably tedious than the alternatives.
* The primary artifact here for all consumers is the generated source code. How that source code is generated is different for each client, but ultimately it's intended to be human readable and reviewed. You cannot review the generator of the code and then reason, "Yes! This AST will generate the source I want.". It is the generated source code that is meant to be reviewed and consumed by our clients, which is why it's checked into version control.
* If the generated files are out-of-sync, then we can put a procedure (git commit hook, etc.) into place that fixes that, because it definitely sounds like a problem. I totally agree.
* The time spent with repository gymnastics and yakshaving is time best spent elsewhere.

File: generator/sbpg/targets/javascript.py [hunk 2]
```diff
@@ -0,0 +1,155 @@
...
+  'u64': 8,
+  's8': 1,
+  's16': 2,
+  's32': 4,
+  's64': 8,
+  'float': 4,
+  'double': 8
+}
+
+PYDOC_CODE = {
```
mookerji: not pydoc. Also does js have different numeric types?

javascript/README.md
```diff
@@ -0,0 +1,13 @@
+# Building the JavaScript libsbp bindings
+
+From the parent directory, run `make javascript` to regenerate bindings.
+
+# Using the JavaScript libsbp bindings
```
mookerji: Not really related to this readme, but can we package the javascript client as well? npm or something?

javascript/sbp/acquisition.js
mookerji: Like I think the comment would go above the function/object closure like maybe: 

```
/**
 * SBP ...
 *
 * @param {Object} options
 * @access public
 */
```

mookerji: Also, fixing some of the spacing and alignment will definitely make the source a bit easier to read.

javascript/sbp/acquisition.js
```diff
@@ -0,0 +1,148 @@
...
+
+  this.sbp = sbp;
+  this.fields = this.parser.parse(msg);
+
+  return this;
+};
+MSG_ACQ_RESULT_DEP_A.prototype = Object.create(SBP.prototype);
+MSG_ACQ_RESULT_DEP_A.prototype.constructor = MSG_ACQ_RESULT_DEP_A;
+MSG_ACQ_RESULT_DEP_A.prototype.toBuffer = function () {
+  var buffers = [];
```
mookerji: Hmmm, so this inlining of allocating all these buffers and concatenating them seems a bit tedious. Also, does `binary-parser` not have a method for binary serializing?

javascript/sbp/observation.js
```diff
@@ -0,0 +1,1016 @@
...
+  buffers = buffers.concat( ObservationHeader.prototype.toBuffer.apply({ fields: this.fields.header }));
+  for (var i in this.fields.obs) {
+    buffers = buffers.concat( PackedObsContent.prototype.toBuffer.apply({ fields: this.fields.obs[i] }));
+  }
+  return Buffer.concat(buffers);
+};
+MSG_OBS.prototype.parser = new Parser()
+  .endianess('little')
+  .nest('header', { type: ObservationHeader.prototype.parser })
+  .array('obs', { type: PackedObsContent.prototype.parser, readUntil: 'eof' });
```
mookerji: eof?

The observation messages can definitely be super-complicated to get right, from the declarative parsing point of view. Did this work? The tracking message is similar.

python/sbp/settings.py
```diff
@@ -300,7 +300,7 @@ def to_json_dict(self):
 class MsgSettingsReadByIndexReq(SBP):
   """SBP class for message MSG_SETTINGS_READ_BY_INDEX_REQ (0x00A2).
 
-  You can have MSG_SETTINGS_READ_BY_INDEX_REQ inherent its fields directly
+  You can have MSG_SETTINGS_READ_BY_INDEX_REQ inherit its fields directly
```
mookerji: Eh, yeah, it probably would have made more sense to keep the generated python stuff with this comment spelling correction into a separate commit.

javascript/sbp/acquisition.js
mookerji: nit: Ok, the alignment comment was bullshit. Crunching the spacing down in the comments would be cool.

javascript/README.md
JoshuaGross: yeah, I intend to do that - hopefully as part of the build process - once this PR has been approved

generator/sbpg/targets/javascript.py
JoshuaGross: Unfortunately not, everything is `number`. Numbers can have decimal points, or not.

javascript/sbp/acquisition.js
JoshuaGross: It appears that they don't. I'm happy to make this `toBuffer` routine less tedious-looking. I already moved some of the stuff to `sbp.js` (it used to all be inlined).

javascript/sbp/observation.js
JoshuaGross: It's passing the tests so I guess so? `eof` is end-of-file, is that not what's expected? Basically the payload is parsed separately (as in Haskell and Python implementations AFAIK) so we can read to the end of the payload which reads until (in effect) immediately before the crc.

python/sbp/settings.py
JoshuaGross: word. I can pull this out

generator/sbpg/targets/javascript.py
mookerji: Yeah, you're probably going to want to change that in generated comments.

javascript/README.md
mookerji: Lovely

javascript/sbp/acquisition.js
mookerji: Wow, that's really fucking weird.

Uh, yeah a utility function or something might be DRY here.  (or maybe submitting a PR to `binary-parser` for that method). For what it's worth though, the use case of actually materializing a buffer from a JSON object here is likely to be rare (in the short term)? Typically, that's something you'd want to do if you were writing commands over to the Piksi (like from the console), so perhaps it's not actually worth worrying about too much.

javascript/sbp/observation.js
mookerji: I was confused for a minute. If `eof` == `end of buffer`, that's fine.

javascript/sbp/observation.js
```diff
@@ -0,0 +1,1016 @@
...
+};
+PackedObsContentDepA.prototype.parser = new Parser()
+  .endianess('little')
+  .uint32('P')
+  .nest('L', { type: CarrierPhase.prototype.parser })
+  .uint8('cn0')
+  .uint16('lock')
+  .uint8('prn');
+
+var MSG_OBS_DEP_A = function (sbp, msg) {
```
mookerji: These object names should probably be cased (vs capped). All caps has typically been used for constants.

JoshuaGross: will do

javascript/sbp/observation.js
JoshuaGross: yeah, exactly

javascript/sbp/observation.js
```diff
@@ -0,0 +1,1016 @@
...
+};
+PackedObsContent.prototype.parser = new Parser()
+  .endianess('little')
+  .uint32('P')
+  .nest('L', { type: CarrierPhase.prototype.parser })
+  .uint8('cn0')
+  .uint16('lock')
+  .uint32('sid');
+
+var MSG_OBS = function (sbp, msg) {
```
mookerji: Hrmmmm, what's both `sbp` and `msg` doing here? Is `sbp` a mixin?

You probably want to keep the original calling contract (using an existing sbp object or instantiating an object from a dict of fields):

```python
def __init__(self, sbp=None, **kwargs):
    if sbp:
      super( MsgTrackingIq,
             self).__init__(sbp.msg_type, sbp.sender, sbp.length,
                            sbp.payload, sbp.crc)
      self.from_binary(sbp.payload)
```

JoshuaGross: `msg` should be renamed to `payload` or I can just use `sbp.payload`

javascript/sbp/observation.js
JoshuaGross: done

generator/sbpg/targets/javascript.py
JoshuaGross: Done. I changed to the actual JavaScript type but with more detail, which I think folks will find useful. Example: `u16` now maps to `number (unsigned 16-bit int, 2 bytes)`

JoshuaGross:
@mookerji I replaced the inlined `toBuffer` code generation with `fieldSpecs`.

javascript/sbp/piksi.js
```diff
@@ -0,0 +1,366 @@
...
+
+**/
+
+var SBP = require('./sbp');
+var Parser = require('binary-parser').Parser;
+
+/**
+ * SBP class for message MSG_ALMANAC (0x0069).
+ *
+ * This is a legacy message for sending and loading a satellitealamanac onto the Piksi's flash memory from the host.
```
mookerji: word wrap this description and the field comments as well?

javascript/sbp/acquisition.js
```diff
@@ -0,0 +1,93 @@
...
+var SBP = require('./sbp');
+var Parser = require('binary-parser').Parser;
+
+/**
+ * SBP class for message MSG_ACQ_RESULT (0x0014).
+ *
+ * This message describes the results from an attempted GPS signalacquisition search for a satellite PRN over a code phase/carrierfrequency range. It contains the parameters of the point in theacquisition search space with the best signal-to-noise (SNR)ratio.
+ *
+ * Fields in the SBP payload (`sbp.payload`):
+ * @field snr number (float, 4 bytes) SNR of best point. Currently dimensonless, but will haveunits of dB Hz in the revision of this message.
```
mookerji: nit: word wrap on comment and spacing (`haveunits`).

javascript/sbp/msg.js
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  return crc;
+}
+
+module.exports = {
+  decode: function decode (msg) {
+    var sbp = parser.parse(msg);
+
+    return new sbpTable[sbp['msg_type']](sbp);
```
mookerji: Hey, so what happens here if the dispatch fails because the `msg_type` is missing? I'm assuming you get some kind of dictionary-related exception/error. In the past, we've caught these, logged a warning, and passed the raw SBP message through.

https://github.com/swift-nav/libsbp/blob/master/python/sbp/table.py#L68

javascript/sbp/file_io.js
```diff
@@ -0,0 +1,240 @@
...
+
+**/
+
+var SBP = require('./sbp');
+var Parser = require('binary-parser').Parser;
+
+/**
+ * SBP class for message MSG_FILEIO_READ_REQ (0x00A8).
+ *
+ * The file read message reads a certain length (up to 255 bytes)from a given offset into a file, and returns the data in aMSG_FILEIO_READ_RESP message where the message length fieldindicates how many bytes were succesfully read.The sequencenumber in the request will be returned in the response.If the message is invalid, a followup MSG_PRINT message willprint "Invalid fileio read message". A device will only respondto this message when it is received from sender ID 0x42.
```
mookerji: nit: More spacing errors `aMSG_FILEIO_...`

javascript/sbp/msg.js
```diff
@@ -0,0 +1,210 @@
...
+    var getFramedMessage = function () {
+      var headerBuf, payloadBuf;
+      var preamble, msgType, sender, length, crc;
+      var payloadCrc;
+
+      // Find preamble byte
+      var preamblePos;
+      for (preamblePos = 0; preamblePos < streamBuffer.length; preamblePos++) {
+        preamble = streamBuffer.readUInt8(preamblePos);
+        if (preamble == 0x55) {
```
mookerji: nit: magic number

javascript/sbp/msg.js
JoshuaGross: Great, good call. Fixing this and adding a test in the next commit.

javascript/sbp/msg.js
JoshuaGross: resolved

javascript/sbp/file_io.js
JoshuaGross: resolved - I fixed all the spacing errors (AFAICT). I improved the word wrapping significantly.

JoshuaGross:
@mookerji @mfine 

1. Rebased into two commits, one for generation code/scaffolding, and one for the generated code.
2. Removed noisy Python changes that were in previous commits.
3. I cleaned up the generated comments significantly - they're actually readable now.
4. Ponies.

mfine:
This is awesome. The 2 commit split works great! Thanks!

JoshuaGross:
Fixed to work with NodeJS 0.10.x, 0.12.x, v4.x and v5.x (this is why the Travis build was failing).

mookerji:
Ship it!

JoshuaGross:
:dancer:

PR #286, opened 2015-11-05
Tooling to convert JSON to binary, and binary to JSON

mfine:
Couple of simple tools to read, write from, to stdin, stdout. Converts JSON to binary, and binary to JSON.

/cc @mookerji


mfine:
Going to bring this in.

PR #291, opened 2015-11-06
Update instructions to include details on publishing NPM package.

JoshuaGross:


JoshuaGross:
@mfine  @mookerji

Makefile
```diff
@@ -82,6 +82,7 @@ javascript:
                        -r $(SBP_MAJOR_VERSION).$(SBP_MINOR_VERSION) \
 		       --javascript;\
 	cd $(SWIFTNAV_ROOT);
+	- npm version "v$(SBP_MAJOR_VERSION).$(SBP_MINOR_VERSION).$(SBP_PATCH_VERSION)" --no-git-tag-version
```
mfine: what does the leading dash do?

mfine:
Looks good :+1:

Makefile
JoshuaGross: whoa, eagle eyes! It ignores any errors from that line. In this case, `npm version` will throw an error if the version doesn't change, which we don't really care about.

mfine: cool!

PR #297, opened 2015-11-09
Per-message test YAML

jacobmcnamee:
Generate a test YAML file for each message separately (instead of grouping by package). 
Store files in subdirs named by package, e.g. acquisition/test_MsgAcqResult.yaml. 
Add test data for many current message types.

Separating by message type makes it much easier to add test cases for new message types without modifying existing tests.


JoshuaGross:
Looks great and helped smoke out some bugs in the JavaScript client, so :+1:

PR #300, opened 2015-11-09
Fix JavaScript corrupt preamble tests

JoshuaGross:
Some tests will cause a corrupt 'length' value that is much longer than one message, which will cause the test to stall and wait for more data unless the stream is large. In real-life cases, the constant stream of bytes from a device will quickly exceed the corrupt length. To account for this, we repeat the message several times in the 'corrupt preamble' tests.

@jacobmcnamee this will fix the tests in your PR

### DIFF
--- a/javascript/tests/test_dispatch_decoder.js
+++ b/javascript/tests/test_dispatch_decoder.js
@@ -37,16 +37,60 @@ describe('test packages based on YAML descriptors, through the dispatcher', func
             });
           });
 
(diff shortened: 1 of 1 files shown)


jacobmcnamee:
I'm still getting errors when I test with #297 

16) should parse binary sbp and payload with leading extra preamble
17) should parse binary sbp and payload with leading extra preamble (2)

JoshuaGross:
Moving this to #301

PR #306, opened 2015-11-16
fix directory path in build instructions in c/README.md

PaulBouchier:
The instructions currently say to make the build subdirectory in the libsbp root directory. This is incorrect - it should be made in the c subdirectory.

### DIFF
--- a/c/README.md
+++ b/c/README.md
@@ -18,11 +18,10 @@ sudo apt-get install build-essential pkg-config cmake
 On other systems, you can obtain CMake from your operating system
 package manager or from http://www.cmake.org/.
 
-Once you have the dependencies installed,
+Once you have the dependencies installed, from this (the libsbp/c directory)
 create a build directory where the library will be built:
 
 ```shell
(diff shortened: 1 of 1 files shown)


c/README.md
```diff
@@ -18,10 +18,11 @@ sudo apt-get install build-essential pkg-config cmake
 On other systems, you can obtain CMake from your operating system
 package manager or from http://www.cmake.org/.
 
-Once you have the dependencies installed, From the libsbp root
-directory, create a build directory where the library will be built:
+Once you have the dependencies installed,
```
mookerji: Hey @PaulBouchier: unfortunately these instructions are invalid if you're reading the README in the c directory. What about changing the existing text to `From this (the libsbp c) directory` or something.

PaulBouchier:
Good point. I fixed it as you suggested. Thanks. Hopefully good to be pulled now.

mookerji:
@PaulBouchier : marvelous! Thanks for helping out.

PR #307, opened 2015-11-16
Automating testing of C libraries.

JoshuaGross:
Generate C tests. There are a bunch of tests that are failing on decoded field values which makes me think there may actually be bugs in the library, or I'm misunderstanding these results. I'd appreciate some feedback either way.

@mookerji @jacobmcnamee @mfine @gsmcmullin


JoshuaGross:
Note that the commits are split into the code-generation bits, and the generated code.

PR #308, opened 2015-11-17
Add a write interface to the HTTP file-like handle.

mookerji:
Add a write interface and refresh to the HTTP file-like handle.
Refactors the HTTP handle to work more like a session client that can
create a write/read stream on-demand. Main object keeps a threadpool
executor for PUTs so that streaming network connections don't hang up
the calling thread.

/cc @mfine @JoshuaGross

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -6,6 +6,7 @@ flake8==2.3.0
 httpretty
 requests>=2.8.1
 futures==2.2.0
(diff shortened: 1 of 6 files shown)


JoshuaGross:
fix build plz?

mookerji:
Fixed! 

Adding tests and support for the pragma change.

PR #310, opened 2015-11-18
User data message

fnoble:
Adding a message that can be used to package arbitrary user data.

Seems like there were some changes to the generation steps so not sure if I did everything correctly.

/cc @mfine @mookerji


mfine:
Maybe call it `MSG_USER_DATA` in case we come up with other messages to go in the User package? Otherwise, looks ok - we should get a test in for it, too, eventually.

mookerji:
Do we also want to indicate the maximum payload length?

fnoble:
@mfine @mookerji both good points, updated

mookerji:
Ack, and you're also going to want to add your new package to https://github.com/fnoble/libsbp/blob/user_data/python/sbp/table.py#L27 and the table below. Sorry for not pointing that out earlier.

JoshuaGross:
We're probably going to want to auto-generate all the table stuff eventually. No reason not too.

njoubert:
ME LIKEY, ME WANT!

PR #313, opened 2015-11-20
Fix Haskell builds (fix import codegen for `GnssSignal`)

JoshuaGross:
make sure that all necessary imports are actually imported (codegen fix)


JoshuaGross:
@mfine  @mookerji

mfine:
sweet!

PR #320, opened 2015-12-07
Move MSG_BASE_POS to MSG_BASE_POS_LLH, Add MSG_BASE_POS_ECEF

mfine:
Other observation protocols are using ECEF coordinate system to publish their base station locations. Internally, we're converting the LLH to ECEF anyway. Provide a means to bypass the conversion and consume and publish ECEF base position messages.


denniszollo:
When we make changes to message names but don't change anything about the message, perhaps we should consider leaving an alias in the library with the old name.  For example, in this PR, we would add not only the SBP_MSG_BASE_POS_LLH as an identifier (and it's corresponding class name), but also, since the content of the message did not change, we'd retain a pointer of the old SBP_MSG_BASE_POS (and the corresponding class name).  In this way tools would continue to function without needing updates themselves.  We'd only break downstream tools when the CONTENT as opposed to the name, changes.  Any thoughts?

JoshuaGross:
The goal should be to *never* break downstream tools by deprecating old messages and creating new ones, right? Anyway, I'm in favor of doing that here, too - deprecate the old message name and add the new ones.

mfine:
The name hasn't changed - it's still `0x0044` - the label for the name has changed. If tools upgrade the version they build against, they will have to change the label.

mookerji:
lgtm!

PR #321, opened 2015-12-08
SBPGnssSignal -> GnssSignal

mfine:
Remove the SBP prefix on the structure. We don't prepend `SBP` anywhere in the library - it's implied in the the current namespacing of `swiftnav.sbp`.

Any objections to removing the prefix? I'll fix up downstream libraries.

/cc @mookerji @fnoble @jacobmcnamee @gsmcmullin


JoshuaGross:
This seems like a lot of changes for relatively little payoff, especially fixing up downstream stuff, but seems reasonable otherwise

mfine:
Ahh, so there's conflict with SwiftNav's gnss_signal_t. Great :banana:. How about calling this MsgGnssSignal - that would line up well with the conversion routines (ephemeris to msg_ephemeris). I know it's not a message, but the prefix works - wouldn't be a bad prefix to put on all the subtypes, tbh.

jacobmcnamee:
Without the prefix there will be no way to differentiate in C. Current we do distinguish between the external SBP struct and an internal (non-prefixed) struct. I'd rather keep the prefix.

mfine:
Right, what do you think about using "Msg" as a prefix?

jacobmcnamee:
SBP is more accurate than Msg. You could modify the code generators to spit out prefixes for languages without packages / namespaces (like C) and omit them otherwise. Not sure it's worth the effort, though.

mfine:
SBP is already part of the namespace and the generators for other languages are already using it in their namespaces, so using SBP again is redundant (and bad looking in some languages). Msg does at least follow the pattern with other structs, e.g., it's a message relates struct. But if we want to leave it as is, that's fine too.

mfine:
Ooooohhh... could add a hard coded rule to the C generator for this structure and prepend it with SBP - without for everyone else. That should leave everybody happy.

mfine:
(we probably should have pretended everything in C with sbp_)

jacobmcnamee:
I agree - really we should have sbp_ in front of all the definitions for languages like C without namespaces / packages. Maybe for now we can remove all prefixes from the yaml (which I like) and hardcode the C generator to add sbp_ just to gnss_signal_t (as a stopgap until we hopefully add sbp_ to everything for C) like you suggested.

mfine:
Added a `COLLISIONS` array to the C generator so that we don't leak the C collision resolution back into the library names.

PR #323, opened 2015-12-17
hide 'fd is not an integer' error

scottswift:


mookerji:
@scottswift what's the context of this bugfix?

scottswift:
running serial_link on timon in the capture script leads to many file descriptor errors that don't seem to be fatal. this passes them through as IOErrors, but I didn't diagnose the underlying problem

JoshuaGross:
should we merge this or punt it?

mookerji:
My thought is punt? I want to understand what the actual problem is instead of paper over it?

PR #324, opened 2015-12-17
Change the document style to match new branding

mfine:
Style changes and logo changes to the SBP data sheet per the new brand guidelines.

/cc @fnoble @cbeighley @robhranac


cbeighley:
Looks great.

PR #329, opened 2016-01-16
Uart state

denniszollo:
Deprecate the UART state message and create a new one that adds in the "period of reception" info about observation messages from the base as a "latency" object.  

This gives us an idea of the link quality back to the base station.  The latency stuff only works for messages we received and decoded, it doesn't take into account any dropped packets.  This new obs_period will.  This will be useful in pixhawk debugging to help understand how often the observations make it to the rover.


JoshuaGross:
This is almost a month old. Can we close this or merge it? @mookerji @jacobmcnamee @gsmcmullin

denniszollo:
It would only make sense to merge this if we decide to merge the corresponding firmware and tools PRs.

PR #335, opened 2016-02-09
Readable makefile + readable Make errors + fix C codegen arbitrary renaming

JoshuaGross:
Setup instructions are really confusing and scattered everywhere. We should resolve that as a documentation problem, and as a `Makefile` problem. I've attempted to do both here by making requirements more explicit in some cases, and wherever possible failing early in the `Makefile` if some necessary dependency isn't installed. 

Motivated by @sashamitelman's trouble running `make` this afternoon.

@mookerji @mfine @jacobmcnamee @denniszollo


c/include/libsbp/acquisition.h
```diff
@@ -42,7 +42,7 @@ be in units of dB Hz in a later revision of this message.
  */
   float cp;     /**< Code phase of best point [chips] */
   float cf;     /**< Carrier frequency of best point [hz] */
-  sbp_gnss_signal_t sid;    /**< GNSS signal for which acquisition was attempted */
+  gnss_signal_t sid;    /**< GNSS signal for which acquisition was attempted */
```
mfine: woops

JoshuaGross: yeah this is some wonkiness, updating now

JoshuaGross:
I recommend looking at everything except the "regenerate C tests" commit, as that's a lot of noise.

denniszollo:
Maybe this issue should be kept out of customers hands by publishing the latest release tarballs for each language, rather than having them checkout the generator source.

mfine:
Does anything here preclude that?

JoshuaGross:
@denniszollo I specifically added instructions and highlighted how to install from `pypi`, `npm` and `hackage`.

JoshuaGross:
For anyone wanting to run & install this locally regardless, it's really unclear what prerequisites are needed. That's the main thing I changed in the Makefile - it will fail with a very specific message telling you to install pdflatex, python, pip, stack, whatever - before it would fail cryptically without a hint.

python/README.md
```diff
@@ -1,10 +1,18 @@
-SBP Bindings for Python
-=======================
+# SBP Bindings for Python
```
mookerji: ACK ACK ACK: It's RST for pypi!

mookerji:
@JoshuaGross revert the rst change and feel free to merge at will.

python/README.md
JoshuaGross: worst

PR #339, opened 2016-03-02
Add message to output NDB updates

rge-exafore:
This commit adds definition of message ID and data type to
output NDB changes over SBP.

### DIFF
--- a/c/include/libsbp/piksi.h
+++ b/c/include/libsbp/piksi.h
@@ -187,6 +187,18 @@ typedef struct __attribute__((packed)) {
 } msg_mask_satellite_t;
 
 
+/** Output information element updated in NDB
+ *
+ * This message from Piksi to host carries information on updated
+ * data element in NDB.
+ */
+#define SBP_MSG_NDB_UPDATE     0x0080
+typedef struct __attribute__((packed)) {
+  u32 mask;    /**< Indication of information element type */
+  u8 data[0]; /**< New value of updated information element */
+} msg_ndb_update_t;
+
+
 /** \} */
 
 #endif /* LIBSBP_PIKSI_MESSAGES_H */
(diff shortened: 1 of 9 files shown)


mookerji:
hey @rge-exafore , check out https://github.com/swift-nav/libsbp/pull/310/files. We actually use code generation to create the code from a specification. You should actually update this: https://github.com/swift-nav/libsbp/blob/master/spec/yaml/swiftnav/sbp/piksi.yaml and then use the Makefile to create the new code.

mookerji:
@jacobmcnamee want to chime about the message, since you know more about the navigation DB.

mookerji:
hey @rge-exafore, my advice is sync up over Skype with @jacobmcnamee about the design of the messages. My feedback is in two areas: 

- *Existing nav messages*. My understanding is that the navigation DB is an implementation detail of how ephemerides/almanacs/etc. are stored, so why not use the existing messages for reporting updates to the DB? For example, if a quantity (such as the almanac) in the NavDB is updated, the piksi would report that quantity anyway. Otherwise, if you report just binary blobs with masks, as is required by `MSG_NDB_UPDATE`, you'll still have to store information elsewhere about how to further decode `data`. 
- *Raw filesystem reading and writing*. I could imagine a case where you want to bulk read/write contents the navDB and don't really care about the underlying content of the data (i.e,. that an array of data is an ephemeris). We have a few good examples of this already, in the bootloader and the file system [messages](https://github.com/rge-exafore/libsbp/blob/i63-nav-db/spec/yaml/swiftnav/sbp/file_io.yaml#L27).  If your goal then is bulk read/write, you could introduce a set of messages like the `file_io` messages that are specifically for the nav DB.

mookerji:
Closing in favor of consuming the existing messages.

PR #340, opened 2016-03-09
Don't require python testing packages for installation

peddie:
WIP


peddie:
OK, this is at last no longer a work in progress. @mookerji @JoshuaGross

mookerji:
Cute, @peddie. Very cute.

PR #348, opened 2016-05-03
SBP C: Introduce "global" callback that will handle all message types

ejconlon:
Sometimes you just want to handle all message types and you don't want to
to give libsbp a giant whitelist. This review introduces a "global"
callback type: simply pass SBP_GLOBAL_CALLBACK (0x00) as the msg_type in
sbp_register_callback and the callback will be invoked for all messages.

/cc @swift-nav/skylark @fnoble @gsmcmullin @jacobmcnamee

### DIFF
--- a/c/include/libsbp/sbp.h
+++ b/c/include/libsbp/sbp.h
@@ -36,6 +36,9 @@
 /** Default sender ID. Intended for messages sent from the host to the device. */
 #define SBP_SENDER_ID 0x42
 
+/** Fake message id that indicates a handler should be invoked for every message type. */
+#define SBP_GLOBAL_CALLBACK 0x00
+
 /** SBP callback function prototype definition. */
 typedef void (*sbp_msg_callback_t)(u16 sender_id, u8 len, u8 msg[], void *context);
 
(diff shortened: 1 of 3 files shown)


JoshuaGross:
Instead of introducing a new message ID, why not refactor it to have a global callback register function?

ejconlon:
Because that's a tougher change to make!

mfine:
:+1: on not using a message id for this.

mfine:
Or consider a callback mask - any message in the mask will head to the callback.

mookerji:
Adding global message handling isn't the worst idea, but what @JoshuaGross is saying here is correct. 

We actually have a global message callback handler in the Python library, so we should actually just have one in the C library as well. The implications of adding a new concept of virtual message handling as a concept through a protected msg_type value is a substantial conceptual one. (not that new concepts are bad, just that we've typically discussed changes like this more in the past). Quickly looking at https://github.com/swift-nav/libsbp/blob/master/c/src/sbp.c, it does indeed seem like a tricky, but not impossible change.

That being said, I'd like to see an example use case of where this is needed, because our existing firmware handling certainly hasn't needed it.

ejconlon:
ACK. It's not needed in firmware, it's needed in test code.

ejconlon:
In trying to implements the callbacks for this, I realized we actually need the callback type to change, adding a msg_type argument.  I am just going to close this since that change has a huge surface area.

PR #349, opened 2016-05-03
Catch UnicodeDecodeError

mfine:
Other advice is to pass `ensure_ascii=False` to `json.dumps`.

/cc @jacobmcnamee


mookerji:
👍 

Do we have a log demonstrating this? If so, we could/should add it as a test.

mfine:
@mookerji no sample log I can find... it's coming out of the new bootloader spewing things - will try and create something that will fail.

mfine:
Added a test case. Think it's catching this, but not totally sure.

PR #351, opened 2016-05-04
WIP: SBP Haskell: Add QuickCheck instances for messages and some tests

ejconlon:


ejconlon:
This is to debug some conversion failures.  So far nothing has come up except for the fact that we need manual instances for some of these, unfortunately.

PR #353, opened 2016-05-05
Uart state 2

denniszollo:
resurrecting my uart_state pull request and closes #329.  This is to add an observation period field to the uart_state message as to keep stats not only on latency but also dropped packets.

Deprecate the UART state message and create a new one that adds in the "period of reception" info about observation messages from the base as a "latency" object.

This gives us an idea of the link quality back to the base station. The latency stuff only works for messages we received and decoded, it doesn't take into account any dropped packets.   I'm holding off generating the PDF documentation until we make a libsbp release with the upcoming firmware release.   There will be a few more changes.


denniszollo:
@mfine @mookerji @JoshuaGross  Anytime to review this one?  I had to resurrect it from long past and We've now decided to merge this feature into firmware and console to support radio link testing.

Should I also increment the version number to allow specification in requirements.txt?

mookerji:
looking!

spec/yaml/swiftnav/sbp/piksi.yaml
```diff
@@ -168,6 +168,33 @@ definitions:
...
+            desc: State of UART FTDI (USB logger)
+        - latency:
+            type: Latency
+            desc: UART communication latency
+        - obs_period:
+            type: Latency
+            desc: Observation receipt period
+
+
+ - MSG_UART_STATE_DEPA:
```
mookerji: Can you change the descriptions to say `Deprecated`?

https://github.com/denniszollo/libsbp/blob/uart_state_2/spec/yaml/swiftnav/sbp/observation.yaml#L281

mookerji:
Update the description and feel free to merge. I think you'll want to bump sbp for Python and deploy it to pypi. Let me know if you need help with it?

mfine:
@denniszollo only issue is `Latency` the right measurement of a period? Latency speaks to an observed period and has the fields for max, min, avg, and smoothed - is that how the observation period should be reported? Or should there be a single measurement?

denniszollo:
Based upon Mark's feedback I'm going to rename things for clarity.  I'll also update the fields in the deprecated message to deprecated.

denniszollo:
This is ready for review again and should now incorporate the prior feedback - I am planning to wait until I do the new observation message for a pypi push and will need help for that part.

spec/yaml/swiftnav/sbp/piksi.yaml
```diff
@@ -140,6 +140,32 @@ definitions:
...
+            units: ms
+            desc: Minimum period
+        - pmax:
+            type: s32
+            units: ms
+            desc: Maximum period
+        - current:
+            type: s32
+            units: ms
+            desc: Smoothed estimate of the current period
```
mfine: :+1: Nice job!

mfine:
Changes look good!

PR #354, opened 2016-05-19
Cors obs

denniszollo:
need new obs message.  After merge need a new veresion and pypi release.

Note that scaling has changed on pseudorange from 1 cm to 2cm for count.

Will require update to anything using sbp eventually:
piksi_tools (done in PR: https://github.com/swift-nav/piksi_tools/pull/273)
sbp_log_analysis (need to do)
gnss_analysis (need to do)

/cc @mookerji @mfine @JoshuaGross  any time to review?


mookerji:
Yeah, looking.

mookerji:
I'd say that we update the descriptions as well to describe why these messages have changed. Then we'll have to 1. release the library, 2. update the tools, 3. package release candidate for tools, 4. release firmware.

haskell/src/SwiftNav/SBP/Observation.hs
```diff
@@ -686,6 +690,44 @@ $(deriveJSON defaultOptions {fieldLabelModifier = fromMaybe "_packedObsContentDe
...
 $(makeLenses ''PackedObsContentDepA)
 
+-- | PackedObsContentDepB.
+--
+-- Pseudorange and carrier phase observation for a satellite being tracked.
+-- Pseudoranges are referenced to a nominal pseudorange.
+data PackedObsContentDepB = PackedObsContentDepB
+  { _packedObsContentDepB_P  :: Word32
+    -- ^ Pseudorange observation
+  , _packedObsContentDepB_L  :: CarrierPhase
```
ljbade: Can we add note on carrier phase sign? Also should have units

haskell/src/SwiftNav/SBP/Observation.hs
```diff
@@ -686,6 +690,44 @@ $(deriveJSON defaultOptions {fieldLabelModifier = fromMaybe "_packedObsContentDe
              ''PackedObsContentDepA)
 $(makeLenses ''PackedObsContentDepA)
 
+-- | PackedObsContentDepB.
+--
+-- Pseudorange and carrier phase observation for a satellite being tracked.
+-- Pseudoranges are referenced to a nominal pseudorange.
+data PackedObsContentDepB = PackedObsContentDepB
+  { _packedObsContentDepB_P  :: Word32
```
ljbade: Units

haskell/src/SwiftNav/SBP/Observation.hs
```diff
@@ -686,6 +690,44 @@ $(deriveJSON defaultOptions {fieldLabelModifier = fromMaybe "_packedObsContentDe
...
+    -- ^ Pseudorange observation
+  , _packedObsContentDepB_L  :: CarrierPhase
+    -- ^ Carrier phase observation
+  , _packedObsContentDepB_cn0 :: Word8
```
ljbade: Units

haskell/src/SwiftNav/SBP/Observation.hs
denniszollo: This is a type, I can't comment about the units.

haskell/src/SwiftNav/SBP/Observation.hs
denniszollo: same- the units are some weird fixed point format.

spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -33,22 +33,6 @@ definitions:
...
-        - i:
-            type: s32
-            units: cycles
-            desc: Carrier phase whole cycles
-        - f:
-            type: u8
-            units: cycles / 256
-            desc: Carrier phase fractional part
-
  - ObservationHeader:
```
mfine: You're changing the contents of `ObservationHeader` but continuing to use it from the deprecated messages. You'll need a deprecated `ObservationHeader` structure s well to preserve the functioning of the deprecated messages.

spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -543,4 +588,150 @@ definitions:
...
+        - obs:
+            type: array
+            fill: PackedObsContentDepB
+            map_by: sid
+            desc: |
+              Pseudorange and carrier phase observation for a
+              satellite being tracked.
 
+ - MSG_EPHEMERIS_DEP_C:
+    id: 0x0047
```
mfine: Bummer if this is only for an `id` change to get it in a different mask.

spec/yaml/swiftnav/sbp/observation.yaml
denniszollo: I didn't change the header.....

mfine: Woops, missed the separator between the next set of changes for `PackedObsContent`. Sorry.

mookerji: @mfine, neither the interface nor content of the fields is changing. What is changing is the way those fields in the observations are computed, so I don't think that needs to change.

mfine: Yeah, I don't think it needs to change - I misread the diff and incorporated `PackedObsContent` changes into the observation header structure.

spec/yaml/swiftnav/sbp/observation.yaml
denniszollo: It is only for hte ID, but I have had big problems not being able to mask of the comparatively large ephemeris message so there is value in it.

denniszollo:
This is ready but is still waiting on a few more unit tests after we have compatible firmware.

denniszollo:
deprecated in favor of : https://github.com/swift-nav/libsbp/pull/356

PR #355, opened 2016-05-25
Added back in the file driver updated for current design

axlan:
It throws an IOEerror on write attempts


axlan:
I'm closing this since it was pointed out that the File read/write functions can be used directly without a driver.

PR #356, opened 2016-05-26
Cors obs bare bones

denniszollo:
Go back to the simple SBP change.

Question - should this be version 0.53? or 1.0?


mookerji:
1.0, per our versioning policy.

PR #366, opened 2016-08-04
Remove delta

mfine:
Delta is derivable from the timestamps. Don't include it in the logs.

/cc @jkretzmer


JoshuaGross:
:+1:

So, it looks like the Java version also has a `delta()` function and output. Can you remove it there too?

mfine:
Good catch. Removed.

PR #369, opened 2016-08-09
Add device monitor message.

wltr:
This message adds support for internal voltage and die temperature measurements on Zynq based devices.

@mfine @denniszollo @switanis


spec/yaml/swiftnav/sbp/piksi.yaml
```diff
@@ -282,3 +282,28 @@ definitions:
...
+            type: float
+            units: V
+            desc: Processor VCCINT
+        - cpu_vccaux:
+            type: float
+            units: V
+            desc: Processor VCCAUX
+        - cpu_temperature:
+            type: float
+            units: C
```
ljbade: degrees C

spec/yaml/swiftnav/sbp/piksi.yaml
```diff
@@ -282,3 +282,28 @@ definitions:
...
+            desc: Processor temperature
+        - fe_temperature:
+            type: float
+            units: C
```
ljbade: degrees C

ljbade:
Nice

JoshuaGross:
Looks good, I would also like to see more specific units like Leith indicated. Also, run the code generators if you can.

switanis:
1. I strongly recommend to use scaled integers for passing values instead of floats. Integers can be shorter, are easier to deal and examine, clearly define range and resolution, are not subject of rounding, faster to manipulate. For voltage we can use u16 scaled by 1000. That will give range 0 to 65.535 V with 1mV resolution. Or make it s16 to cover from -32.768 to 32.767 V with 1mV res. For temp we can use s16 scaled by 100. That will give us range -327.68 (really -273.15) to 327.67 deg C with 0.01 deg C resolution. 

2. Can we measure board supply voltage as well? That'd be the most useful for customers.

wltr:
Thanks for the comments! 

**Updates:**
- Use `degree C`
- Use `s16` with scaling factor
- Add `dev_vin` to monitor VIN on the GNSS module (has yet to be implemented in firmware)

@ljbade @JoshuaGross @switanis

JoshuaGross:
Looks good. Going to fix the merge conflict and pull in manually.

PR #371, opened 2016-08-11
add back in deprecated observation message

denniszollo:
@jkretzmer.  How do we communicate this?  Please take a look at the sbp doc and see if it makes sense to you.


JoshuaGross:
Could you un-deprecate it by adding back a different message? Not sure what the best thing to do is.

PR #377, opened 2016-09-09
Device monitor message: Fix VIN scaling.

wltr:
Realized this is board dependent and will be taken care of somewhere else. SBP message should be just a generic voltage level applicable to any device.

/cc @JoshuaGross

### DIFF
--- a/spec/yaml/swiftnav/sbp/piksi.yaml
+++ b/spec/yaml/swiftnav/sbp/piksi.yaml
@@ -292,7 +292,7 @@ definitions:
     fields:
         - dev_vin:
             type: s16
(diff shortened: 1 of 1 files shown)


JoshuaGross:
:+1:

I think this requires a major version bump since we're changing the meaning of the field substantially.

JoshuaGross:
Minor version bump, rather, as opposed to patch version.

JoshuaGross:
This field has already made it into a previous SBP release, right?

wltr:
Yes, it was in 1.1.3, but I shouldn't have put the `/ 11` in the unit description.

PR #378, opened 2016-09-09
Move from timestamp to time

mfine:
Move from time to timestamp so we can differentiate.

/cc @margaret


mfine:
I don't think we should change the meaning of timestamp. Let's start with a new field, `time`, so that breakage will be clear. I don't think it's worth the effort to update all the data logs used in the tests - just going to skip the multi json test.

mfine:
Fixed all the logs and tests. Ugh.

PR #379, opened 2016-09-13
Add reacq-related SBP message

dt-exafore:
https://github.com/swift-nav/exafore_planning/issues/326
Changes according to design: https://swiftnav.hackpad.com/Piksi-v3-Re-Acquisition-Control-Design-Phase-1-Qia9kaydkvP#:h=Debugging

### DIFF
--- a/spec/yaml/swiftnav/sbp/acquisition.yaml
+++ b/spec/yaml/swiftnav/sbp/acquisition.yaml
@@ -70,3 +70,68 @@ definitions:
             desc: |
               PRN-1 identifier of the satellite signal for which
               acquisition was attempted
+
+ - AcqSvProfile:
+    short_desc: Acq perfomance measurement and debug
+    desc: |
+      Profile for a specific SV for debugging purposes
+      The message describes SV profile during acquisition time.
+      The message is used to debug and measure the performance.
+    fields:
+        - job_type:
+            type: u8
+            desc: SV search job type (deep, fallback, etc)
+        - status:
+            type: u8
+            desc: Acquisition status 1 is Success, 0 is Failure
+        - cn0:
+            type: u16
+            desc: CN0 value. Only valid if status is '1'
+            units: dB-Hz*10
+        - int_time:
+            type: u8
+            units: ms
+            desc: Acquisition integration time
+        - sid:
+            type: GnssSignal
+            desc: GNSS signal for which acquisition was attempted
+        - bin_width:
+            type: u16
+            units: Hz
+            desc: Acq frequency bin width
+        - timestamp:
+            type: u32
+            units: ms
+            desc: Timestamp of the job complete event
+        - time_spent:
+            type: u32
+            units: us
+            desc: Time spent to search for sid.code
+        - cf_min:
+            type: u32
+            units: Hz
+            desc: Doppler range lowest frequency
+        - cf_max:
+            type: u32
+            units: Hz
+            desc: Doppler range highest frequency
+        - cf:
+            type: u32
+            units: Hz
+            desc: Doppler value of detected peak. Only valid if status is '1'
+        - cp:
+            type: u32
+            units: chips*10
+            desc: Codephase of detected peak. Only valid if status is '1'
+
+ - MSG_ACQ_SV_PROFILE:
+    id: 0x001E
+    short_desc: Acquisition perfomance measurement and debug
+    desc: |
+      The message describes all SV profiles during acquisition time.
+      The message is used to debug and measure the performance.
+    fields:
+        - acq_sv_profile:
+            type: array
+            fill: AcqSvProfile
+            desc: SV profiles during acquisition time
--- a/c/include/libsbp/acquisition.h
+++ b/c/include/libsbp/acquisition.h
@@ -63,6 +63,39 @@ acquisition was attempted
 } msg_acq_result_dep_a_t;
 
 
+/** Acq perfomance measurement and debug
+ *
+ * Profile for a specific SV for debugging purposes
+ * The message describes SV profile during acquisition time.
+ * The message is used to debug and measure the performance.
+ */
+typedef struct __attribute__((packed)) {
+  u8 job_type;      /**< SV search job type (deep, fallback, etc) */
+  u8 status;        /**< Acquisition status 1 is Success, 0 is Failure */
+  u16 cn0;           /**< CN0 value. Only valid if status is '1' [dB-Hz*10] */
+  u8 int_time;      /**< Acquisition integration time [ms] */
+  sbp_gnss_signal_t sid;           /**< GNSS signal for which acquisition was attempted */
+  u16 bin_width;     /**< Acq frequency bin width [Hz] */
+  u32 timestamp;     /**< Timestamp of the job complete event [ms] */
+  u32 time_spent;    /**< Time spent to search for sid.code [us] */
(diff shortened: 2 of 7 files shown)


spec/yaml/swiftnav/sbp/acquisition.yaml
```diff
@@ -70,3 +70,50 @@ definitions:
...
+ - MSG_ACQ_SV_PROFILE:
+    id: 0x001E
+    short_desc: Reacq perfomance measurement and debug
+    desc: |
+      The message describes SV profile during acquisition time.
+      The message is useful to debug and measure the performance of
+      the new (re-)acq design.
+    public: true
+    fields:
+        - N:
```
mfine: How does this message get laid out? This field refers to a count of SV's to be included in the message - yet there's no subsequent collection of anything. There are other messages in SBP to reference where there are collections of message objects, e.g., MSG_OBS, MSG_TRACKING_IQ, MSG_TRACKING_STATE.

denniszollo: I think the acquisition can only do on SV at a time, so no need to have an array.

spec/yaml/swiftnav/sbp/acquisition.yaml
```diff
@@ -70,3 +70,50 @@ definitions:
...
+            units: Hz
+            desc: Doppler range lowest frequency
+        - cf_max:
+            type: u32
+            units: Hz
+            desc: Doppler range highest frequency
+        - cf:
+            type: u32
+            units: Hz
+            desc: Doppler value of detected peak
```
denniszollo: shouldn't you put some kind of CN0 in for the detected peak as well?

spec/yaml/swiftnav/sbp/acquisition.yaml
ljbade: yeah it would make sense to send multiple messages since each acq event is independent of the other events

adel-mamin: The motivation behind sending the acq results for multiple SVs in one SBP was to reduce the overhead related to SBP header and footer, which is 8 bytes in total. One SV acq data is 21 bytes. So the overhead is 8/23 -> 38%. Combining multiple acq data SVs into one SBP can reduce it to 8/255 -> ~3%. We expect quite a large number of these messages to be send per second, when there are few tracked SVs.

adel-mamin: @mfine. Agree, that needs to be fixed.

spec/yaml/swiftnav/sbp/acquisition.yaml
adel-mamin: @denniszollo Agree, CN0 would be useful and needs to be added. There is already a message MSG_ACQ_RESULT I updated the re-acq design document accordingly.

spec/yaml/swiftnav/sbp/acquisition.yaml
```diff
@@ -70,3 +70,79 @@ definitions:
             desc: |
               PRN-1 identifier of the satellite signal for which
               acquisition was attempted
+
+ - AcqSvProfile:
+    short_desc: Reacq perfomance measurement and debug
```
adel-mamin: Let's replace 'Reacq' with 'Acq' as this messages can be used for general acquisition performance analysis and debugging.

spec/yaml/swiftnav/sbp/acquisition.yaml
```diff
@@ -70,3 +70,79 @@ definitions:
...
+    desc: |
+      Profile for a specific SV for debugging purposes
+      The message describes SV profile during acquisition time.
+      The message is useful to debug and measure the performance of
+      the new (re-)acq design.
```
adel-mamin: Let's rephrase it a bit: 'The message is used to debug and measure the performance of acquisition.'

spec/yaml/swiftnav/sbp/acquisition.yaml
```diff
@@ -70,3 +70,79 @@ definitions:
...
+            units: Hz
+            desc: Doppler range lowest frequency
+        - cf_max:
+            type: u32
+            units: Hz
+            desc: Doppler range highest frequency
+        - cf:
+            type: u32
+            units: Hz
+            desc: Doppler value of detected peak
```
adel-mamin: Let's add: 'Only valid if status is '1'

spec/yaml/swiftnav/sbp/acquisition.yaml
```diff
@@ -70,3 +70,79 @@ definitions:
...
+        - cp:
+            type: u32
+            units: chips*10
+            desc: Codephase of detected peak
```
adel-mamin: Let's add: 'Only valid if status is '1'

spec/yaml/swiftnav/sbp/acquisition.yaml
```diff
@@ -70,3 +70,79 @@ definitions:
...
+
+ - MSG_ACQ_SV_PROFILE:
+    id: 0x001E
+    short_desc: Reacq perfomance measurement and debug
+    desc: |
+      The messaVge describes all SV profiles during acquisition time.
+      The message is useful to debug and measure the performance of
```
adel-mamin: Let's rephrase it the same way as above.

spec/yaml/swiftnav/sbp/acquisition.yaml
```diff
@@ -70,3 +70,77 @@ definitions:
...
+            units: Hz
+            desc: Doppler value of detected peak. Only valid if status is '1'
+        - cp:
+            type: u32
+            units: chips*10
+            desc: Codephase of detected peak. Only valid if status is '1'
+
+ - MSG_ACQ_SV_PROFILE:
+    id: 0x001E
+    short_desc: Reacq perfomance measurement and debug
```
adel-mamin: Re-acq -> Acquistion

adel-mamin:
Looks good to me.

spec/yaml/swiftnav/sbp/acquisition.yaml
```diff
@@ -70,3 +70,77 @@ definitions:
...
+            desc: Codephase of detected peak. Only valid if status is '1'
+
+ - MSG_ACQ_SV_PROFILE:
+    id: 0x001E
+    short_desc: Acquisition perfomance measurement and debug
+    desc: |
+      The message describes all SV profiles during acquisition time.
+      The message is used to debug and measure the performance.
+    fields:
+        - N:
```
mfine: The idiom in sbp has been to use "fill arrays" for messages like this - instead of keeping track of the number of elements in the array with a count, e.g., `N`, you just have the array, and the clients eat up as many array elements at they can in the buffer. For example from the obs messages:

```yaml
        - obs:
            type: array
            fill: PackedObsContent
```

@mookerji what do you think?

mookerji: yeah, I think you can just leave the last field as an array and it'll get globbed up. I recommend just using that to stay consistent.

dt-exafore: Guys, thanks for your comments. Please, check again I removed N from the structure.

mfine: Looks good! :+1:

PR #380, opened 2016-09-16
[JS] toJSON method is very wrong - it should return an object to be stringified

JoshuaGross:


JoshuaGross:
also added a variety of example scripts

mookerji:
👍

PR #381, opened 2016-09-16
Add a whitelist option to the javascript library. Can speed up dispatching substantially

JoshuaGross:
Add a whitelist option to the javascript library. Can speed up decoding/dispatching substantially if you know you want to ignore some messages. Basically if there's a whitelist and a message is filtered out, it isn't fully decoded - saving some CPU cycles. 

Was dropping >90% of position messages previously at 10hz, now dropping 0% of the position messages. YMMV depending on how aggressively you're filtering.

### DIFF
--- a/javascript/benchmark_dispatch_whitelist.js
+++ b/javascript/benchmark_dispatch_whitelist.js
@@ -1,0 +1,46 @@
+// Benchmark how much time is saved in dispatch by ignoring some messages (in this case half)
+var Benchmark = require('benchmark');
+var stream = require('stream');
+var path = require('path');
+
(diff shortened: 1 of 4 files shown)


JoshuaGross:
cc @mfine @mookerji

JoshuaGross:
Added a benchmark and some unit tests. The benchmark doesn't run automatically, but currently shows that filtering out half of your messages (when only decoding two messages) has about an 18% performance increase. Should scale with the ratio of messages that are being filtered out before decode.

mookerji:
👍 !

PR #383, opened 2016-10-12
Correct data types in reacq SBP

dt-exafore:
Based on https://github.com/swift-nav/exafore_planning/issues/400


dt-exafore:
@adel-mamin Please, take a look

PR #386, opened 2016-10-13
Proposed changes before launch

cbeighley:


pre-piksi-multi-launch-updates.md
```diff
@@ -0,0 +1,121 @@
...
+  - DOPS (to add)
+  - Pos ECEF
+  - Pos LLH
+  - Baseline ECEF
+  - Baseline NED
+  - Vel ECEF
+  - Vel NED
+  - Baseline heading 
+- Have flags be consistent across all solution messages (except for GPS time)
+  - 0 = Fix not available or invalid
```
mookerji: Are these the NMEA flags?

pre-piksi-multi-launch-updates.md
```diff
@@ -0,0 +1,121 @@
...
+these and implement them before shipping Piksi Multi.
+
+These changes attempt to accomplish a few things:
+- Make messages more bandwidth efficient
+- Make user side logic of using messages more obvious
+- Remove state from the protocol, where possible
+
+### 1. Replace obs forwarding with general purpose forwarding message
+
+Current there is a convention in the protocol that if the sender_id of a
```
mookerji: Can you add an example of what the forwarded wrapper looks like? If we're going to be explicit about forwarding messages like this, then we should also add a TTL field. @mfine 

We're also going to have to make sure that the libsbp code generator can properly handle this case.

pre-piksi-multi-launch-updates.md
```diff
@@ -0,0 +1,121 @@
...
+MSG_OBS is 0, it means it is a forwarded observation from another device. We
+propose to replace this with a general purpose forwarding message that wraps
+the forwarded message, discarding the preamble and checksum of the forwarded
+message. This way, the sender_id of the forwarded message will be preserved and
+will remove the need for special handling of sender_id = 0 on obs messages.
+
+### 2. sender_id = device_uuid & 0xFFFF
+
+We need to decide where sender_id will come from on Piksi Multi. In deciding
```
mookerji: 👍

pre-piksi-multi-launch-updates.md
```diff
@@ -0,0 +1,121 @@
...
+  - 0 = Fix not available or invalid
+  - 1 = GPS SPS Mode, fix valid
+  - 2 = Differential GPS, SPS Mode, fix valid
+  - 3 = GPS PPS Mode, fix valid
+  - 4 = Real Time Kinematic. System used in RTK mode with fixed integers
+  - 5 = Float RTK. Satellite system used in RTK mode, floating integers
+  - 6 = Estimated (dead reckoning) Mode
+  - 7 = Manual Input Mode
+  - 8 = Simulator Mode 
+- GPS time: 3 bit time fix field, rest of flag bits reserved:
```
mookerji: We should finally get around to adding support for enum flags in the SBP clients.

pre-piksi-multi-launch-updates.md
```diff
@@ -0,0 +1,121 @@
...
+  - 0: none
+  - 1: coarse
+  - 2: fine
+  - 3 - 255: reserved
+
+### 6. Add UTC time message
+
+Users want UTC time rather than GPS time.
```
mookerji: Are we outputting both?

pre-piksi-multi-launch-updates.md
```diff
@@ -0,0 +1,121 @@
...
+
+- u8 flags  (0-3: Source 0: No DGPS 1: SBAS 2: Code Difference 3: RTK 4: PPP, 4-8 reserved)   
+- u8[6] Station ID
+- u16 age_of_corrections (deci-seconds)  0.1 seconds
+- u8 number of GPS satellites from base station
+- u8[3] reserved
+
+### 8. Add NAV complete message
+
+Sent after rest of solution messages to indicate all solution message for epoch
```
mookerji: This does not seem "stateless" to me.

pre-piksi-multi-launch-updates.md
```diff
@@ -0,0 +1,121 @@
...
+have been sent.
+
+### 9. Add fields to MSG_STARTUP
+- Cause of start - Power on, software reset, watchdog, hardware reset
+- Start type: cold, warm, hot
```
mookerji: Is including the state of the acquisition (cold/warm/hot) stuff typical? Do we want to convey that information or is it debug state?

pre-piksi-multi-launch-updates.md
```diff
@@ -0,0 +1,121 @@
...
+  - 2: fine
+  - 3 - 255: reserved
+
+### 6. Add UTC time message
+
+Users want UTC time rather than GPS time.
+
+### 7. Add DGPS status message
+
+- u8 flags  (0-3: Source 0: No DGPS 1: SBAS 2: Code Difference 3: RTK 4: PPP, 4-8 reserved)
```
mookerji: This status message seems redundant.

- The solution status flag should cover this. 
- Station id is not a bad idea, but why not just have it be a fixed-size string? 
- Using age-of-corrections already reported using latency?
- The number of base station satellites can be derived from the base station observation message.

pre-piksi-multi-launch-updates.md
mookerji: Or just one (`MSG_UTC_TIME`).

pre-piksi-multi-launch-updates.md
```diff
@@ -0,0 +1,121 @@
...
+- `MSG_FLASH_READ_REQ`
+- `MSG_FLASH_READ_RESP`
+- `MSG_FLASH_PROGRAM`
+- `MSG_FLASH_ERASE`
+- `MSG_FLASH_DONE`
+- `MSG_STM_FLASH_LOCK_SECTOR`
+- `MSG_STM_FLASH_UNLOCK_SECTOR`
+- `MSG_M25_FLASH_WRITE_STATUS`
+
+The following messages could be part of a device ID message that is
```
mookerji: But we can't get rid of polling messags, because we still need to have some polling-based messages (like settings), so why not just keep these DNA, etc. messages similarly accessible?

pre-piksi-multi-launch-updates.md
mfine: The idea was to encapsulate a SBP message in the payload of an SBP message - no preamble, no crc. This should work ok with the parser, though how to express that in the yaml is not clear - it would need to be a special rule or something.

And I don't think this message would be forwarded on, but not sure about that. If it were to be forwarded on, would need to TTL for sure. If not, would be another exceptional rule to cover this case.

Maybe all these exceptions make this not such a sound idea.

pre-piksi-multi-launch-updates.md
mfine: But the polling or REQ/RESP messages are only confined now to FILEIO/SETTINGS - we should try and constrain it to those operations and not keep around these older legacy REQ/RESP messages.

pre-piksi-multi-launch-updates.md
JoshuaGross: SBP isn't really stateless, anyway - for example, `MSG_POS_ECEF` and `MSG_POS_LLH` contain `tow` but not `wn`, which you need to get from the preceding `MSG_GPS_TIME` message.

I'd like additional context for why this should be added, regardless.

pre-piksi-multi-launch-updates.md
JoshuaGross: 1. It seems like this could be achieved by pointing people to simple conversion functions. We already have a JavaScript library that does all of these conversions - and the code would be super simple to port. https://github.com/swift-nav/gpstime

2. More bandwidth required for this message doesn't seem great.

3. I don't see how this would work - if we add a UTC time message then the `tow` of messages that come after it wouldn't be interoperable with the message, anyway, since leap-seconds is the issue with conversion between GPS and UTC. If the proposal is to remove `tow` from all other messages and require that people use `MSG_UTC_TIME` and `MSG_GPS_TIME` statefully to recover the time of messages, I can sort of see that making sense as a use-case, but that's also not a good idea.

pre-piksi-multi-launch-updates.md
mookerji: I would argue that those are "approximately" stateless. The end-user in a real-time setting will not need to know week number.

pre-piksi-multi-launch-updates.md
```diff
@@ -0,0 +1,121 @@
...
+- `MSG_STM_UNIQUE_ID_REQ`
+- `MSG_STM_UNIQUE_ID_RESP`
+
+### 4. Observation message updates
+
+Updates to observation message to reduce bandwidth and allow observations to
+always be sent regardless of whether they're usable (receiving device filters
+them for usability rather than sending device).
+
+- Remove week number (2 bytes)
```
ljbade: This is not a good idea, we have enough headaches in Skylark with RTCM not having ToW and makes it very hard to convert an SBP log to JSON without requiring the user to input the date they made the measurements.

pre-piksi-multi-launch-updates.md
```diff
@@ -0,0 +1,121 @@
...
+- Replace 4 byte obs.sid with 2 bytes, matching the firmware SID representation
```
ljbade: If you change this, change it in every other SBP message that uses SID for consistency.

pre-piksi-multi-launch-updates.md
```diff
@@ -0,0 +1,121 @@
...
+  - Satellite ID (1 byte)
+    - Should we make this actual satellite ID value, not value minus 1?
+  - Signal (Band/Constellation/Etc) Code (1 byte)
+- Replace 2 byte lock counter with 1 byte, conforming to DF402 from RTCM 10403.2
+- Add 1 byte of flags for each satellite - need to talk to Estimation and Exafore about what these should be
```
ljbade: in fimware we are using a u16 for flags
other flags I plan to add:
- Psuedorange valid
- Clock jump
- Phase lock indicator
- RTK OK
- SPP OK

pre-piksi-multi-launch-updates.md
```diff
@@ -0,0 +1,121 @@
...
+  - Bit 0: Time fix valid (or code phase valid)
+  - Bit 1: Carrier phase valid
+  - Bit 2: Half cycle ambiguity resolved
+  - Bits 3-7: Reserved
+- Have observations be sent out at all times, use flags to indicate validity
```
ljbade: I think he means all tracked sats, it will be up to RTK filer etc, to read the status flags to decide if to use it

pre-piksi-multi-launch-updates.md
ljbade: @JoshuaGross the relationship between GPS and UTC is not a simple number of seconds, there is also a polynomial correction equation you need to apply from the GPS navigation data stream

ljbade: I believe we should output both GPS and UTC, and allow the customer to turn on/off the messages. This matches the functionality of ever other GNSS receiver on the market.

pre-piksi-multi-launch-updates.md
ljbade: @JoshuaGross Can we request adding WN to POS ECEF and LLH, I hate not having WN available, every message should be self contained for log analysis.

ljbade:
From my other PR: https://github.com/swift-nav/libsbp/pull/384
- fix the acquisition message to use CN/0, and remove that weird comment about SNR

Add WN to messages which don't have it, and add keep it in observatoins. Sure not needed for real time operations, but makes it a pain when working with SBP logs. Also makes conversion to RINEX difficult. Although perhaps we could rely on the GPS time message when converting SBP logs to RINEX? (And just require user to enter date if not missing).

Add SNR instead of CN/0 for observations (or have both). @valeri-atamaniouk explained nicely to me that he wants to add SNR for use in the RTK filter as it is a more reliable indicator of measurement quality than CN/0 since CN/0 does not reflect changes in tracking loop integration length.

Add a way to get measured doppler. Since not needed by RTK (and not usable anyway according to @anth-cole) we should but this in a separate optional message. This would be used to generate RINEX logs, and for analysing this value which is used in SPP velocity solving. Would make it easier to fix bugs/issues with velocity.

Use receiver time not GPS time for observations. This is a key requirement for full RINEX compatibility. However until we have clock steering (@colin would it be possible to make this requirement for final release?) we need higher than ms precision. I am open to using a fixed point type solution here to reduce the number of bytes needed for nanosecond resolution.

pre-piksi-multi-launch-updates.md
cbeighley: @mookerji Yes this would be identical to NMEA

ljbade:
Two more things:

Optimising the observation header will have little impact compared to changes to the per-satellite fields, particulary when we add GLONASS support and might be sending up to 40 sats measurements each epoch. Thus saving 2 bytes for WN seems pointless.

In the POS LLH message there is a status flag that indicates Piksi can output either ellipsoidal or geiod height. Piksi has never had a geoid, and thus never supported geoid heights. Can we rename this field to reserved until we can implement a geoid? Will avoid customer confusion.

pre-piksi-multi-launch-updates.md
ljbade: hmm Ublox does at least. it might be useful to customers, e.g. a car, if it needs to know roughly how long it should expect to get an RTK lock, in case it had some sort of indication to the user?

pre-piksi-multi-launch-updates.md
switanis: Agree with @ljbade - we should have week number there.

pre-piksi-multi-launch-updates.md
switanis: Would be good to write down various use cases when messages are forwarded and used.
Would be good to know from what port data is forwarded from.
We should make it to be able to forward data in other protocols (NMEA0183, NMEA2000, RTCMv3,...).

ljbade:
I forgot to note here, @anth-cole is hoping we can check if latest firmware can recreate computed doppler from base SBP observations accurately. We should run this test before we lock down the new messages in case changes are needed.

There is also the possibility we will need lock_time so RTK can use that to judge observation quality, although perhaps we can calculate this at SBP reception.

pre-piksi-multi-launch-updates.md
fnoble: This is an extremely bad idea as others have stated, it encourages a very bad implementation pattern that is fragile and relies on guaranteed delivery of messages.

pre-piksi-multi-launch-updates.md
fnoble: Would be good to poll the estimation team on what info they might like to expose here. There could be a flag for why IAR is failing for exmaple

pre-piksi-multi-launch-updates.md
mookerji: We could be really cheeky and have a `MSG_ETA`, with helpful time estimates of when you'll get a fix.

fnoble:
Also we should kill the MSG_IAR_STATE message as its no longer used

pre-piksi-multi-launch-updates.md
cbeighley: Decided in meeting on Oct 18:
- Turn this into a more general forwarding message, with 1 byte for source ID and 1 byte for protocol ID

pre-piksi-multi-launch-updates.md
cbeighley: Decided in meeting on Oct 18:
- Use lower 16 bits of UUID
- Uniquely map serial numbers to UUID
- Hardware sticker will have Serial Number and UUID on it
  - If there's not enough room on sticker we will revisit this
  - @denniszollo will communicate with Ops on this

pre-piksi-multi-launch-updates.md
cbeighley: Decided in meeting on Oct 18:
- Bump message IDs
- 0 is the value for Invalid
- Baseline and Position messages will have same numerical values for Float/Fixed
- Other messages are independent

pre-piksi-multi-launch-updates.md
cbeighley: Decided in meeting on Oct 18:
- Separate message with age of corrections with every epoch
- DGPS state message at lower frequency / when messages arrive
  - Corrections type
  - Correction source ID (string)
  - Latency

pre-piksi-multi-launch-updates.md
cbeighley: Decided in meeting on Oct 18:
- Don't add this.

pre-piksi-multi-launch-updates.md
cbeighley: Decided in meeting on Oct 18:
- Do it.

pre-piksi-multi-launch-updates.md
cbeighley: Decided on Oct 18:
- Use obs as propose, add measured doppler from Leith's proposal here: https://github.com/swift-nav/libsbp/pull/385/files

pre-piksi-multi-launch-updates.md
denniszollo: - Psuedorange valid comes form the time valid flag
- clock jump would be indicated by a reset of the lock time 
- Phase lock are the combination of Bit 1 and Bit 2
- RTK OK and SPP OK are left up to the consumer

denniszollo:
Good point on the GEOD, lets move this back to reserved.  Also, do we want to keep the RAIM repair flags in solutions?

denniszollo:
now superseded by #389

PR #390, opened 2016-10-25
Draft UTC MSG_PVT message.

mookerji:
Closes:
- https://github.com/swift-nav/skylark-tales/issues/425

/cc @fnoble @cbeighley @denniszollo @swift-nav/skylark @ljbade @anth-cole

### DIFF
--- a/spec/yaml/swiftnav/sbp/navigation.yaml
+++ b/spec/yaml/swiftnav/sbp/navigation.yaml
@@ -96,7 +96,6 @@ definitions:
             units: 0.01
             desc: Vertical Dilution of Precision
 
-
  - MSG_POS_ECEF:
     id: 0x0200
     short_desc: Single-point position in ECEF
@@ -472,3 +471,114 @@ definitions:
                     values:
                         - 0: Float RTK
                         - 1: Fixed RTK
+
+ - MSG_PVT:
+    id: 0x0208
+    short_desc: Position velocity time (UTC) solution
+    desc: |
+      The position-velocity-time (PVT) message reports the combined absolute
+      geodetic coordinates, status flags of the position solution, velocity, and
+      full UTC time of the receiver. If the rover receiver knows the surveyed
+      position of the base station and has an RTK solution, this reports a
+      pseudo-absolute position solution using the base station position and the
+      rover's RTK baseline vector.
+    fields:
+        - tow:
+            type: u32
+            units: ms
+            desc: GPS time of week rounded to the nearest millisecond
+        - year:
+            type: u16
+            units: years
+            desc: Year (UTC)
+        - month:
+            type: u8
+            units: months
+            desc: Month of the year (UTC) (ranges from to 1 to 12)
+        - day:
+            type: u8
+            units: days
+            desc: Day of the month (UTC) (ranges from 1 to 31)
+        - mins:
+            type: u8
+            units: minutes
+            desc: Minute of the hour (UTC) (ranges from 0 to 59)
+        - sec:
+            type: u8
+            units: s
+            desc: |
+              Seconds of the minute (UTC) (ranges from 0 to 60). The upper range
+              includes an the possibility of an extra leap second correction.
+        - ns:
+            type: s32
+            units: ns
+            desc: Nanosecond residual of UTC sec (ranges from -1e9 to 1e9)
+        - lat:
+            type: double
+            units: deg
+            desc: Latitude
+        - lon:
+            type: double
+            units: deg
+            desc: Longitude
+        - height:
+            type: s32
+            units: mm
+            desc: Height above WGS84 ellipsoid
+        - heightMSL:
+            type: s32
+            units: mm
+            desc: Height above mean sea level
+        - h_accuracy:
+            type: double
+            units: mm
+            desc: |
+              Horizontal position accuracy estimate (not
+              implemented). Defaults to 0.
+        - v_accuracy:
+            type: u16
+            units: mm
+            desc: |
+              Vertical position accuracy estimate (not
+              implemented). Defaults to 0.
+        - n_sats:
+            type: u8
+            desc: Number of satellites used in solution.
+        - vel_n:
+            type: s32
+            units: mm/s
+            desc: Velocity North coordinate
+        - vel_e:
+            type: s32
+            units: mm/s
+            desc: Velocity East coordinate
+        - vel_d:
+            type: s32
+            units: mm/s
+            desc: Velocity Down coordinate
+        - pdop:
+            type: u16
+            units: 0.01
+            desc: Position Dilution of Precision
+        - flags:
+            type: u8
+            desc: Status flags
+            fields:
+              - 0-2:
+                  desc: Fix mode
+                  values:
+                    - 0x000: Invalid
+                    - 0x001: RTK Float
+                    - 0x010: RTK Fixed
+                    - 0x011: SPP
+                    - 0x100: Reserved
+                    - 0x101: Reserved
+                    - 0x110: Reserved
+                    - 0x111: Reserved
+              - 3-6:
+                  desc: Reserved
+              - 7:
+                  desc: RAIM repair flag
+                  values:
+                    - 0: No RAIM repair
+                    - 1: RAIM repair
--- a/spec/yaml/swiftnav/sbp/navigation.yaml
+++ b/spec/yaml/swiftnav/sbp/navigation.yaml
@@ -581,14 +581,14 @@ definitions:
                     - 0x101: Reserved
                     - 0x110: Reserved
                     - 0x111: Reserved
-              - 3-6:
+              - 3-5:
                   desc: UTC flags
                   values:
-                    - 0x000: Invalid UTC date
-                    - 0x001: Valid UTC date, defaults used
-                    - 0x010: Valid UTC leap seconds
+                    - 0x00: Invalid UTC date
+                    - 0x01: Valid UTC date, defaults used
+                    - 0x10: Valid UTC leap seconds
               - 7:
                   desc: RAIM repair flag
                   values:
-                    - 0: No RAIM repair
-                    - 1: RAIM repair
+                    - 0x0: No RAIM repair
+                    - 0x1: RAIM repair
--- a/spec/yaml/swiftnav/sbp/navigation.yaml
+++ b/spec/yaml/swiftnav/sbp/navigation.yaml
@@ -476,12 +476,18 @@ definitions:
     id: 0x0208
     short_desc: Position velocity time (UTC) solution
     desc: |
+
       The position-velocity-time (PVT) message reports the combined absolute
       geodetic coordinates, status flags of the position solution, velocity, and
       full UTC time of the receiver. If the rover receiver knows the surveyed
       position of the base station and has an RTK solution, this reports a
       pseudo-absolute position solution using the base station position and the
       rover's RTK baseline vector.
+
+      When explicit GPS to UTC time conversion parameters are not available from
+      the GPS navigation message, MSG_PVT will use the best available receiver
+      defaults. As such, UTC seconds may be incorrect by a few seconds, but will
+      be flagged as such.
     fields:
         - tow:
             type: u32
@@ -525,10 +531,10 @@ definitions:
             type: s32
             units: mm
             desc: Height above WGS84 ellipsoid
-        - heightMSL:
+        - height_msl:
             type: s32
             units: mm
-            desc: Height above mean sea level
+            desc: Height above mean sea level (not implemented)
         - h_accuracy:
             type: double
             units: mm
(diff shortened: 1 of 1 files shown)


spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -96,7 +96,6 @@ definitions:
             units: 0.01
             desc: Vertical Dilution of Precision
 
-
```
mookerji: Hmmmmmm

mookerji:
This is a strawman proposal. Please provide feedback.

The current proposed message is 65 bytes. Note that there are two double-precision floats (16 bytes total) for latitude and longitude, that we could arguably replace by two Q16.32 types (12 bytes total) if we wanted to be ~1mm accurate on lat/lons. The combined message total from `MSG_GPS_TIME` (11), `MSG_POS_ECEF` (32), `MSG_POS_LLH` (34), `MSG_BASELINE_ECEF` (20), `MSG_BASELINE_LLH` (22)`, `MSG_VEL_ECEF` (20), `MSG_VEL_LLH` (22) is 161 bytes.

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -472,3 +471,114 @@ definitions:
...
+            desc: Latitude
+        - lon:
+            type: double
+            units: deg
+            desc: Longitude
+        - height:
+            type: s32
+            units: mm
+            desc: Height above WGS84 ellipsoid
+        - heightMSL:
```
ljbade: We won't be able to fill out this field until we have a geoid model in Piksi

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -472,3 +471,114 @@ definitions:
...
+              implemented). Defaults to 0.
+        - v_accuracy:
+            type: u16
+            units: mm
+            desc: |
+              Vertical position accuracy estimate (not
+              implemented). Defaults to 0.
+        - n_sats:
+            type: u8
+            desc: Number of satellites used in solution.
```
ljbade: Number of satellites or number of signals?

@anth-cole ?

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -472,3 +471,114 @@ definitions:
...
+            desc: Velocity East coordinate
+        - vel_d:
+            type: s32
+            units: mm/s
+            desc: Velocity Down coordinate
+        - pdop:
+            type: u16
+            units: 0.01
+            desc: Position Dilution of Precision
+        - flags:
```
ljbade: I think we need to be able to specify different source for position and velocity?

e.g. currently we can have RTK position, but we only support SPP velocity

ljbade: we might also want flags indicating UTC status, e.g have we received leap second info yet (if not then we would have to send GPS time, or an assumed number of leap seconds)

spec/yaml/swiftnav/sbp/navigation.yaml
anth-cole: This is more of a product question to me, what are we trying to communicate with this field? If it's geometric strength, then number of sats makes sense, if not then number of signals might make more sense. Neither gives us really a lot of information without also knowing the PRN's etc. I'd probably lean more towards number of satellites as this is easier to explain to consumers what it means

spec/yaml/swiftnav/sbp/navigation.yaml
anth-cole: I disagree on both points
1. We should output a velocity solution with appropriate accuracy, the consumer shouldn't care if it's SPP or RTK
2. A flag indicating we have receiver UTC leap second info would be very infrequently used. Leap seconds change rarely, and we can also have future leap second, so the flag should only be useful if we haven't turned the receiver on for around ~18months, which is probably a rare case

ljbade: @anth-cole so you think we should also add a velocity accuracy estimate? Or if we can't guarantee that we meet the product spec (e.g. bad sat geometry etc), should we just not output a velocity at all?

ljbade: Reason being how can a user integrate out velocity output into a control loop without knowing the quality of the output?

anth-cole: I think a velocity (or Position) estimate without an accompanying confidence estimate is largely useless for a precise positioning product. How we output that confidence estimate is more an implementation issue

mookerji: re: velocity, we use the derivative of carrier phase measurements and not the derivative of positions to calculate velocity, so it doesn't actually make sense to flag it as RTK or SPP (I think @fnoble echoed this a few weeks). If that's the case, I question if it even makes sense to give a velocity variance, as it's fundamentally derivable from the carrier phase measurement noise and the sampling rate.

spec/yaml/swiftnav/sbp/navigation.yaml
mookerji: Correct me if I'm wrong (or if the documentation is wrong), but are we currently outputting the number of satellites or the number of signals in MSG_POS_ECEF? It's the latter, right?

/cc @benjamin0

spec/yaml/swiftnav/sbp/navigation.yaml
mookerji: Good point. I'll leave it as undocumented.

spec/yaml/swiftnav/sbp/navigation.yaml
ljbade: @mookerji yes that is currently how we do it, but what if we want to change the method? e.g. we add a SPP Kalman filter, or perhaps use the dynamic filter state from RTK. Of course we could just redefine this message and/or datasheet if we do implement such a change.

spec/yaml/swiftnav/sbp/navigation.yaml
denniszollo: It says number of satellites, but it is actually number of signals currently.

spec/yaml/swiftnav/sbp/navigation.yaml
denniszollo: I dedicated 3 bits to UTC validity in the UTC_TIME message proposal.

image

denniszollo:
we are going to have a very tough time refactoring the solution thread to put this out. the flow control will be nuts.

spec/yaml/swiftnav/sbp/navigation.yaml
ljbade: @denniszollo ah excellent a user can use that message if they need more detail then what is available in PVT

mookerji:
@denniszollo your concerns are duly noted, I moved that to a different issue here: https://github.com/swift-nav/skylark-tales/issues/432

spec/yaml/swiftnav/sbp/navigation.yaml
mookerji: Other receivers also output the number of signals? If so, we should try to be consistent with that from the customer expectations point-of-view.

spec/yaml/swiftnav/sbp/navigation.yaml
mookerji: @denniszollo just saw this. I actually don't think that we should include hardware/platform-specific things like "factory default", NVM, etc. We should keep the contents of the message as abstracted as possible. I read a bit about how the UTC-GPS this morning and decided to go with an invalid/warning/valid approach, basically. Whatever we go with, we should likely settle on something shared.

mookerji:
Updated! Looking for feedback on: 

- UTC time flags
- Missing data we might want to include
- Do we want to include any additional reserved fields?
- Do we want to replace lat/lon with a fixed-width representation? If so I can work out the details to make sure we have enough precision.

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -472,3 +471,123 @@ definitions:
...
+                    - 0x010: RTK Fixed
+                    - 0x011: SPP
+                    - 0x100: Reserved
+                    - 0x101: Reserved
+                    - 0x110: Reserved
+                    - 0x111: Reserved
+              - 3-5:
+                  desc: UTC flags
+                  values:
+                    - 0x00: Invalid UTC date
```
ljbade: Could be perhaps say, "invalid UTC date, GPS time returned instead" or something?

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -472,3 +471,123 @@ definitions:
...
+                    - 0x01: Valid UTC date, defaults used
+                    - 0x10: Valid UTC leap seconds
```
ljbade: maybe say leap seconds, and polynomial adjustment (gets UtC down to a few nanoseconds vs potentially tens of ns off)

ljbade:
Hmm, are we providing velocity accuracy info in the dedicated velocity messages? If customer needs that level of detail for a control loop etc

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -472,3 +471,114 @@ definitions:
...
+            type: u8
+            units: s
+            desc: |
+              Seconds of the minute (UTC) (ranges from 0 to 60). The upper range
+              includes an the possibility of an extra leap second correction.
+        - ns:
+            type: s32
+            units: ns
+            desc: Nanosecond residual of UTC sec (ranges from -1e9 to 1e9)
+        - lat:
```
mookerji: Perhaps! I think this was the intended to be a really basic, NMEA-style atomic message (similar to UBX-NAV-PVT in the M8 protocol).

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -472,3 +471,114 @@ definitions:
                     values:
                         - 0: Float RTK
                         - 1: Fixed RTK
+
+ - MSG_PVT:
+    id: 0x0208
+    short_desc: Position velocity time (UTC) solution
```
mookerji: See response below. I think the intension here was to make this very easy to consume, and having UTC is consistent with that.

mookerji:
@ljbade see my note about velocity accuracy. I don't think it makes sense.

ljbade:
@mookerji yes but that will not give us flexibility in future to change it to use kalman filer or rtk dynamic model derived velocity, if we find our current pure LSQ doppler implementation is not good enough for our customer use cases

but im happy if you think if we do such a change we will just define a new SBP message with accuracy field at the time

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -472,3 +471,123 @@ definitions:
...
+                  desc: UTC flags
+                  values:
+                    - 0x00: Invalid UTC date
+                    - 0x01: Valid UTC date, defaults used
+                    - 0x10: Valid UTC leap seconds
+              - 7:
+                  desc: RAIM repair flag
+                  values:
+                    - 0x0: No RAIM repair
+                    - 0x1: RAIM repair
```
switanis: Comments:
- add gps week
- make lat/lon as scaled integer
- use altitude instead of height
- n_sats rename to n_sats_used
- replace vel_x with commonly used SOG and COG
- fix mode flag: value 3 for SPP is rather unexpected, better use:
    0 - no fix
    1 - SPS
    2 - DGPS (reserved)
    3 - RTK Float
    4 - RTK Fixed
- UTC flags should be the same as ones used in UTC message Dennis is preparing
- remove RAIM repair, I see no value for user
- do we ever expect 2D fixes with invalid altitude? May need more individual field validity flags.
- add HDOP
- add data latency

spec/yaml/swiftnav/sbp/navigation.yaml
valeri-atamaniouk: What about adding a leap seconds counter as well? Otherwise it would not be possible to compute time interval with 1s precision when measurements are from different sides of leap second event.

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -472,3 +471,123 @@ definitions:
...
+            units: mm/s
+            desc: Velocity East coordinate
+        - vel_d:
+            type: s32
+            units: mm/s
+            desc: Velocity Down coordinate
+        - pdop:
+            type: u16
+            units: 0.01
+            desc: Position Dilution of Precision
```
nsirola: Is the PDOP value of use if we already output h_accuracy and v_accuracy in millimeters (once implemented) and presumably also vel_accuracy as requested below?

spec/yaml/swiftnav/sbp/navigation.yaml
denniszollo: It should be number of space vehicles for this message.  number of signals is interesting, but if you want that go to the tracker and observation outputs.

PR #400, opened 2016-11-02
Give Haskell code better casing for GPSTime

mfine:
This is pretty ugly fix, but `gPSTime_tow` was pretty bad.


ljbade:
wow that was pretty bad

JoshuaGross:
hilarious

PR #412, opened 2016-11-28
Add a new request/response pair for execution of shell commands.

gsmcmullin:
A new request message MSG_SHELL_COMMAND_REQ is defined to request execution of a shell command on piksi.  The shell output may be sent back to the host using the existing MSG_LOG messages, and the final exit code is returned in the new MSG_SHELL_COMMAND_RESP message.


JoshuaGross:
I'd like to see justification on this, as I have concerns around

- security (we should be making it harder to gain control of devices, not easier)
- what happens if it's a long-running process?
- is there a whitelist of commands that can be run?

gsmcmullin:
I've updated it to remove the word 'shell'.

The rest of the concerns are around implementation, not protocol.

- Since it is possible to load new firmware on the Piksi, ultimately you can make to run whatever you want.  Not allowing arbitrary commands only adds inconvenience, not security.
- If the process does not terminate, the response will not be sent.
- A command whitelist will be added to the implementation.

fnoble:
lgtm

(Only reviewed the YAML changes, didn't check or try building the bindings)

JoshuaGross:
If codegen has been run and committed for all changes, this looks good - do you want this released as 2.1? I'm happy to do that once this is merged.

gsmcmullin:
Codegen has been run and changes committed in a separate commit.  I'll need a new release to updated the pointers in the piksi_tools and piksi_buildroot PRs. Thank you.

PR #418, opened 2016-12-04
Raw IMU data message

fnoble:
Adding message for raw IMU data.


spec/yaml/swiftnav/sbp/imu.yaml
```diff
@@ -0,0 +1,55 @@
...
+definitions:
+
+ - MSG_IMU_RAW:
+    id: 0x0900
+    short_desc: Raw IMU data
+    desc: |
+      Raw data from the Inertial Measurement Unit, containing accelerometer and
+      gyroscope readings.
+    fields:
+        - tow:
```
mookerji: Perhaps we want this as a flag?

spec/yaml/swiftnav/sbp/imu.yaml
```diff
@@ -0,0 +1,55 @@
...
+            type: s16
+            desc: Acceleration in the body frame X axis
+        - acc_y:
+            type: s16
+            desc: Acceleration in the body frame Y axis
+        - acc_z:
+            type: s16
+            desc: Acceleration in the body frame Z axis
+        - gyr_x:
+            type: s16
```
mookerji: Units! m/s and m/s/s (or g's)?

spec/yaml/swiftnav/sbp/imu.yaml
mookerji: Also, this is technically time-tagged or something, right?

mookerji: Other things that come to mind for a flag: whether or not the IMU measurement is calibrated.

fnoble:
@mookerji : some good points that I had also been thinking about. Here was my rationale for doing it this way:

 - The message should be as compact as possible because it may be sent at 400+ Hz
 - There is a question how "raw" should it be, I opted to make it just the raw s16 value read from the sensor without any calibration applied but I can see the argument for doing some minimal processing to make it more consumable at the expense of giving people down the line the freedom to do that themselves
 - The IMU has several different selectable full-scale ranges (+/- 2, 4, 8 or 16g), this means that it's not possible to give consistent units as they depend on what settings are selected
 - I originally had a flags field containing the selected range, the validity etc. but I thought it wasn't worth just sending back the IMU settings at 400 Hz and the time validity can be incorporated into the `tow` field as the full range isn't used by valid values

Another question is if 1/256 ms ~ 3.9 us time resolution is sufficient?

s-landers:
I think it makes sense to leave the contents of this message in this raw state, since that keeps them compact  - calibrated values would need to be moved to a floating or fixed point value, both of which would need to be expanded from the original full-scale range based format.  This also allows other parts of the system to perform their own calibration.

A 3.9 us time resolution sounds fine to me.

I agree that it would be overkill to send IMU settings out as part of this message.  Would settings like the full-scale range then become a system setting that can be read or written like other system settings are?

One other thing to consider is having a way to access IMU temperature.  I don't think it's appropriate to make part of this message, since 400 Hz of temperature data seems unnecessary - maybe it would make sense to have an auxiliary message that is sent at a lower rate that could include temperature and any other values that don't belong in a high rate message.

fnoble:
@s-landers : Sounds good, what rate do you think is necessary for the temperature data? Also, what do you think about the magnetometer readings? I was thinking of putting them in a separate message at a lower rate too.

spec/yaml/swiftnav/sbp/imu.yaml
```diff
@@ -0,0 +1,94 @@
...
+            desc: IMU configuration
+            fields:
+              - 4-7:
+                  desc: Gyroscope Range
+                  values:
+                      - 0: +/- 2000 deg / s
+                      - 1: +/- 1000 deg / s
+                      - 2: +/- 500 deg / s
+                      - 3: +/- 250 deg / s
+                      - 4: +/- 125 deg / s
```
denniszollo: Perhaps rather than, or in addition to range here, we should put LSB's value in deg/s or g's

fnoble:
Notes from meeting earlier today:
 - Go with raw value straight from sensor
 - Magnetometer not needed for now
 - Temperature would be useful, only needed at 1Hz
 - Idea to define "auxiliary" message containing temp as well as IMU full scale range configuration such that it can be unambiguously determined from the logs
 - Aux message could have "IMU type" field so it can be extended to support different kinds of IMU in the future

Given the feedback I've gotten from @s-landers and @denniszollo I'm going to merge these messages in their current form shortly.

mookerji:
👍

Issue #416, opened 2016-12-01
Compile failure / inaccurate version bounds with sbp-2.1.1


hvr:
In my function as a Hackage Trustee (see https://github.com/haskell-infra/hackage-trustees)  I have to inform you, that the latest release you uploaded to Hackage is still broken (see below).

Please note that this also means that Hackage can't build the documentation and that users can't install your package, and therefore results in a bad user experience.

If you have any questions or need any assistance, please let me know.

```
Configuring component exe:sbp2yaml from sbp-2.1.1
Warning: The package has a version range for a dependency on an internal library: sbp -any && ==2.1.1, sbp -any && ==2.1.1, sbp -any && ==2.1.1, sbp -any && ==2.1.1, sbp -any && ==2.1.1. This version range has no semantic meaning and can be removed.
Preprocessing executable 'sbp2yaml' for sbp-2.1.1...
[1 of 1] Compiling Main             ( main/SBP2YAML.hs, /tmp/matrix-worker/1480622119/dist-newstyle/build/x86_64-linux/ghc-8.0.1/sbp-2.1.1/c/sbp2yaml/build/sbp2yaml/sbp2yaml-tmp/Main.o )
Linking /tmp/matrix-worker/1480622119/dist-newstyle/build/x86_64-linux/ghc-8.0.1/sbp-2.1.1/c/sbp2yaml/build/sbp2yaml/sbp2yaml ...

Configuring component lib from sbp-2.1.1
Preprocessing library sbp-2.1.1...
[ 1 of 19] Compiling SwiftNav.SBP.Encoding ( src/SwiftNav/SBP/Encoding.hs, /tmp/matrix-worker/1480622119/dist-newstyle/build/x86_64-linux/ghc-8.0.1/sbp-2.1.1/build/SwiftNav/SBP/Encoding.o )
[ 2 of 19] Compiling SwiftNav.CRC16   ( src/SwiftNav/CRC16.hs, /tmp/matrix-worker/1480622119/dist-newstyle/build/x86_64-linux/ghc-8.0.1/sbp-2.1.1/build/SwiftNav/CRC16.o )
[ 3 of 19] Compiling SwiftNav.SBP.Types ( src/SwiftNav/SBP/Types.hs, /tmp/matrix-worker/1480622119/dist-newstyle/build/x86_64-linux/ghc-8.0.1/sbp-2.1.1/build/SwiftNav/SBP/Types.o )
[...]

Issue #413, opened 2016-11-30
Compile failure / inaccurate version bounds with latest release


hvr:
Both the Hackage doc-builder and the Hackage matrix builder flagged the 2.10 release due to compile failures, see

 - https://hackage.haskell.org/package/sbp-2.1.0/reports/

Unfortunately, the Hackage matrix builder found that all releases going back to 0.51 (which added the executables `sbp2json` and `json2sbp`) result in a compile failure, because their install-plans have degraded due to lack of version bounds.

Specifically for the recent release, the Hackage matrix builder complained for GHC 8.0:

```
Configuring component lib from sbp-2.1.0
Preprocessing library sbp-2.1.0...
[ 1 of 19] Compiling SwiftNav.SBP.Encoding ( src/SwiftNav/SBP/Encoding.hs, /tmp/matrix-worker/1480485529/dist-newstyle/build/x86_64-linux/ghc-8.0.1/sbp-2.1.0/build/SwiftNav/SBP/Encoding.o )
[ 2 of 19] Compiling SwiftNav.CRC16   ( src/SwiftNav/CRC16.hs, /tmp/matrix-worker/1480485529/dist-newstyle/build/x86_64-linux/ghc-8.0.1/sbp-2.1.0/build/SwiftNav/CRC16.o )
[ 3 of 19] Compiling SwiftNav.SBP.Types ( src/SwiftNav/SBP/Types.hs, /tmp/matrix-worker/1480485529/dist-newstyle/build/x86_64-linux/ghc-8.0.1/sbp-2.1.0/build/SwiftNav/SBP/Types.o )
[ 4 of 19] Compiling SwiftNav.SBP.TH  ( src/SwiftNav/SBP/TH.hs, /tmp/matrix-worker/1480485529/dist-newstyle/build/x86_64-linux/ghc-8.0.1/sbp-2.1.0/build/SwiftNav/SBP/TH.o )
[ 5 of 19] Compiling SwiftNav.SBP.Bootload ( src/SwiftNav/SBP/Bootload.hs, /tmp/matrix-worker/1480485529/dist-newstyle/build/x86_64-linux/ghc-8.0.1/sbp-2.1.0/build/SwiftNav/SBP/Bootload.o )

src/SwiftNav/SBP/Bootload.hs:52:93: error:
    Ambiguous occurrence ‘stripPrefix’
    It could refer to either ‘BasicPrelude.stripPrefix’,
[...]

Issue #250, opened 2015-09-09
Fix Haskell client haddock generation


mookerji:
/cc @mfine

PR #422, opened 2016-12-12
Tweak stack.yaml for Hackage compatibility

alanz:
This will transparently add upper and lower bounds to all package dependencies not currently having them, based on the snapshot in use, when the package is uploaded via either `stack upload` or `stack sdist`.

See https://www.fpcomplete.com/blog/2015/09/stack-pvp

### DIFF
--- a/haskell/stack.yaml
+++ b/haskell/stack.yaml
@@ -3,3 +3,4 @@ flags: {}
 packages:
   - '.'
 extra-deps: []
+pvp-bounds: both


tonymorris:
Hi. Any chance of getting this patch in soon?

tfausak:
This would be a great addition. It would fix a lot of other issues, like #250, #413, and #416. 

I think the Haskell maintainers are @JoshuaGross, @mfine, and @mookerji. Can y'all take a look at this?

mookerji:
Sorry this didn't get a reply!

It's on our roadmap to resolve the package versioning issue, and we'll resolve this PR when we have more slack on our internal timeline. We have a several important/internal/downstream projects supporting Swift's core product development that will break if we attempt to resolve all package versioning requirements right now (tl;dr: binary dependencies). Fixing these issues will require a time investment on our part before we can accept the PR.

In the meantime, using stack as is mentioned in the README (https://github.com/swift-nav/libsbp/tree/master/haskell) should work.

mfine:
We support more than one LTS version - LTS-5 and LTS-6.  Naively setting upper and lower `pvp-bounds` would have broken our downstreams. See: #498.

If there are actual issues with this library that anyone's running into with their piksi or sbp data, please get in touch with support.

PR #423, opened 2016-12-12
Remove old deprecateds

mfine:
Remove deprecated messages that have been removed for awhile. In cleaning things up for 2.0, let's let these old message go.

/cc @denniszollo @ljbade @mookerji


mfine:
Also removing:

+ MSG_CW_RESULTS
+ MSG_CW_START
+ MSG_INIT_BASE
+ MSG_IAR_STATE

mfine:
We should throw out the whole `flash.yaml` module.

mfine:
We should throw out the `MSG_BOOTLOADER_` messages.

mookerji:
👍

mookerji:
@mfine actually, I think I'm going to defer to @denniszollo on this one, in part because we have devices (v2's) in the field that may need some of these messages in the developer console.

mfine:
We shouldn't support running the console off master with v2 piksis. That's a terrible idea.

denniszollo:
I don't mean the console so much as just log processing.  I think we should jut be very explicit that our pattern is changing.  Prior to this point all of our messaging was backwards compatibale since the start of libsbp.  Now we are saying that old logfiles and v2 devices are not compatible with master libsbp. 

In order to process those logs, you need to explicitly use  an old libsbp release.  This means we will need to keep a compatibility table of device and firmware versions against lisbsbp and console versions in our documentation.

mfine:
> In order to process those logs, you need to explicitly use an old libsbp release. This means we will need to keep a compatibility table of device and firmware versions against lisbsbp and console versions in our documentation.

That sounds like a great idea and was the original intention of our versioning. The original intention was not to carry forward every message we ever came up with. Further, supporting multi-versioning will be an increasingly bleak prospect - we should keep it scoped as originally intended (to major releases).

mookerji:
@denniszollo if it's for the case of log processing, I worry that keeping around old message definitions introduces a versioning/interface question and support burden that we're not really well-placed to deal with (or rather, keeping message definitions around in their current form isn't really the right answer to this problem).  Also, on better look at the PR, the vast majority of the messages given here aren't really applicable to any manner of v2 post-processing analysis, and it's unclear to me if the depreciated `MSG_OBS`, etc. could be particularly useful?

PR #425, opened 2016-12-22
add CDC driver

denniszollo:
This adds a CDC driver primarily to support CDC AMC devices on OSX versions past Yosemite.

The "base driver" didn't handle disconnection out of the box
the "pyserial driver" didn't work on any OSX versions past Yosemite. 

I determine that the device is disconnected based upon when reads from the device return nothing.  I'm not sure if this is the best method, but it's the only thing I could get to work without a more sophisticated driver.  Referrd to [posix read](http://pubs.opengroup.org/onlinepubs/009695399/functions/read.html), the [fread in stdlib]([url](http://www.cplusplus.com/reference/cstdio/fread/)) and [python file.read()](https://docs.python.org/2/library/stdtypes.html#bltin-file-objects) to decide on the empty string comparison.  IOError is tested to be thrown when the port is disconnected or the CDC device goes offline.

/cc @fnoble @mfine

### DIFF
--- a/python/sbp/client/drivers/cdc_driver.py
+++ b/python/sbp/client/drivers/cdc_driver.py
@@ -1,0 +1,73 @@
+# Copyright (C) 2016 Swift Navigation Inc.
+# Contact: Dennis Zollo <dzollo@swiftnav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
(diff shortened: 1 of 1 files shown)


denniszollo:
@JoshuaGross @mfine  

Can I get a code review of this?  This allows us to use the usb port on Piksi without relying on broken PySerial.

JoshuaGross:
Works for me so I'm okay merging it, but I'm concerned about the lack of closing the handle?

denniszollo:
I put the close back in and just suppressed my warning.  This is tested to work.

PR #430, opened 2017-01-13
make handler callable threadsafe, Catch pyserial writeTimeoutErrors

denniszollo:
When using skylark on the console, lately we have seen this problem when you try and write something else to the serial port concurrently. 

Here we put a mutex around writing to the framer.  This fixes our interaction between skylark and other stuff in the console.  After a review and any comments we can try and cut an sbp release.  Potentially we could do instead this with some kind of queue rather than blocking writes to avoid holding up other threads, but I think this approach is acceptable.

/cc @jkretzmer @mfine @switanis

### DIFF
--- a/python/sbp/client/handler.py
+++ b/python/sbp/client/handler.py
@@ -38,6 +38,7 @@ class Handler(object):
     self._receive_thread.daemon = True
     self._sinks = [] # This is a list of weakrefs to upstream iterators
     self._dead = False
+    self._write_lock = threading.Lock()
 
   def _recv_thread(self):
     """
@@ -241,7 +242,8 @@ class Handler(object):
     self.remove_callback(cb, msg_type)
 
   def __call__(self, msg, **metadata):
-    self._source(msg, **metadata)
+    with self._write_lock:
+      self._source(msg, **metadata)
 
   class _SBPQueueIterator(object):
     """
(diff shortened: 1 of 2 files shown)


cbeighley:
@denniszollo I would use the `threading.Lock` object for this: http://stackoverflow.com/questions/10525185/python-threading-how-do-i-lock-a-thread.

mfine:
This ends up just wrapping the underlying driver's write with a lock. It's not clear that that's not threadsafe, but guess this would be safe to add.

python/sbp/client/handler.py
```diff
@@ -17,6 +17,7 @@
 import threading
 import weakref
 from Queue import Queue
+import time
```
mfine: you're not using this import.

python/sbp/client/handler.py
```diff
@@ -241,7 +243,10 @@ def cb(msg, **metadata):
     self.remove_callback(cb, msg_type)
 
   def __call__(self, msg, **metadata):
+    self._framer_write_locked.acquire(True)
```
mfine: http://code.activestate.com/lists/python-list/582952/

```python
// use a more succinct name for the lock - and use context to grab it and release it:
with self.write_lock:
  self._source(msg, **metadata)
```

python/sbp/client/handler.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2015 Swift Navigation Inc.
+# iopyright (C) 2015 Swift Navigation Inc.
```
mfine: /i/C/

mfine:
Outside of the copyright nit, looks good.

denniszollo:
I rebased down to the two substantial commits and pushed back up - I think it's ready.

PR #434, opened 2017-01-25
Add option to initialize with file handle.

benjamin0:
I think this should preserve backwards compatibility but also allow us to pass in a file handle.

For example,
```python
JSONLogIterator(handle=sys.stdin)
```

@mfine @akleeman

### DIFF
--- a/python/sbp/client/loggers/base_logger.py
+++ b/python/sbp/client/loggers/base_logger.py
@@ -23,9 +23,13 @@ class BaseLogger(object):
   filename : string
     File to log to.
   """
-  def __init__(self, filename, mode="w", tags={}, dispatcher=None):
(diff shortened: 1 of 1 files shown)


python/sbp/client/loggers/base_logger.py
```diff
@@ -23,9 +23,13 @@ class BaseLogger(object):
   filename : string
     File to log to.
```
akleeman: doc update

mfine:
Woof. That new interface make things even more complicated. Let's just move to a handle and do the right thing here: https://github.com/swift-nav/libsbp/pull/435. Let's not support the backwards interface - it's doing the wrong thing with file manage regardless - e.g., never closing them.

PR #435, opened 2017-01-26
Remove filename - use filehandles

mfine:
Use filehandles instead of filenames.

The passing in a of a filename is a bad pattern - it bakes in the management of the file into the object, doesn't get cleaned up, etc.

And let's not preserve backwards compatibility - let's just fix things to do the right thing.

/cc @benjamin0


benjamin0:
👍

PR #436, opened 2017-01-30
Add MSG_NDB_EVENT

pmiettinen:

PR #438, opened 2017-02-01
Rename member variable to avoid name collision

pmiettinen:
'sender' variable name is already taken by default in SBP message, rename.


pmiettinen:
Reviewed here https://github.com/swift-nav/libsbp/pull/436

PR #440, opened 2017-02-08
Pdf publish version fix

denniszollo:
This uses the makefile's idea of the version for the pdf document

It also adds a distribution target to downloads.swiftnav.com for which you will need s3cmd and aws credentials

/cc @mfine @JoshuaGross

### DIFF
--- a/generator/sbpg/targets/resources/sbp_base.tex
+++ b/generator/sbpg/targets/resources/sbp_base.tex
@@ -64,7 +64,7 @@
 \title{Swift Navigation Binary Protocol}
 \version{\input{\jobname.info}}
 \author{Swift Navigation}
-\mysubtitle{Protocol Specification \theversion}
+\mysubtitle{Protocol Specification (((version)))}
 \date{\today}
 
 \begin{document}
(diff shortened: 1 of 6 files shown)


generator/sbpg/targets/resources/sbp_base.tex
```diff
@@ -32,6 +32,7 @@
 \numberwithin{table}{subsection}
 \numberwithin{field}{subsection}
 
+\renewcommand{\version}{(((version)))}
```
mfine: I'd just remove this and use the macro input directly everywhere - It's confusing now because you're using the macro in some spots and the function in others.

denniszollo: I can't inside the the .sty file, so in order to do this I'd have to move the footnote stuff out of the .sty file

denniszollo: or leave the macro hardcoded to a string or something here

mfine: :+1:

JoshuaGross:
awesome!

PR #441, opened 2017-02-15
Add MsgAlmanacGps and MsgAlmanacGlo

pmiettinen:


spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -1110,3 +1110,122 @@ definitions:
...
+        - valid:
+            type: u8
+            desc: Status of ephemeris, 1 = valid, 0 = invalid
+        - health_bits:
+            type: u8
+            desc: |
+              Satellite health status.
+              GPS: ICD-GPS-200, chapter 20.3.3.3.1.4
+              SBAS: 0 = valid, non-zero = invalid
+              GLO: 0 = valid, non-zero = invalid
```
JoshuaGross: It's not clear to me, from this description, which bit of `health_bits` maps to each constellation; and could you describe (at least abbreviated) whatever is in "ICD-GPS-200, chapter 20.3.3.3.1.4" instead of just referencing it? Additionally, the `valid` bit above says that 1 is valid and 0 is invalid, which is what I would expect from a binary protocol. Can we be consistent here and say that 1 is valid and 0 is invalid in the health bits as well?

spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -1110,3 +1110,122 @@ definitions:
...
+          units: rad
+          desc: Inclination
+      - af0:
+          type: double
+          units: s
+          desc: Polynomial clock correction coefficient (clock bias)
+      - af1:
+          type: double
+          units: s/s
+          desc: Polynomial clock correction coefficient (clock drift)
```
JoshuaGross: Are these units really seconds and `s/s`? I would think you'd want a smaller unit here so we don't lose precision in transmission? I'm less certain about this than other comments I've made, but would be very interested in an explanation.

pmiettinen: IS-GPS-200H Table 20-VI. Units are correct and scale factors are 2^(-20) and 2^(-38).

pmiettinen: Ah, I'll add the scale factors, similar to MSG_GROUP_DELAY for example.

pmiettinen: I guess we are doing similarly to ephemeris, that scaling is done before sending MsgAlmanac on Piksi so the scaling units are not added here.

spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -1110,3 +1110,133 @@ definitions:
...
+            units: s
+            desc: Curve fit interval
+        - valid:
+            type: u8
+            desc: Status of almanac, 1 = valid, 0 = invalid
+        - health_bits:
+            type: u8
+            desc: |
+              Satellite health status for GPS:
+                - bits 5-7: NAV data health status. See IS-GPS-200H
```
JoshuaGross: Thanks, this section is much more useful now!

JoshuaGross:
looks good!

PR #443, opened 2017-03-01
Increase timeouts for connections and reads

mfine:
Are timeouts are pretty tight. Especially now that connections will be held waiting for a position. This could improve connection stability on the client.

/cc @denniszollo @mookerji


python/sbp/client/drivers/network_drivers.py
```diff
@@ -88,8 +88,8 @@ def write(self, s):
 class HTTPException(Exception):
   pass
 
-DEFAULT_CONNECT_TIMEOUT = 10
-DEFAULT_READ_TIMEOUT = 10
+DEFAULT_CONNECT_TIMEOUT = 30
+DEFAULT_READ_TIMEOUT = 120
 DEFAULT_TIMEOUT = (DEFAULT_CONNECT_TIMEOUT, DEFAULT_READ_TIMEOUT)
 MAX_CONNECT_RETRIES = 10
```
denniszollo: perhaps we should reduce the number of retries given the timeout change?  I think in this case we are talking 10s of minutes that requests will keep spinning.

mfine: Yeah, that sounds good - 3 or 5?

There will be some downsides to this increased timeout during connection outages, but it will start acting more like a long poll and not be so frenetic on first connections.

mookerji:
👍 Thanks! I was thinking about making this change during the hackathon.

PR #444, opened 2017-03-01
Split sbp_process() into two functions. 

benjamin0:
Add `sbp_process_payload` which allows directly processing a payload.

If this looks good, I'll add a docstring.

@mookerji @denniszollo @mfine cc: @anth-cole

### DIFF
--- a/c/src/sbp.c
+++ b/c/src/sbp.c
@@ -389,6 +389,19 @@ s8 sbp_process(sbp_state_t *s, u32 (*read)(u8 *buff, u32 n, void *context))
   return SBP_OK;
 }
 
+/** Directly process a SBP message.
+ * If a SBP message has already been decoded (for example, from a binary
+ * stream or from a JSON log file) use this function to directly process it.
+ *
+ * \param s State structure
+ * \param sender_id SBP message sender id
+ * \param msg_type SBP message type
+ * \param msg_len SBP message length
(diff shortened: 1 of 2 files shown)


c/src/sbp.c
```diff
@@ -372,15 +372,8 @@ s8 sbp_process(sbp_state_t *s, u32 (*read)(u8 *buff, u32 n, void *context))
...
-        s8 ret = SBP_OK_CALLBACK_UNDEFINED;
-        sbp_msg_callbacks_node_t* node;
-        for (node = s->sbp_msg_callbacks_head; node; node = node->next) {
-          if (node->msg_type == s->msg_type) {
-            (*node->cb)(s->sender_id, s->msg_len, s->msg_buff, node->context);
-            ret = SBP_OK_CALLBACK_EXECUTED;
-          }
-        }
-        return ret;
+        return sbp_process_payload(s, s->sender_id, s->msg_type, s->msg_len,
```
mfine: just a nit - I'd assign the return value of `sbp_process_payload` to `ret` and continue to return `ret`. Easier for the debugger to inspect the value returned, etc.

benjamin0: good point, fixed

benjamin0:
I added a docstring and think this is good to go now.

PR #449, opened 2017-03-14
Jdiamond/network status

axlan:
Added message to query and return state of network adapters. Data is made to reflect output of ifaddrs struct returned by getifaddrs() in c.

The purpose of this message is to be able to determine network state through the piksi console, mainly to allow connecting over TCP after a DHCP configuration.

This is dependency for https://github.com/swift-nav/piksi_buildroot/pull/151, and once both these are in the console will be updated to display this information

@denniszollo

### DIFF
--- a/c/include/libsbp/piksi.h
+++ b/c/include/libsbp/piksi.h
@@ -34,7 +34,7 @@
  * This is a legacy message for sending and loading a satellite
  * alamanac onto the Piksi's flash memory from the host.
  */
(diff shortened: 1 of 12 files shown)


mookerji:
@axlan also, for future reference, we try to keep commits of message definition and package updates on separate PRs (this might be in the CONTRIBUTING file, but I may be wrong). That being said, keeping this as-is is fine.

PR #450, opened 2017-03-15
python: pyserial driver.  Add hdw flow control to API

denniszollo:
we need hdw flow control options for the serial driver.


denniszollo:
@mookerji @JoshuaGross 
Any qualms?
Otherwise it's going in!

PR #452, opened 2017-03-22
Utc msg fixes

denniszollo:
Stefan and I decided that the UTC message wasn't as user friendly as it could be and decided to quickly redefine it so that it make sense.  

Now you don't need to use the GPS TOW in order to get UTC time precise beyond seconds.  The GPS TOW can be considered the way to match up each solution epoch.  The formula for getting the seconds in the minute  is msg.seconds + msg.nanoseconds * 10e-9 which is simple and straightforward.

/cc @switanis @mfine @mookerji @JoshuaGross 

Since the message is still not produced by the hardware and hadn't been used, combined with the fact that field widths are identical, I have kept the same message ID the same as before.


mfine:
Are we going to be consistent across all messages with nanoseconds handling? In other areas, it continues to be a range that includes negative numbers.

```
sbp/ext_events.yaml
41:              from -500000 to 500000)

sbp/gnss.yaml
104:              from -500000 to 500000)

sbp/navigation.yaml
61:              from -500000 to 500000)
598:              from -500000 to 500000)
```

denniszollo:
Based upon the discussions and feedback, I propose that we change the other ns fields to be called ns_residual as to differentiate them from this ns field.

I also think that having a "unix epoch" message doesn't make sense, because if you look at the unix epoch, it is defined from the UTC time message, rather than the other way around and there is no concept of astronomical unix epoch.

denniszollo:
Unfortunately, renaming this field will break a lot of other code when libsbp is upgraded

mfine:
> I also think that having a "unix epoch" message doesn't make sense, because if you look at the unix epoch, it is defined from the UTC time message, rather than the other way around and there is no concept of astronomical unix epoch.

I don't follow this - what's "rather than the other way around" mean? Time is generated from an epoch - the time segments that are being populated in the current message are derivable from an epoch. I don't understand what's missing here.

denniszollo:
After our synch up meeting, the following decision were made:

- This PR can go in unchanged -
- we will additionally define a new message that represents GPS - UTC time delta for the next release
- We will plan for release of piksi_buildroot to set a default message mask on UART 1 for release that decimates some messages that are unneeded at a high data rate to alleviate serial port congestion. Messages that will be cut out are:
   - MSG_BASELINE_ECEF
   - MSG_POS_ECEF
   - MSG_VEL_ECEF
because these are not used by console at present.
Messages to be decimated to 1hz
   - MSG_UTC_TIME
   - MSG_DGNSS_STATUS
   - MSG_DOPS

PR #454, opened 2017-03-29
inertial_updates: new messages & navigation changes

denniszollo:
/cc @fnoble @jkretzmer @switanis 

These are the proposed messages to support inertial outputs.  They have been rebooted after our recent work.

See doc here:
[sbp.pdf](https://github.com/denniszollo/libsbp/blob/307011b18a9b556f66fa6601a8ef68b23ea85cb3/docs/sbp.pdf)


denniszollo:
Additional todo: add flag for in what reference point the messages is referenced

PR #458, opened 2017-04-10
Add SBP_MSG_FCNS_GLO

dt-exafore:
Based on https://github.com/swift-nav/exafore_planning/issues/685

### DIFF
--- a/c/include/libsbp/observation.h
+++ b/c/include/libsbp/observation.h
@@ -710,6 +710,28 @@ coordinate system
 } msg_almanac_glo_t;
 
 
+/** GLONASS SV orbital and frequency slots mapping information
+ *
+ * The message reports mapping information regarding GLONASS SV orbital and
+ * frequency slots.
+ * Mapped as follow:
+ * index (SV orbital slot)  fcns[index]
+ * 0                        0xFF
+ * 1                        FCN for SV orbital slot 1
+ * ...                      ...
+ * 28                       FCN for SV orbital slot 28
+ * 29                       0xFF
+ * 30                       0xFF
+ * 31                       0xFF
+ */
+#define SBP_MSG_FCNS_GLO             0x0072
+typedef struct __attribute__((packed)) {
+  u16 wn;        /**< GPS Week number [week] */
+  u32 tow_ms;    /**< GPS Time of week [ms] */
+  u8 fcns[32];  /**< GLONASS fequency number per orbital slot */
+} msg_fcns_glo_t;
+
+
 /** \} */
 
 #endif /* LIBSBP_OBSERVATION_MESSAGES_H */
--- a/haskell/src/SwiftNav/SBP.hs
+++ b/haskell/src/SwiftNav/SBP.hs
@@ -97,6 +97,7 @@ data SBPMsg =
    | SBPMsgEphemerisSbas MsgEphemerisSbas Msg
    | SBPMsgEphemerisSbasDepA MsgEphemerisSbasDepA Msg
    | SBPMsgExtEvent MsgExtEvent Msg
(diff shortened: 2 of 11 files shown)


c/include/libsbp/observation.h
```diff
@@ -612,6 +612,21 @@ coordinate system
...
+ *
+ * The message reports mapping information regrding GLONASS SV orbital and
+ * frequency slots. Frequency numbers -7 to 6 have been reported as 1 to 14
+ * correspondingly. 0xFF for invalid. The 0th element and elements from 29th
+ * to 30th of the map are always invalid.
+ */
+#define SBP_MSG_FCNS_GLO             0x0072
+typedef struct __attribute__((packed)) {
+  u16 wn;      /**< GPS Week number [week] */
+  double tow;     /**< GPS Time of week [s] */
```
adel-mamin: `tow` -> `tow_s`
Is it always going to be valid?

c/include/libsbp/observation.h
```diff
@@ -612,6 +612,21 @@ coordinate system
 } msg_almanac_glo_t;
 
 
+/** GLONASS SV orbital and frequency slots mapping information
+ *
+ * The message reports mapping information regrding GLONASS SV orbital and
+ * frequency slots. Frequency numbers -7 to 6 have been reported as 1 to 14
```
adel-mamin: Let's make the description a bit differently. Maybe providing a table like this:
```
index (SV orbital slot)  fcns[index]
0                        0xFF
1                        FCN for SV orbital slot 1
...                      ...
28                       FCN for SV orbital slot 28
29                       0xFF
30                       0xFF
31                       0xFF
```
BTW, why 32 slots in `fcns`?

c/include/libsbp/observation.h
```diff
@@ -612,6 +612,21 @@ coordinate system
 } msg_almanac_glo_t;
 
 
+/** GLONASS SV orbital and frequency slots mapping information
+ *
+ * The message reports mapping information regrding GLONASS SV orbital and
```
adel-mamin: `regrding` -> `regarding`

c/include/libsbp/observation.h
mbavaro: Well... we could indeed reduce it to 28.. but in the ICD the `n` field is 5 bits and its valid range is stated to be all 0-31 values. So, at the expense of +4 bytes, we will never risk to be unable to map something that is transmitted by the satellite. But this was just a thought.. feel free to re-spin if you think it's more appropriate.

c/include/libsbp/observation.h
mbavaro: Good point! I guess it should, as the ME should indeed have that piece of information when transmitting this message.. and a parser will benefit from placing this message in some time frame. However, if this introduces complexity (e.g., the message needing to be gated explicitly if time is not available), I would be in favor to either add a flag or remove the time. 
My reasoning was that indeed we do already have messages with a time tag like `MSG_GPS_TIME` and so handling could be similar.
One could also question to appropriateness of using a GPS time tag in a Glonass message.. but again in my reasoning was that being a "receiver time" rather than any specific constellation time.

dt-exafore:
@adel-mamin corrected, please review.

c/include/libsbp/observation.h
adel-mamin: It looks like `u32` is a more typical type for `tow` in SBP messages and units are in `[ms]`. Would it make sense to change `double tow_s` to `u32 tow_ms` then?
If time is unknown, then we could use `(u32)TOW_INVALID` and describe it explicitly in the comments. We may or may not use it, but at least it would give us more freedom to send it even if the time is really unknown.

PR #465, opened 2017-05-11
Add decimal SBP MSG_ID to doc

denniszollo:
generated PDF looks like this:

[sbp.pdf](https://github.com/swift-nav/libsbp/files/994674/sbp.pdf)

image


anth-cole:
+1 this will be so useful! another option would be a table at the start that maps them all

Issue #467, opened 2017-05-12
#462 broke pyserial driver on windows


denniszollo:
#462 
There is no termios on windows so we need to do something better.

PR #468, opened 2017-05-17
Fixup termios.error exception handling for Windows

denniszollo:
Intended as a fix for #467 

Anyone have something more pythonic to catch the low level exception on osx and linux but do nothingo n windows?


mfine:
Looks ok to me.

jck:
looks good.

Atleast, this is a good solution without trying to dig into where specifically the termios.error exception is raised.

PR #470, opened 2017-05-20
adding the spectrum analyzer message

mbavaro:
self explanatory, not decorated much yet


spec/yaml/swiftnav/sbp/specan.yaml
```diff
@@ -0,0 +1,65 @@
+# Packages define a logical collection of SBP messages. Packages can
```
mookerji: @mbavaro looks like you accidentally copied these comments over from the tutorial example....

python/sbp/specan.py
```diff
@@ -0,0 +1,140 @@
...
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+
+"""
```
mookerji: Note, you're going to have to manually update: https://github.com/swift-nav/libsbp/blob/master/python/sbp/table.py#L17, which is not programmatically generated right now.

mfine:
What is `specan`? Is that a term of art around spectrum analysis?

mbavaro:
@mfine yeah, not the best choice but it's the same as in PFWP
@mookerji I rebased, modified table.py and related examples .. do you think it is OK now?

mfine:
@mbavaro are there going to be more spectrum analyzer messages? It seems overkill to start a new file/group for this - can we just put it in `piksi` instead?

mbavaro:
@mfine more messages might be defined in the future: we don't deal much with interference at all, but I could envisage some messages to configure the spectrum analyzer traces (channels, bandwidths, averages, windows, ...), as well as messages for changing IF filter parameters or setting notches there. Unless it hurts anyone's sense of beautiful code, I would leave like this.

mfine:
@mbavaro well we can always add the grouping later - I'd would roll this into piksi.yaml. Other "single" groupings like ext-events haven't really been worth the split out and file pollution.

mookerji:
@mbavaro ack, yeah, I was going to echo @mfine's comment before but didn't finish the review. As an alternative to `piksi.yaml`, I would also recommend `acquisition.yaml` or `logging.yaml`. If the configuration you're describing comes to fruition, then the acquisition module might be more appropriate.

mbavaro:
I see it hurts your sense of beauty. OK I am moving it to piksi.yaml.

mfine:
Looks great, thanks!

mookerji:
:shipit:

PR #471, opened 2017-05-22
Adding `extern "C"` to sbp.h for inclusion into C++ programs

brentalanmiller:
Otherwise linking will fail with messages like `undefined reference to 'sbp_state_init(sbp_state_t*)'`

### DIFF
--- a/c/include/libsbp/edc.h
+++ b/c/include/libsbp/edc.h
@@ -13,8 +13,16 @@
 #ifndef LIBSBP_EDC_H
 #define LIBSBP_EDC_H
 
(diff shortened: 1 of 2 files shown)


ljbade:
@brentalanmiller thank you this is excellent!

Can you also wrap `include/libsbp/edc.h`? That is the only other file with extern functions.

brentalanmiller:
@ljbade Sure, no problem. Just committed.

mookerji:
Excellent, thanks @brentalanmiller and @ljbade !

PR #474, opened 2017-05-31
Add FCN and d_tau to MSG_EPHEMERIS_GLO

dt-exafore:
Based on https://github.com/swift-nav/exafore_planning/issues/750

### DIFF
--- a/haskell/sbp.cabal
+++ b/haskell/sbp.cabal
@@ -1,5 +1,5 @@
 name:                  sbp
-version:               2.2.3
+version:               2.2.3-dirty
 synopsis:              SwiftNav's SBP Library
 homepage:              https://github.com/swift-nav/libsbp
 license:               LGPL-3
--- a/haskell/sbp.cabal
+++ b/haskell/sbp.cabal
@@ -1,5 +1,5 @@
 name:                  sbp
-version:               2.2.3-dirty
+version:               2.2.3
 synopsis:              SwiftNav's SBP Library
 homepage:              https://github.com/swift-nav/libsbp
 license:               LGPL-3
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "libsbp",
-  "version": "2.2.3",
+  "version": "2.2.3-dirty",
   "description": "libsbp bindings for JavaScript. More information here: http://swift-nav.github.io/libsbp/",
   "files": [
     "javascript/*",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "libsbp",
-  "version": "2.2.3-dirty",
+  "version": "2.2.3",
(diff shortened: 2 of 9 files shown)


haskell/sbp.cabal
```diff
@@ -1,5 +1,5 @@
 name:                  sbp
-version:               2.2.3
+version:               2.2.3-dirty
```
peddie: Shouldn't this become `2.2.4`?

dt-exafore: I don't know. It was autogenerated.

mfine: @dt-exafore it means when you generated the code, there were uncommitted changes - this should get backed out.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "libsbp",
-  "version": "2.2.3",
+  "version": "2.2.3-dirty",
```
mfine: @dt-exafore also this.

haskell/sbp.cabal
dt-exafore: @mfine Does it mean that the process is, first, commit my changes in .yaml and python tests if any, second, generate other code and finally make one more commit with generated files?
I usually make change, generate code and commit everything at once.

mfine: Process is usually commit changes in `.yaml`, generate code, fix tests if any and commit. The `-dirty` means you ran the generator with uncommitted changes:

```
➜  ~/repos/libsbp git:(master) git describe --tags --dirty --always
v2.2.3
➜  ~/repos/libsbp git:(master) echo woops > README.md 
➜  ~/repos/libsbp git:(master) ✗ git describe --tags --dirty --always
v2.2.3-dirty
➜  ~/repos/libsbp git:(master) ✗ 
```

dt-exafore:
@mfine I generated it again and it removes "-dirty", however the version is still 2.2.3, should it be 2.2.4?

mfine:
@dt-exafore no, that usually happens during the releasing of the library.

nsirola:
Before merging, let's check if there were any more GLO ephemeris fields that we could add in the same go.

nsirola:
GLO ephemeris / non-immediate data elements that are not yet being reported in the SBP message (ICD section 4.4):

P - technological parameter (not relevant?)
P1 - flag of the immediate data updating (not relevant, already included in fit interval?)
P2 - flag of oddness or evenness of the value t_b (not relevant??)
P3 - flag indicating a number of satellites for which almanac is 
transmitted within given frame (not relevant?)
P4 - flag to show that ephemeris parameters are present (not relevant?)


Additionally there's the GLONASS almanac contains the Keplerian orbit elements etc. that is not decoded yet, save for:

tau_c - GLONASS time scale correction to UTC(SU) time
tau_gps - correction to GPS time relative to GLONASS time

anth-cole:
lgtm - I don't think the others are needed in the ephemeris

PR #475, opened 2017-06-01
python: handle bad exception from pyserial on EINTR.

gsmcmullin:
The port is still perfectly usable after an interrupted system call.  This should be handled internally in pyserial, but here's a workaround.


jck:
It looks like this is fixed in pyserial

https://github.com/pyserial/pyserial/blob/master/serial/serialposix.py#L514

PR #477, opened 2017-06-07
use setuptools_scm instead of version.py

jck:


jck:
I'm going to redo this carefully.

PR #478, opened 2017-06-08
Python deps

denniszollo:
Harmonize python dependencies with piksi_tools

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,7 +191,7 @@ test-c:
 
 test-python:
 	$(call announce-begin,"Running Python tests")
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -191,7 +191,7 @@ test-c:
 
 test-python:
 	$(call announce-begin,"Running Python tests")
-	cd $(SWIFTNAV_ROOT)/python/ && python -m pip install --user -r requirements.txt && python -m tox
+	cd $(SWIFTNAV_ROOT)/python/ && python -m pip install --user -r requirements.txt && pip install --user -r test_requirements.txt && python -m tox
```
jck: tox is the only thing which is needed. tox will install all the dependencies specified in `tox.ini` in a virtualenv.

PR #479, opened 2017-06-09
Windows c

denniszollo:
Allow libsbp c binding to be compiled by msvc compiler:

- Uses attribute packed for gcc
- Uses #pragma pack for msvc
- Attempts to autodetect the compiler in common.h
- if not autodetected, you will need to use preprocessor to define toolchain macro

Next step, an example that can compile cross platform, add extern C into structure definitinos

### DIFF
--- a/c/include/libsbp/acquisition.h
+++ b/c/include/libsbp/acquisition.h
@@ -26,6 +26,10 @@
 #include "common.h"
 #include "gnss.h"
 
+#if defined _MSC_VER || defined TOOLCHAIN_PRAGMA_PACK
+#pragma pack(1)
+#endif
+
 
 /** Satellite acquisition result
  *
@@ -36,7 +40,11 @@
  * ratio.
  */
 #define SBP_MSG_ACQ_RESULT         0x002F
+#if defined _MSC_VER || defined TOOLCHAIN_PRAGMA_PACKED
+typedef struct {
+#else
 typedef struct __attribute__((packed)) {
+#endif
   float cn0;    /**< CN/0 of best point [dB Hz] */
   float cp;     /**< Code phase of best point [chips] */
   float cf;     /**< Carrier frequency of best point [hz] */
@@ -49,7 +57,11 @@ typedef struct __attribute__((packed)) {
 * Deprecated.
  */
 #define SBP_MSG_ACQ_RESULT_DEP_C   0x001F
+#if defined _MSC_VER || defined TOOLCHAIN_PRAGMA_PACKED
+typedef struct {
+#else
 typedef struct __attribute__((packed)) {
+#endif
   float cn0;    /**< CN/0 of best point [dB Hz] */
   float cp;     /**< Code phase of best point [chips] */
   float cf;     /**< Carrier frequency of best point [hz] */
@@ -62,7 +74,11 @@ typedef struct __attribute__((packed)) {
 * Deprecated.
  */
 #define SBP_MSG_ACQ_RESULT_DEP_B   0x0014
+#if defined _MSC_VER || defined TOOLCHAIN_PRAGMA_PACKED
+typedef struct {
+#else
 typedef struct __attribute__((packed)) {
+#endif
   float snr;    /**< SNR of best point. Currently in arbitrary SNR points, but will
 be in units of dB Hz in a later revision of this message.
  */
@@ -77,7 +93,11 @@ be in units of dB Hz in a later revision of this message.
 * Deprecated.
  */
 #define SBP_MSG_ACQ_RESULT_DEP_A   0x0015
+#if defined _MSC_VER || defined TOOLCHAIN_PRAGMA_PACKED
+typedef struct {
+#else
 typedef struct __attribute__((packed)) {
+#endif
   float snr;    /**< SNR of best point. Currently dimensonless, but will have
 units of dB Hz in the revision of this message.
  */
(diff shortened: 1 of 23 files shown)


denniszollo:
I will merge this shortly unless I hear objections.  Reviewed by @fnoble a few weeks back and he thought it looked good.

mfine:
This is kind of nasty. Is there a conventional/idiomatic way to handle this? My preference would be to avoid the top defines and just do all the macro preprocessing directly - it's generated code, and I don't think the extra macro abstractions bring much.

denniszollo:
You are okay with the macros like "TOOLCHAIN_ATTR_PACKED", but would rather have the toolchain detected and this macro defined in each file instead of the extra layer of indirection in common.h? 


I did quite a lot of research and didn't find an agreed upon strategy for this, and I think my version has been used before.

jck:
I defines vs modifying the generator

mfine:
> You are okay with the macros like "TOOLCHAIN_ATTR_PACKED", but would rather have the toolchain detected and this macro defined in each file instead of the extra layer of indirection in common.h?

Not ok with that macro either. I would have removed the toolchain detection in common.h and just done the check on each usage. Around each struct definition, I would have done:

```c
#ifndef _MSC_VER
typedef struct __attribute__((packed)) {
#else
typedef struct {
#endif
```

At the beginning of the message file, I would have done:

```c
#ifdef _MSC_VER
pack(1)
#endif
```

At the end of the message file, I would have done:

```c
#ifdef  _MSC_VER
#pragma pack()
#endif
```

Already, the common #define of TOOLCHAIN_ATTR_PACKED in all of the messages file is way redundant, and you're locating it there just to line it up with the weird MSC behavior of beginning and end of file.

Also note above how there's no brittle error on not detecting GCC or MSC - there's no other compiler type detections in the library and there's other non windows compilers besides GCC that also support packed attributes (we have downstreams that use other compilers than GCC and would seemingly get broken by this change). By testing just MSC, you make a minimal change that impacts only the functionality you're trying to bring in, and not everyone else.

I don't think building abstractions in macros is a good idea - I think it makes the code a lot harder to read than being direct with what the macros are doing. And especially because its generated code, it's easy to generate all of the boilerplate and be direct about it.

> I prefer defines vs modifying the generator

In the code you'll see the defines are defined in the generator for each messages file. There shouldn't ever be any issue with updating the generator.

mfine:
Actually, I would have not made the first check negative logic - positive instead:

```c
#ifdef _MSC_VER
typedef struct {
#else
typedef struct __attribute__((packed)) {
#endif
```

mfine:
See: https://github.com/swift-nav/libsbp/compare/master...mfine:mfine/msc

ljbade:
@mark your approach makes sense, and detecting just the Microsoft compiler for these sort of quirks is what I have seen before in other larger codebases

We should probably put in the readme a list of the compilers/versions that we have tested/support

gsmcmullin:
GCC supports `#pragma pack` but it makes me want to puke: https://gcc.gnu.org/onlinedocs/gcc-7.1.0/gcc/Structure-Layout-Pragmas.html#Structure-Layout-Pragmas
It's still probably better than doing both and using ugly macros.

You can also make the attribute go away by doing `#define __attribute__(x)` or `-D__attribute__\(x\)=` in the Makefile.

denniszollo:
I do not understand why we want to treat the microsoft compiler as a "one-off".  Many other C compilers are lacking in the _attribute_ type of interface (the intel c compiler for one). 

I think that having the toolchain macro makes it extendible to other systems that we don't know about.   This is the embedded world; there are lots of strange toolchains.   If we put the MSV_V inline like this, we are making it very hard for someone to use these headers on a complier we aren't planning for.

mfine:
Checks can be for different compilers can be added as needed - don't see what the issue is with extending to support other compilers as needed. You could swap the logic, but then you'll need to cover all the known supported platforms - GCC, CLANG, etc. - which seems more risky.

denniszollo:
The use case I'm worried about is for someone who isn't at Swift and can't generate the code but wants to use the headers.  

If they don't have __attribute_ packed in their toolchain but they aren't Windows, in Mark's approach they will have to define _MSC_VER, which I don't feel that comfortable with.  It would be necessary to hand modify all the headers in that case most likely.

Does anyone think that use case is probable or something to consider?

mfine:
I'm not saying to define `_MSC_VER` - we can add more toolchain checks on behalf of customers as they need them.

michaelaeriksen:
Any plans to merge this?

denniszollo:
After a year of this languishing, I am reviving this PR.  I think I have a good compromise between my original implementation and Mark's comments that the abstraction was indirect and overblown.

There is no toolchain or extra abstraction, and the logic means that it tends towards the GCC version.  It does not rely on any unclear macros in my opinion

However, if a user has the code as an artifact and no regeneration capabilities, they could get any behavior they want in the header files just through defining some extra macros.

mfine:
This approach looks ok to me - /cc @silverjam for more feedback around this approach.

silverjam:
My 2c: I would make common macros in common.h that handle the pragma pack and the __attribute__ gymnastics, then detect and configure the macros in common.h based on specific compilers versions... With the final fall through  (when we don't know what compiler it is) being "you configure things".

But... I've seen this done both ways, I think the more common approach is to make the code generic by inserting macros.

silverjam:
Support for this was merged in https://github.com/swift-nav/libsbp/pull/568

PR #487, opened 2017-06-20
Add IOD to SBP_MSG_EPHEMERIS_GLO

dt-exafore:
Based on @anth-cole request here https://github.com/swift-nav/piksi_firmware_private/pull/840

### DIFF
--- a/spec/yaml/swiftnav/sbp/observation.yaml
+++ b/spec/yaml/swiftnav/sbp/observation.yaml
@@ -609,8 +609,59 @@ definitions:
           desc: Acceleration vector of the SV at tb in PZ-90.02 coordinates sys
           units: m/s^2
 
- - MSG_EPHEMERIS_GLO:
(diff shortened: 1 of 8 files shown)


spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -658,6 +660,58 @@ definitions:
...
+          fill: double
+          map_by: sid
+          desc: Acceleration vector of the SV at tb in PZ-90.02 coordinates sys
+          units: m/s^2
+      - fcn:
+          type: u8
+          desc: Frequency slot. FCN+8 (that is [1..14]). 0 or 0xFF for invalid
+      - iod:
+          type: u8
+          desc: Issue of ephemeris data
```
ljbade: This should have a note that this is the lower 7 bits of tb as defined in the RTCM 3.1 specification, for DF110

ljbade:
@dt-exafore added a small comment on some more detail we should add to the field description

PR #498, opened 2017-07-03
Support lts-8

mfine:
+ Add support for lts-8.
+ Also add explicit configuration for lts-5 and lts-6.
+ Add `lower` pvp bounds configuration - this should only be run with `lts-5`.
+ Add strict designations to message fields.


mfine:
resolves #422

PR #499, opened 2017-07-10
add makefile target to generate all langauges

jck:


Makefile
```diff
@@ -54,6 +54,7 @@ help:
 	@echo
 
 all: deps-generator c python javascript java haskell docs
+gen: gen-c gen-python gen-javascript gen-java gen-haskell
```
mookerji: @jck is this not what `make all` already does?

ljbade: yes I have used `make all` before without problems

PR #509, opened 2017-09-07
python client cleanup, improvements

jck:
* convert to pep8(4 spaces)
* add support for TCPDriver reconnect

### DIFF
--- a/python/sbp/client/drivers/base_driver.py
+++ b/python/sbp/client/drivers/base_driver.py
@@ -8,59 +8,61 @@
 # EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
+
 class BaseDriver(object):
-  """
-  BaseDriver
-
-  The :class:`BaseDriver` class wraps IO sources of SBP messages and provides
-  context management.
-
-  Parameters
-  ----------
-  handle : port
-    Stream of bytes to read from and write to.
-  """
-  def __init__(self, handle):
-    self.handle = handle
-
-  def __enter__(self):
-    self.flush()
-    return self
-
-  def __exit__(self, *args):
-    self.flush()
-    self.close()
-
-  def read(self, size):
     """
-    Read wrapper.
+    BaseDriver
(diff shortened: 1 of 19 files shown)


python/sbp/client/drivers/base_driver.py
```diff
@@ -8,59 +8,61 @@
...
 
-  def write(self, s):
-    """
-    Write wrapper.
+        Parameters
+        ----------
+        size : int
+          Number of bytes to read.
+        """
+        return self.handle.read(size)
```
pmiettinen: One thing to consider (most likely in a separate PR though) is to do the reads in bigger chunks into a internal memory buffer. This helps with CPU load at least with serial ports. Reading in too big blocks will cause noticeable delays though.

python/sbp/client/examples/bytes.py
```diff
@@ -11,44 +11,57 @@
...
+        nargs=1,
+        help="specify the filename to write to.")
+    return parser.parse_args()
+
 
 def main():
-  args = get_args()
+    args = get_args()
+
+    with PySerialDriver(args.serial_port[0], args.baud[0]) as driver:
```
pmiettinen: To decrease indentation levels and increase readability, I think the context manager stacking could be like:

```Python
   with PySerialDriver(args.serial_port[0], args.baud[0]) as driver,\
        Handler(driver.read, driver.write) as link,\
        ByteLogger(args.filename[0]) as logger:

        link.add_callback(logger)
        link.start
```

Matter of taste of course.

python/sbp/client/handler.py
```diff
@@ -18,256 +17,267 @@
...
+            self._source(msg, **metadata)
+
+    class _SBPQueueIterator(object):
+        """
+        Class for upstream iterators.  Implements callable interface for adding
+        messages into the queue, and iterable interface for getting them out.
+        """
+
+        def __init__(self, maxsize):
+            self._queue = Queue(maxsize)
```
pmiettinen: collections.deque has better performance.

jck:
@pmiettinen great suggestions. The third commit(https://github.com/swift-nav/libsbp/pull/509/commits/1d582b92b4a72fa15d06604c9ec1a006cb63f338) is the only non stylistic change in this PR. Will address the other comments in another PR.

PR #514, opened 2017-09-29
Change default baud rate, make example more verbose

silverjam:
Change the default baudrate to 115200 since this is more common?  Configuring to 1000000 cause the example to fail with Piksi Multi on OS X.

Add some print statement so we have an of what the app is trying to do, since this isn't a unix utility it shouldn't be a big deal to make it a little more verbose.

### DIFF
--- a/c/example/example.c
+++ b/c/example/example.c
@@ -20,23 +20,28 @@ void setup_port()
 {
   int result;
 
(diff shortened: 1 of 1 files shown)


mfine:
:shipit:

PR #515, opened 2017-09-29
Remove `deps` target; fix help/hint text

silverjam:
The 'deps' target doesn't exist anymore, deps are installed as part of building the libs for a particular language so it seems unnecessary.

Fix the "hint" text that's displayed to help the user install the right dependenvies prior to building.  The echo command needs '-e' to evaluate \n as a newline.


mookerji:
:shipit: !

PR #516, opened 2017-10-04
wip: serial double buffering

denniszollo:
Making this PR so I can add comments

### DIFF
--- a/python/sbp/client/drivers/pyserial_driver.py
+++ b/python/sbp/client/drivers/pyserial_driver.py
@@ -8,9 +8,41 @@
 # EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
+import threading
+
 from .base_driver import BaseDriver
 import serial
 import serial.tools.list_ports
+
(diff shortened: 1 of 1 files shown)


python/sbp/client/drivers/pyserial_driver.py
```diff
@@ -8,9 +8,41 @@
...
+
+    def connection_lost(self, exc):
+        self.transport = None
+        super(Buf, self).connection_lost(exc)
+
+    def data_received(self, data):
+        """Buffer received data, find TERMINATOR, call handle_packet"""
+        # print('pre: prod acq')
+        # self.new_data.acquire()
+        # print('post: prod acq')
```
denniszollo: What is the terminator?  I think we should make the terminator 0x55 perhaps.  By default it is the null terminator.  I think the "jerkiness" of the protocol is an artifact the fact that the chunks we read and wait for are determined by the null terminator.

python/sbp/client/drivers/pyserial_driver.py
```diff
@@ -75,13 +111,25 @@ def read(self, size):
...
           Number of bytes to read.
         """
-        try:
-            return self.handle.read(size)
-        except (OSError, serial.SerialException):
-            print
-            print "Piksi disconnected"
-            print
-            raise IOError
+        # print('pre: consume acq')
```
denniszollo: What happens when the serial FD goes away?  Does it throw the same exception as before?

PR #517, opened 2017-10-06
Update Java parser API

jonrwat:
The architecture of the Java parser is updated to be consistent with the
SnapCam project.  This includes development of the parser which does not
execute its own thread, but instead ingests supplied serial data a byte at
the time, and returns the parsed message once an entire message is
received.


jonrwat:
Did not mean to open pull-request to swift repository.  Making update to forked branch.

PR #519, opened 2017-10-25
Rename the js library to sbp.

mfine:
Be consistent with other packages.

### DIFF
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -2,15 +2,15 @@
 
 ## Installing from npm
 
-`npm install libsbp`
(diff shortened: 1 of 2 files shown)


ljbade:
lgtm, but should deprecate libsbp on npm, and add a note that points to sbp

also make sure to create the sbp package under the swiftnav org in npm

mfine:
@ljbade can you create the sbp package under the swiftnav org?

mfine:
@ljbade can you create the sbp package under the swiftnav org?

PR #524, opened 2017-11-14
Adding GLONASS code-phase bias message

scarcanague:
The purpose of this message is to provide the GLONASS code-phase bias. This message is notably expected to be filled in with the RTCM 1230 message.
I was not planning on adding the `GLONASS Code-Phase bias indicator`, as it is simply a sign instruction.

@anth-cole : Could you make a review? You should simply review the yaml file, as all the rest is generated automatically

### DIFF
--- a/spec/yaml/swiftnav/sbp/observation.yaml
+++ b/spec/yaml/swiftnav/sbp/observation.yaml
@@ -1755,3 +1755,33 @@ definitions:
           type: double
           units: rad
           desc: Argument of perigee at instant of t_lambda
+          
+ - MSG_GLO_BIASES:
+    id: 0x0075
+    short_desc: GLONASS L1/L2 Code-Phase biases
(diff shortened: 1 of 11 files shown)


spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -1755,3 +1755,33 @@ definitions:
           type: double
           units: rad
           desc: Argument of perigee at instant of t_lambda
+          
+ - MSG_GLO_BIASES:
+    id: 0x0075
```
anth-cole: This looks good to me, catches everything we need from the 1230 and I think it's right to leave out the alignment field. If they're unaligned we'll just ignore them.

mfine:
File generation looks ok - looks like you just need to bump the message count check in python.

PR #525, opened 2017-11-22
 Add MSG_SETTINGS_WRITE_RESP

gsmcmullin:


spec/yaml/swiftnav/sbp/settings.yaml
```diff
@@ -43,6 +43,29 @@ definitions:
             process to this message when it is received from sender ID
             0x42.
 
+ - MSG_SETTINGS_WRITE_RESP:
+    id: 0x00AF
+    short_desc: Acknowledgement with status of MSG_SETTINGS_WRITE
+    desc: |
+      Return the status of a write request with the current value of the
```
ljbade: Hmm "current value" is a bit confusing as it sounds like the pre-update value. Perhaps "updated value"?

If you think that is also confusing could add second sentence like "If the write is rejected the current value is returned instead."

PR #528, opened 2017-12-01
adding fftmonitor

bryanbrady:


python/sbp/client/util/fftmonitor.py
```diff
@@ -0,0 +1,198 @@
...
+printing them out.
+"""
+
+# from sbp.client.drivers.pyserial_driver import PySerialDriver
+
+from collections import defaultdict
+from sbp.client.drivers.network_drivers import TCPDriver
+from sbp.client import Handler, Framer
+from sbp.gnss import GPSTimeSec
+try:
```
bryanbrady: This needs to be fixed.

python/sbp/client/util/fftmonitor.py
```diff
@@ -0,0 +1,196 @@
...
+            self.incomplete_ffts[channel][timestamp]['frequencies'],
+            frequencies,
+            axis=0)
+
+        self.incomplete_ffts[channel][timestamp]['amplitudes'] = np.append(
+            self.incomplete_ffts[channel][timestamp]['amplitudes'],
+            amplitudes,
+            axis=0)
+
+        if len(self.incomplete_ffts[channel][timestamp]['frequencies']) == 512:
```
axlan: Is this 512 available as a defined constant somehwhere?

python/sbp/client/util/fftmonitor.py
```diff
@@ -0,0 +1,196 @@
...
+            current_fft = self.incomplete_ffts[channel].pop(timestamp)
```
axlan: What's the expected behavior if you miss a message?

bryanbrady: Ignore it and wait for the next one.

python/sbp/client/util/fftmonitor.py
bryanbrady: If it is, I haven't found it. Probably should be, though.  It was hardcoded in the Spectrum Analyzer in the Console

PR #529, opened 2017-12-07
Bb/monitor updates

bryanbrady:


python/sbp/client/util/settingmonitor.py
```diff
@@ -0,0 +1,40 @@
...
+monitor SBP_MSG_SETTINGS_READ_RESP messages for desired values
+"""
+
+import time
+
+class SettingMonitor(object):
+    def __init__(self):
+        self.settings = []
+
+    def capture_setting(self, sbp_msg, **metadata):
```
axlan: While I think this level of integration is fine for this utility, I think it would be a bit confusing for someone to use without a comment explaining the usage or an example. Alternatively if this Monitor handled the registering the callback I think it might be more self explanatory.

bryanbrady:
@axlan Thanks for the feedback. I've added comments and a usage example.

python/sbp/client/util/settingmonitor.py
```diff
@@ -36,5 +56,72 @@ def wait_for_setting_value(self, section, setting, value, wait_time=5.0):
...
+                time.sleep(2)
+
+                # Capture setting messages
+                link.add_callback(monitor.capture_setting,
+                                  SBP_MSG_SETTINGS_READ_RESP)
+
+                link.add_callback(print_setting, SBP_MSG_SETTINGS_READ_RESP)
+
+                # Disable spectrum analyzer
+                link(MsgSettingsWrite(setting='%s\0%s\0%s\0' % (
```
axlan: It might be worth providing utility functions for handling the NULL deliminated settings.

python/sbp/client/util/settingmonitor.py
```diff
@@ -12,21 +12,41 @@
...
 
     def wait_for_setting_value(self, section, setting, value, wait_time=5.0):
+        """Function to wait wait_time seconds to see a
+        SBP_MSG_SETTINGS_READ_RESP message with a user-specified value
+        """
         expire = time.time() + wait_time
         ok = False
         while not ok and time.time() < expire:
-            settings = filter(lambda x: (x[0],x[1]) == (section, setting),
+            settings = filter(lambda x: (x[0], x[1]) == (section, setting),
```
axlan: This sort of depends on the intended usage, but it seems like this would make it so that you couldn't do something like:
MsgSettingsReadReq msg1
MsgSettingsReadReq msg2
...
wait_for_setting_value msg1
wait_for_setting_value msg2

since it's possible that you'd end up clearing the second message when you call the first wait. It seems like coupling the request and the wait together makes sense. Alternatively, you could just clear the messages that match the requested type and use different logic for checking if they're in the buffer.

python/sbp/client/util/settingmonitor.py
bryanbrady: I agree. The same pattern shows up in the Console's settings view and I'm guessing other places, too. I think it might be better to do this as a separate task once we identify other places where this pattern is used.

python/sbp/client/util/settingmonitor.py
bryanbrady: The settings/messages aren't cleared in the wait (`settings` -vs- `self.settings`). That needs to happen explicitly.

However, after looking more at this logic, I'm liking the part about looking at the last message less and less. If the user wants to see the "last" message, then they can explicitly clear the messages. I like your idea of adding the ability to clear only certain messages.

bryanbrady: Hmm, I still think we need that "last" message logic in there at some point though, since there may well be multiple messages of the same type in there.

python/sbp/client/util/settingmonitor.py
denniszollo: If/when we redesign settings I'd like to use a different setting deliminator.

python/sbp/client/util/settingmonitor.py
bryanbrady: @axlan I think the last message logic should stay in there, but I'm open to suggestions. I've added the ability to selectively clear the settings based on a combination of section/setting/value.

python/sbp/client/util/settingmonitor.py
pmiettinen: 110% support for different deliminator. C string functions don't work well with the current one.

PR #533, opened 2017-12-12
Add a flag that is high when antenna is shorted, according to MAX4829 IC

lstrz:
# Design Notes

Documents heartbeat message changes introduced in swift-nav/piksi_firmware_private#1497, when fixing swift-nav/firmware_team_planning#394.


denniszollo:
lgtm

PR #536, opened 2018-01-02
Precise orbit and clock message

scarcanague:
This PR adds a message to be able to support RTCM-like orbit/clock precise correction message. 

cc: @benjamin0 @swiftnav-simon

### DIFF
--- a/spec/yaml/swiftnav/sbp/ssr.yaml
+++ b/spec/yaml/swiftnav/sbp/ssr.yaml
@@ -1,0 +1,226 @@
+# Copyright (C) 2018 Swift Navigation Inc.
+# Contact: Swift Navigation <dev@swiftnav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+package: swiftnav.sbp.ssr
+description: Precise State Space Representation (SSR) corrections format
+stable: False
+public: True
+include:
+  - types.yaml
+  - gnss.yaml
+definitions:
+ - CodeBiasesContent:
+     short_desc: SSR code biases corrections for a particular satellite.
+     desc: |
+      Code biases are to be added to pseudorange.
+      The corrections are conform with typical RTCMv3 MT1059 and 1065.
+     fields:
+     - code:
+            type: u8
+            desc: Signal constellation, band and code
+            fields:
+              - 0-7:
+                  values:
+                      - 0: GPS L1CA
+                      - 1: GPS L2CM
+                      - 2: SBAS L1CA
+                      - 3: GLO L1CA
+                      - 4: GLO L2CA
+                      - 5: GPS L1P
+                      - 6: GPS L2P
+     - value:
+            type: s16
+            units: 0.01 m
+            desc: Code bias value
+ - PhaseBiasesContent:
+     short_desc: SSR phase biases corrections for a particular satellite.
+     desc: |
+      Phase biases are to be added to carrier phase measurements.
+      The corrections are conform with typical RTCMv3 MT1059 and 1065.
+     fields:
+     - code:
+            type: u8
+            desc: Signal constellation, band and code
+            fields:
+              - 0-7:
+                  values:
+                      - 0: GPS L1CA
+                      - 1: GPS L2CM
+                      - 2: SBAS L1CA
+                      - 3: GLO L1CA
+                      - 4: GLO L2CA
+                      - 5: GPS L1P
+                      - 6: GPS L2P
+     - integer_indicator:
+            type: u8
+            desc: Indicator for integer property
+     - widelane_integer_indicator:
+            type: u8
+            desc: Indicator for two groups of Wide-Lane(s) integer property
+     - discontinuity_counter:
+            type: u8
+            desc: |
+              Signal phase discontinuity counter. 
+              Increased for every discontinuity in phase.
+     - bias:
+            type: s32
+            units: 0.1 mm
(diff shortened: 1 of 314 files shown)


spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -0,0 +1,111 @@
...
+include:
+  - types.yaml
+  - gnss.yaml
+definitions:
+
+ - OrbitClockHeader:
+    short_desc: Header for orbit/clock message.
+    desc: Header of a SSR orbit/clock correction message.
+    fields:
+        - t:
```
benjamin0: `time`?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -0,0 +1,256 @@
...
+            type: u8
+            units: s
+            desc: Update interval between consecutive corrections
+        - iod_ssr:
+            type: u8
+            desc: |
+                IOD of the SSR correction. A change of Issue Of Data
+                SSR is used to indicate a change in the SSR 
+                generating configuration
+        - n_corr:
```
mfine: General question - do we want or need to pack content? Observations are packed because of bandwidth, but this pattern around size + index field is a little unwieldy. Are we sure we want to open up another use case for it?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -0,0 +1,256 @@
...
+    id: 0x005E1
+    short_desc: Precise code biases correction
+    desc: |
+      The precise code biases message is to be added
+      to the pseudorange of the corresponding signal
+      to get corrected pseudorange. It is typically 
+      an equivalent to the 1059 and 1065 RTCM message types
+    fields:
+        - header:
+            type: OrbitClockHeader
```
mfine: Code biases are sharing the OrbitClockHeader?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -0,0 +1,256 @@
...
+            type: u16
+            units: 1 / 256 semi-circle
+            desc: Satellite yaw angle
+     - yaw_rate:
+            type: s16
+            units: 1 / 8192 semi-circle / s
+            desc: Satellite yaw angle rate
+     - biases:
+            type: array
+            fill: PhaseBiasesValues
```
mfine: I don't think this will work. I believe fill is greedy - it will take as many `PhaseBiasesValues` as there exist left in the frame. The issue is that this structure `PhaseBiasesPackedContent` is also a fill in `MSG_SSR_PHASE_BIASES` - hence the `PhaseBiasesValues` fill will override the `PhaseBiasesPackedContent`. I think generally you can only have a single fill in a message, and it has to be the last thing in the message.

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -0,0 +1,256 @@
...
+                generating configuration
+     - dispersive_bias:
+            type: u8
+            desc: |
+                Indicator for the dispersive phase biases property.
+     - mw_consistency:
+            type: u8
+            desc: |
+                Consistency indicator for Melbourne-Wubbena linear combinations
+     - n_corr:
```
mfine: Similar comment here about using the `n_x` scheme.

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -0,0 +1,256 @@
...
+                First nibble is the size of the sequence (n),
+                second nibble is the zero-indexed counter 
+                (ith packet of n)
+ - PhaseBiasesValues:
+     short_desc: SSR phase biases corrections for a particular satellite.
+     desc: |
+      Code biases are to be added to pseudorange.
+      The corrections are conform with typical RTCMv3 MT1059 and 1065.
+     fields:
+     - code:
```
mfine: I might be missing something - does the GnssSignal field in PhaseBiasesPackedContent not handle the code already? Or is a signal's code not material and that a given signal will have multiple codes. Would it make more sense then to have just sat?

spec/yaml/swiftnav/sbp/ssr.yaml
denniszollo: I think bandwidth is an important thin to optimize for this message thinking about how we may want use it as the corrections delivery to devices.  I don't see this being consumed very often (even less often than obs messages probably).

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -0,0 +1,256 @@
...
+      Code biases are to be added to pseudorange.
+      The corrections are conform with typical RTCMv3 MT1059 and 1065.
+     fields:
+     - sid:
+            type: GnssSignal
+            desc: GNSS signal identifier (16 bit)
+     - yaw:
+            type: u16
+            units: 1 / 256 semi-circle
+            desc: Satellite yaw angle
```
denniszollo: this cannot encode all possible yaw angles I don't think

2^16 / 2^8 is 256 ( less than 0 to 360)

spec/yaml/swiftnav/sbp/ssr.yaml
scarcanague: @mark Yes I see your point, and that was definitely one of my question so thx for the feedback. 
The idea was that:
-  we dont want to use the corrections until we received all of them.
- There is a common data that would bandwidth if repeated on a single satellite basis
But I see the pain in term of encoding/decoding.
So for the first concern, there are other ways to reach that goal in the filter.  And for the bandwidth concern, I dont think we will send corrections like this to the user in the short term anyway (only used for the network engine interface), as the Skylark user correction message format is far from being defined at the moment.
So I would propose to switch to a static 1 message per satellite correction.
The rate is expected to be 1 orbit/clock/code message every 5s per satellite, so a burst of 3 * (32+24) messages every 5s for GPS+GLONASS. Sounds good?
What do you think about it?

spec/yaml/swiftnav/sbp/ssr.yaml
scarcanague: Yes, the only difference in the headers in RTCM is the satellite reference datum field, which is proposed to be dropped anyway sbp orbit/clock message header

PR #547, opened 2018-02-22
[WIP] Add NetworkUsage struct and MSG_NETWORK_BANDWIDTH_USAGE

benjaminaltieri:
To support async bandwidth metrics to be reported to the console the following message is being added to the piksi module:

MSG_NETWORK_BANDWIDTH_USAGE 0x00BD

While there is alignment to some degree with the MSG_NETWORK_STATE_* messages, the interface string being a common field, this is meant to me emitted as streaming data from the system to indicate real time data usage.

### DIFF
--- a/spec/yaml/swiftnav/sbp/piksi.yaml
+++ b/spec/yaml/swiftnav/sbp/piksi.yaml
@@ -484,6 +484,45 @@ definitions:
             - 0:
                 desc: IFF_UP - interface is up
 
+ - NetworkUsage:
+    short_desc: Bandwidth usage measurement for a single interface.
+    desc: |
+      The bandwidth usage for each interface can be reported
+      within this struct and utilize multiple fields to fully
+      specify the type of traffic that is being tracked. As
+      either the interval of collection or the collection time
+      may vary, both a timestamp and period field is provided,
+      though may not necessarily be populated with a value. 
+    fields:
+      - duration:
+          type: u64
+          units: ms
+          desc: Duration over which the measurement was collected
+      - total_bytes:
+          type: u64
+          desc: Number of bytes handled in total within period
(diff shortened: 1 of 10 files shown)


mfine:
LGTM!

benjaminaltieri:
Thanks @mfine! Do I need to commit any of the generated artifacts (or any other steps)? This is my first time adding to libsbp

mfine:
Yes! So it's good to do an initial commit like this, collect feedback - maybe from others like @denniszollo - and then generate the languages. I usually just generate things with:

```
 $ make deps-generator gen-c gen-python gen-javascript gen-java gen-haskell
```

spec/yaml/swiftnav/sbp/piksi.yaml
```diff
@@ -484,6 +484,45 @@ definitions:
...
+      specify the type of traffic that is being tracked. As
+      either the interval of collection or the collection time
+      may vary, both a timestamp and period field is provided,
+      though may not necessarily be populated with a value. 
+    fields:
+      - duration:
+          type: u64
+          units: ms
+          desc: Duration over which the measurement was collected
+      - total_bytes:
```
cbeighley: Does total bytes just equal tx_bytes + rx_bytes?

benjaminaltieri: Agree if we can commit to always having tx/rx split out. It seems reasonable but in a generic case we may not have access to tx/rx separately? The intent was for total to be the metric, and tx/rx and option added level of detail.

benjaminaltieri:
Alright I've run the generator and added the new source. Can I get a final thumbs up? @denniszollo any gotchas?

PR #550, opened 2018-03-09
TCPDriver ignore interrupted system call exeption

jck:


python/sbp/client/drivers/network_drivers.py
```diff
@@ -74,7 +75,11 @@ def read(self, size):
             return data
         except socket.timeout:
             self._connect()
-        except socket.error, msg:
+        except socket.error as e:
+            # this is fine
+            if e[0] == errno.EINTR:
```
silverjam: Should we retry if this happens?

jck: read is called in a while loop in a higher layer, so it ends up being retried.

PR #552, opened 2018-03-13
[Do not merge] Protobuf c bindings build scripts

benjaminaltieri:
Crack at a source generation step. Cmake script can build a library from generated sources.


benjaminaltieri:
Will move into buildroot package

PR #553, opened 2018-03-13
Add SBAS position quality

anth-cole:
This is the proposal for how to handle SBAS position quality.

Still to do
- Add generated code

@denniszollo

### DIFF
--- a/spec/yaml/swiftnav/sbp/navigation.yaml
+++ b/spec/yaml/swiftnav/sbp/navigation.yaml
@@ -21,11 +21,11 @@ description: |
   phase ambiguity. The pseudo-absolute position solution uses a
   user-provided, well-surveyed base station position (if available)
   and the RTK solution in tandem.
-  
+
   When the inertial navigation mode indicates that the IMU is used,
   all messages are reported in the vehicle body frame as defined by
   device settings.  By default, the vehicle body frame is configured to be
-  coincident with the antenna phase center.  When there is no inertial 
+  coincident with the antenna phase center.  When there is no inertial
   navigation, the solution will be reported at the phase center of the antenna.
 
 stable: True
@@ -137,7 +137,7 @@ definitions:
                   (range 0-999999999)
 
  - MSG_DOPS:
-    id: 0x0208
+    id: 0x0216
     short_desc: Dilution of Precision
     desc: |
       This dilution of precision (DOP) message describes the effect of
@@ -185,10 +185,10 @@ definitions:
                       - 2: Differential GNSS (DGNSS)
                       - 3: Float RTK
                       - 4: Fixed RTK
-
+                      - 5: SBAS Position
 
  - MSG_POS_ECEF:
-    id: 0x0209
+    id: 0x0217
     short_desc: Single-point position in ECEF
     desc: |
       The position solution message reports absolute Earth Centered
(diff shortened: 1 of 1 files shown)


cbeighley:
@anth-cole I don't think you should be revv'ing any message ID's here - I think we simply define 5 to be SBAS for all relevant messages.

anth-cole:
Thanks, removed the message ID rev's that makes the PR pretty simple thanks!

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -185,7 +185,7 @@ definitions:
                       - 2: Differential GNSS (DGNSS)
                       - 3: Float RTK
                       - 4: Fixed RTK
-
+                      - 5: SBAS Position
```
nsirola: This should probably also have 5 as either Dead reckoning or reserved value so that the same enum can be used here and in the other messages?

cbeighley: Yeah, should keep values the same across messages. Can call `5` undefined here.

denniszollo: agreed, lets leave 5 reserved

denniszollo:
What about velocities:
  - won't corrections from SBAS impact the doppler derived velocity solution?  Especially if the velocities are coming out of the full Kalman Filter?  
  - what about Time; won't SBAS code corrections impact our timing accuracy as well?
  - what about our ranges; are we going to correct or change observations at all on the way out based upon info from SBAS?

anth-cole:
@cbeighley done - good point

@denniszollo it'll impact them marginally, but currently we don't apply the rate of change to the doppler obs, and so the velocity solution will still be dominated by the doppler observations and should be roughly the same quality.

SBAS won't impact our time as this comes from the ME single epoch solution and this won't have the SBAS corrections applied. Also, the impact will be really minimal overall even if they were applied. Time also only gets marked 'gps derived' or not gps derived so this fits even with SBAS.

Ranges won't be affected, they'll be broadcast in raw form, and the SBAS corrections also broadcast, so they would need to be applied for any post-processing results

switanis:
What about NMEA output? How SBAS will be mapped there?

While a separate value will be easier for us and our testing I'm afraid it will be harder for customers as they will have to modify their code to use the new mode. In situation like this we should think about customer first.

nsirola:
@switanis NMEA specification says "D = Differential mode Corrections from ground stations or Satellite Based Augmentation System (SBAS)", so it will map both DGNSS and SBAS enums to this same indicator.

anth-cole:
@switanis this is purely the SBP PR, NMEA is handled in a PR in PFWP, that will obviously need updating after this goes in. We have pretty clear design guidelines that prioritize doing a 'clean' design over servicing legacy systems, and I think this fits in that boat.

@cbeighley is this good for you now?

cbeighley:
@anth-cole Under messages for which only differential solutions are relevant (baseline, heading, others?) you might add `- 6: Reserved` so that no one uses the SBAS flag for someone else on these messages.

haskell/sbp.cabal.m4
```diff
@@ -1,5 +1,5 @@
 name:                  sbp
-version:               2.3.10
+version:               2.2.7
```
anth-cole: @denniszollo any idea why this version number would have changed and is it important?

anth-cole: @mfine will regen if this causes problems

PR #556, opened 2018-03-20
Cell Modem Status Message

benjaminaltieri:
This new message facilitates the reporting of cell modem signal levels. As other information may be beneficial I have looked at a few standard commands and added fields accordingly:

AT+CSQ -> signal_strength, signal_error_rate
AT+CCID -> sim_id
AT+COPS? -> network operator (if present)
AT+CPIN?/AT+CREG?/AT+CGREG? - sim_status (roughly mapped to no sim/unregistered/registered on network)

As this may not be exhaustive, an unspecified field as a variable length array allows for extra length left at the end for as-of-yet unknown fields.

If an argument is made to only include what I currently need for cell signal reporting, I am totally fine doing that as well.

### DIFF
--- a/spec/yaml/swiftnav/sbp/piksi.yaml
+++ b/spec/yaml/swiftnav/sbp/piksi.yaml
@@ -523,6 +523,38 @@ definitions:
             fill: NetworkUsage
             desc: Usage measurement array
 
+ - MSG_CELL_MODEM_STATUS:
+    id: 0x00BE
+    short_desc: Cell modem information update message
+    desc: |
+      If a cell modem is present on a piksi device, this message
+      will be send periodically to update the host on the status
+      of the modem and its various parameters.
+      AT+CPIN? AT+CCID AT+CSQ AT+CREG? AT+CGREG? AT+COPS?
+    fields:
+      - signal_strength:
+          type: s8
+          units: dBm
+          desc: Received cell signal strength in dBm, zero translates to unknown
+      - signal_error_rate:
+          type: float
+          desc: BER as reported by the modem, zero translates to unknown
+      - sim_id:
+          type: string
+          size: 24
+          desc: The unique ID of sim card installed (ICCID or other similiar)
+      - network_operator:
+          type: string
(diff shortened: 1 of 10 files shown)


spec/yaml/swiftnav/sbp/piksi.yaml
```diff
@@ -523,6 +523,38 @@ definitions:
...
+          size: 24
+          desc: The unique ID of sim card installed (ICCID or other similiar)
+      - network_operator:
+          type: string
+          size: 28
+          desc: Currently reported network operator
+      - sim_status:
+          type: string
+          desc: Current status of sim whether inserted, ready, etc.
+      - unspecified:
```
mfine: We've historically used `reserved` here for these fields not currently in use.

spec/yaml/swiftnav/sbp/piksi.yaml
```diff
@@ -523,6 +523,38 @@ definitions:
...
+      - sim_id:
+          type: string
+          size: 24
+          desc: The unique ID of sim card installed (ICCID or other similiar)
+      - network_operator:
+          type: string
+          size: 28
+          desc: Currently reported network operator
+      - sim_status:
+          type: string
```
mfine: Sending the message here with unsized string and an array fill won't work - they're both greedy.

spec/yaml/swiftnav/sbp/piksi.yaml
benjaminaltieri: Ok cool I'll probably just axe the other fields for now in the interest of getting this message available for the base functionality

spec/yaml/swiftnav/sbp/piksi.yaml
benjaminaltieri: Good catch thanks!

benjaminaltieri:
I've removed the currently unused fields and switched the last floating parameter to `reserved` as suggested.

Generated the products and all tests pass. Final approval anyone?

PR #558, opened 2018-04-11
Add Z to end of timestamp

mfine:
Append the missing Z to the end of the timestamp. This seems to be the optimal way to handle this in python.

/cc @jck @denniszollo


mfine:
While we're in here... what do you guys think of removing the `session-uid` - that seems to have been a miss for the most part.

mfine:
Ping! Going to bring this in today.

PR #559, opened 2018-04-13
Added Measurement Engine basic tracking state message

mbavaro:
It is instrumental to communicate the state about >63 channels (up to 85) in a single SBP message tracking state. Yes, we are at that point 💪


spec/yaml/swiftnav/sbp/gnss.yaml
```diff
@@ -17,6 +17,28 @@ include:
 
 definitions:
 
+ - MeGnssSignal:
```
mfine: You don't need to introduce this new structure, since its identical to the `GnssSignal` structure. I think it's ok to reuse the `GnssSignal` structure with a slightly enhanced interpretation of the `sat` field in your new message structure.

spec/yaml/swiftnav/sbp/tracking.yaml
```diff
@@ -500,6 +500,33 @@ definitions:
           fill: TrackingChannelState
           desc: Signal tracking channel state
 
+ - MeTrackingChannelState:
```
mfine: I would expand these names consistently and avoid the internal abbreviation:

```
MeasurementTrackingChannelState
MSG_MEASUREMENT_TRACKING_STATE
```

mbavaro:
OK, all green. There is no rush, but unless there are objections this might go next week.

mbavaro:
no functionality change, simply updated description of the `sat` field in `GnssSignal`

PR #560, opened 2018-04-28
Add patch version for c source

denniszollo:
Per @switanis request.  Informative only.  We might consider same for pdf.


mfine:
This might break some of the non-C things.

denniszollo:
I only put it in for the c makefile target, so I think it's okay.

mfine:
Ahh, yeah - that would be good.

PR #561, opened 2018-05-04
Add Beidou ephemeris message

Paakkit:
This PR adds Beidou SBP message.
Beidou has 2 tgd parameters in ephemeris, and thus cannot re-use GPS ephemeris message.

https://swift-nav.atlassian.net/projects/EX/issues/EX-45


Paakkit:
First libsbp PR... let's see how it goes.

Paakkit:
Lots of auto generated changes...

javascript/sbp.bundle.js
```diff
@@ -15817,65 +15915,97 @@ for (var i = 0, len = code.length; i < len; ++i) {
 revLookup['-'.charCodeAt(0)] = 62
 revLookup['_'.charCodeAt(0)] = 63
 
-function placeHoldersCount (b64) {
+function getLens (b64) {
```
Paakkit: I wonder what are these changes, starting from here until the end of this file?
They don't seem to be related to the new message.

mbavaro:
Beidou eph significant bits:
image


GPS
image


Galileo
image

spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -445,6 +445,106 @@ definitions:
...
+      The ephemeris message returns a set of satellite orbit
+      parameters that is used to calculate BDS satellite position,
+      velocity, and clock offset. Please see the BeiDou Navigation
+      Satellite System SIS-ICD Version 2.1, Table 5-9 for more details.
+    fields:
+      - common:
+          type: EphemerisCommonContent
+          desc: Values common for all ephemeris types
+      - tgd:
+          type: array
```
mfine: Does the produced ephemeris use an array for these values as well? Or would it be clearer to break out the two values into distinct fields. I don't love using arrays to group items where distinct fields can be used.

mbavaro:
also, there is a reason why GNSS send semi-circles and not radians.. that is not to embed the definition of `PI` in the parameter.. not for this PR because it would have a massive impact on LSNP too, but eventually might make sense to do the same and leave `PI` to the user

Paakkit:
pdf contained question marks.
Doc generation needed to be run twice to resolve Latex links.

Paakkit:
@mbavaro I'll update this PR to use floats for Beidou.
Let's consider a new PR for changing the existing messages.
Also, I'd consider this a good moment to introduce the Galileo message.

Paakkit:
I spoke too soon about the floats - array of floats gives an error.
The array of doubles is in https://github.com/swift-nav/libsbp/blob/master/java/src/com/swiftnav/sbp/SBPMessage.java, but floats are not handled.

Having `float tgd1` and `float tgd2` would work fine.

mbavaro:
I insist... I think adding `getArrayofFloat()` like this:
https://github.com/swift-nav/libsbp/blob/master/java/src/com/swiftnav/sbp/SBPMessage.java#L159-L164
should be possible.

SBP has some contradictions.. it's designed for minimal payloads (so much that messages are constrained to 255 Bytes) but then our ephemeris messages are completely bloated. Giving it a quick glance, full use of `float`s here should reduce the bandwidth of that message to around 67% of the original, and show to the GNSS community that we know what we are doing and aren't just using `double`s to be on the safe side :)

Happy to hear about your Galileo proposition, let me know if you want to take it or I should.

spec/yaml/swiftnav/sbp/observation.yaml
anth-cole: I'd prefer to break away from the array in the SBP message I think - in this instance it'll be much clearer and easier to understand

anth-cole:
so I'll chime in - I absolutely do not want to have the situation where we have BDS ephemeris (and GAL) using floats and the GPS and GAL ones using doubles. Talking to @denniszollo this morning, he suggested if we're trying to optimise bandwidth, is it worth considering going to the raw nav bits? @mbavaro thoughts on this approach? I'm worried about needing to implement an ephemeris decoder everywhere, but it would optimise out throughput.

With the question of floats, I'd either like to convert GPS and GLO messages first and propagate that change through, then add consistent BDS+GAL eph messages, or discuss the idea of adding raw nav messages. I very much agree the current messages are bloated and cause us to have bandwidth spikes.

spec/yaml/swiftnav/sbp/observation.yaml
Paakkit: The array has been removed.

Paakkit:
This PR has been updated with the suggested changes.
In the meanwhile I'll open a new PR for applying the float changes to the existing ephemeris messages.

Paakkit:
Rebased, and now Travis passes again.

anth-cole:
lgtm - please merge

mbavaro:
@Paakkit I merged the Galileo and broke yours, which I now rebased so it should be good to put in when you're back in the office

PR #563, opened 2018-05-08
SBP Galileo ephemeris

mbavaro:
Adds yet another ephemeris message


spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -445,6 +444,107 @@ definitions:
...
+      - inc_dot:
+          type: float
+          units: rad/s
+          desc: Inclination first derivative
+      - af0:
+          type: double
+          units: s
+          desc: Polynomial clock correction coefficient (clock bias)
+      - af1:
+          type: double
```
Paakkit: float

Paakkit: Ah, now I paid attention to your statement - **significantly** less than 23 bit significand
Where to draw the line then... some parameters are around 20 bits.

mbavaro: let's make it consistent with your new GPS ephemeris PR https://github.com/swift-nav/libsbp/pull/564#discussion_r187079630, and move this to `float`

PR #564, opened 2018-05-09
New SBP ephemeris messages with floats

Paakkit:
This PR 
- Adds new GPS, GLO and SBAS ephemeris messages with floats
- Adds new EphemerisCommon with float
- Adds support for array of floats 

Floats for parameters that have significantly less than 23 bit significand.

https://swift-nav.atlassian.net/projects/EX/issues/EX-88


Paakkit:
Rebased, and now Travis passes again.

PR #568, opened 2018-05-17
Windows c again

denniszollo:
Supercedes #479  @silverjam  How does this suite you based upon feedback from #479?

In Common.h we define TOOLCHAIN_ATTR_PACK.  If MSC is used ,it also defines TOOLCHAIN_PRAGMA PACK.  Header files use these macros. 

I think this is the way to go and it is what I had originally.

I really want to get this in, and afterwards then some tests on appveyor and a fleshed out example that Stefan has written.

### DIFF
--- a/c/include/libsbp/common.h
+++ b/c/include/libsbp/common.h
@@ -22,6 +22,16 @@
 #ifndef COMMON_INT_TYPES
 #define COMMON_INT_TYPES
 
+/* Determine Toolchain */
+
+#if defined _MSC_VER
+#define TOOLCHAIN_ATTR_PACKED    /* Intentionally empty */
+#define TOOLCHAIN_PRAGMA_PACK
+#else
+#define TOOLCHAIN_ATTR_PACKED  __attribute__((packed))
+#undef TOOLCHAIN_PRAGMA_PACK 
+#endif
+
 /** \defgroup common_inttypes Integer types
  * Specified-width integer type definitions for shorter and nicer code.
  *
--- a/c/include/libsbp/common.h
+++ b/c/include/libsbp/common.h
@@ -22,15 +22,30 @@
 #ifndef COMMON_INT_TYPES
 #define COMMON_INT_TYPES
 
-/* Determine Toolchain */
+/* Set packing based upon toolchain */
 
-#if defined _MSC_VER
-#define TOOLCHAIN_ATTR_PACKED    /* Intentionally empty */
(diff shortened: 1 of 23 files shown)


c/include/libsbp/common.h
```diff
@@ -22,6 +22,16 @@
 #ifndef COMMON_INT_TYPES
 #define COMMON_INT_TYPES
 
+/* Determine Toolchain */
+
```
silverjam: This is what I was thinking of:

```c
#if defined(_MSC_VER)

#define SBP_PACK_START \
    __pragma(pack(1));
#define SBP_PACK_END \
    __pragma(pack());

#define SBP_PACKED /* Intentionally empty */

#elif defined(__GNUC__) || defined(__clang__)

#define SBP_PACK_START
#define SBP_PACK_END

#define SBP_PACKED __attribute__((packed))

#else

#if !defined(SBP_PACK_START)
#define SBP_PACK_START
#warning Unknown compiler, please override SBP_PACK_START
#endif

#if !defined(SBP_PACK_END)
#define SBP_PACK_END
#warning Unknown compiler, please override SBP_PACK_END
#endif

#if !defined(SBP_PACKED)
#define SBP_PACKED __attribute__((packed))
#warning Unknown compiler, please override SBP_PACKED
#endif

#endif
```

silverjam: So SBP_PACK_START and SBP_PACK_END would make this look a little cleaner:

```c
#ifdef TOOLCHAIN_PRAGMA_PACK
#pragma pack(1) // or #pragma pack() for the end of the file
#endif
```

Also, using `SBP_PACKED` so it's a little shorter (hopefully less aesthetically offensive) and added the `SBP_` prefix to make it clear it's a macro we're defining/using and will prevent it from colliding with other names (however unlikely that may be) ... we could even make it lowercase `__sbp_attribute_packed__` to make it look more like `__attribute__((packed))`, if we're so inclined.  Anyway, regardless of aesthetics, if we can make it more explicit what compilers we're supporting with the above approach, then I'm fine with this.

denniszollo:
Stefan's feedback on which I am updating for the last time:
  - put GNU check first in common as it is default
  - don't use SBP as prefix for macros as it implies it is something about Swift, when in reality it is related to the complier
  - error if compiler unknown and macros not overridden since this is really an error condition that people should fix sooner rather than later.  This is the conservative approach.

@switanis @ebethon

PR #570, opened 2018-05-22
Add note on carrier phase observation alignment

ljbade:
Please check that this matches what we do in firmware. If we decide to add any signal alignments in the future for Galileo we should update this note.

Sorry for the whitespace changes but my editor automatically strips extra whitespace.


spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -53,7 +53,11 @@ definitions:
     desc: |
       Pseudorange and carrier phase observation for a satellite being
       tracked. The observations are interoperable with 3rd party
-      receivers and conform with typical RTCMv3 GNSS observations.
+      receivers and conform with typical RTCM 3.1 GNSS observations.
+
+      All carrier phase observations are unaligned and output the raw phase
```
anth-cole: Do we want to add this note? It means we'll not comply as we currently do both aligned and unaligned obs

Coming out of the piksi our obs are unaligned (but then we can ask, aligned to what?) but no adjustment is added. Decoding RTCM however, specifically the 1004 message, will have either aligned or unaligned obs depending on the sending reciever. If we add this paragraph, we'll need to also remove the 1/4 cycle adjustment whenever it's been applied to get back to 'unaligned' obs. I think adding this just opens a can of worms on what the format contains.

ljbade: Sure, but how is anyone supposed to use our obs message if we aren't clear on what we are doing? Particularly if we aren't consistent. At a minimum we should state we comply with RTCM 3.1 but not MSM (the current statement on "typical RTCMv3 GNSS observations" isn't accurate now there is MSM too).

Maybe we should wait until after we implement alignment in the firmware to add this note, then we can fix up our RTCM 3.1 decoding to match.

I'll add clarification that I'm referring to alignment to the RINEX 3 reference signal.

Anyway it's clear that our current RINEX 3 convertor doesn't follow the specification which will cause problems if people expect it to work nicely with tools that expect aligned data.

PR #578, opened 2018-06-14
python: pin httpretty to fix tests on travis

jck:


mfine:
:shipit: if it passes

PR #583, opened 2018-08-01
Put memory packing outside the COMMON_INT_TYPES macro

mbavaro:
When combined with LSN/common.h the current arrangement causes quite a bit of problems..

### DIFF
--- a/c/include/libsbp/common.h
+++ b/c/include/libsbp/common.h
@@ -21,32 +21,6 @@
 /* Should match guard in libswiftnav/common.h */
 #ifndef COMMON_INT_TYPES
 #define COMMON_INT_TYPES
-
-/* Set packing based upon toolchain */
-
-#if defined(__GNUC__) || defined(__clang__)
-
-#define SBP_PACK_START /* Intentionally empty */
-#define SBP_PACK_END /* Intentionally empty */
-
-#define SBP_ATTR_PACKED __attribute__((packed))
-
-#elif defined(_MSC_VER)
-
-#define SBP_PACK_START __pragma(pack(1));
-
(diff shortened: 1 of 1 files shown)


mfine:
Man, I'm not good with this stuff - I think @denniszollo had something around this https://github.com/swift-nav/libsbp/pull/479 back in the day.

mbavaro:
When compiling code that has dependencies with both `libswiftnav` and `libsbp`, one bumps on this 

https://github.com/swift-nav/libswiftnav/blob/master/include/libswiftnav/common.h#L30

which defines our int types (`s8`, `u16`, ...) and just those. 
However, the `#ifdef` in libsbp holds these packing pragmas too..

https://github.com/swift-nav/libsbp/blob/master/c/include/libsbp/common.h#L27-L48

So depending on the include order (if the LSN `common.h` is included first) those aren't seen by the pre-compiler and so a bunch of declarations like 
https://github.com/swift-nav/libsbp/blob/master/c/include/libsbp/gnss.h#L28
and 
https://github.com/swift-nav/libsbp/blob/master/c/include/libsbp/gnss.h#L48
throw errors...
I don't see why those pre-processor macros should go inside `COMMON_INT_TYPES`.. I must be missing something?

silverjam:
@mbavaro I wanted to double check the actual error, but given that, this change looks correct, it definitely makes sense that it shouldn't be inside the `COMMON_INT_TYPES` guard.

denniszollo:
Don't we need another preprocessor guard so these macros aren't redefined each time the file is included then ?

denniszollo:
My bad team.  Glad you sorted it out.

PR #585, opened 2018-08-04
Changed read() api from `size_t` to `ssize_t`

mbavaro:
It's a long story.. but right now sbp_process() does not have a way to distinguish a failed read() from a timed-out read(). In order to align the function prototype with the common Linux `man read(2)`, the prototype is changed to signed int (rather than unsigned) and a SBP error is thrown when the `read()` fails.


gsmcmullin:
I agree signed values should be returned by read/write and errors should be propagated up.

It's very difficult to separate the signal from the noise in the diff. I seriously question the 37,865 lines removed, and the clang-format makes it harder to read.  What happened to Linux style?

mbavaro:
Thanks Gareth, I agree across the board with your comments. 
I also just intended to change the API, but then tests started failing. I think some of them are automatically generated, but not all. I cleaned the tests folder and that is why so many lines of codes went.  I will split this PR in 2 parts:
1. minimal API change
2. add a clang_format for the C library.. did you mean this one https://github.com/torvalds/linux/blob/master/.clang-format ?

mbavaro:
@gsmcmullin I believe this is the minimum required to change the API to signed.

mfine:
@mbavaro if you change the passed in values to negative, should you not also update the logic in `sbp_process` to handle negative returned values instead of just adding it to `s->n_read` which is common?

mbavaro:
OK, sorry I was in a conf call and did not update the issue description.. I will now add the commit for .2

denniszollo:
I think we also need to update example here:
https://github.com/swift-nav/libsbp/blob/master/c/example/example.c#L69

mbavaro:
I think it depends on compiler settings.. it does not seem to upset the example;
https://github.com/swift-nav/libsbp/blob/master/c/example/example.c#L74
that we convert from `signed` 
https://www.sigrok.org/api/libserialport/0.1.0/a00008.html#ga4c6aa5ea99d787e9336bab9d977463e4
to `u32` before returning...

another approach would be to update the API to the new and deprecate explicitly the current https://github.com/swift-nav/libsbp/blob/master/c/example/example.c#L69 
as it is done for messages..

PR #592, opened 2018-08-27
Build `libsbp.a` instead of `libsbp-static.a`

mbavaro:
This works better for me in a CMake driven build system where specifying `cmake ../ -DCMAKE_SYSTEM_NAME=Linux -DBUILD_SHARED_LIBS=OFF` naturally links statically LIBSBP.
Without this, the linker needs to know the libsbp static library is called something else (`libsbp-static.a`). I am not sure why that would be as the name wouldn't conflict with the dynamically linked one..

### DIFF
--- a/c/src/CMakeLists.txt
+++ b/c/src/CMakeLists.txt
@@ -11,14 +11,12 @@ set(libsbp_SRCS
   sbp.c
 )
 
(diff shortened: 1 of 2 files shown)


mbavaro:
Obsoleted by https://github.com/swift-nav/libsbp/pull/594 I think

PR #594, opened 2018-09-19
Get rid of separate target for static library and stop fiddling BUILD…

martin-swift:
…_SHARED_LIBS flag.


benjaminaltieri:
Trying understand if this would be breaking any downstream projects, a quick [github search](https://github.com/swift-nav/libsbp/search?q=BUILD_SHARED_LIBS&unscoped_q=BUILD_SHARED_LIBS) definitely shows that if nothing else we should document this change in behavior.

nsirola:
At least firmware builds with minimal changes: https://github.com/swift-nav/piksi_firmware_private/pull/2350

martin-swift:
I want to merge this; @mbavaro has given his approval - last call for objections!

benjaminaltieri:
go for it!

PR #598, opened 2018-09-21
Updated INS Status message

jkretzmer:
This PR updates the definition of the INS Status message based on discussions with CRL.


jkretzmer:
cleaning up against current master.

PR #610, opened 2018-10-08
Add cmake toolchain support for cortex-m7

yiiiyang2010:
- set fpu as -mfpu=fpv5-d16 for supporting double precision floating point for fpu v5.


c/cmake/Toolchain-gcc-arm-embedded.cmake
```diff
@@ -41,21 +41,26 @@ message(STATUS "Cross-compiling with the gcc-arm-embedded toolchain")
...
     "${CMAKE_C_FLAGS}"
     "-mcpu=cortex-m4 -march=armv7e-m -mthumb"
     "-mfloat-abi=hard -mfpu=fpv4-sp-d16"
   )
+  
+elseif (CMAKE_SYSTEM_PROCESSOR STREQUAL "cortex-m7")
 
+  set(CMAKE_C_FLAGS
+    "${CMAKE_C_FLAGS}"
+    "-mcpu=cortex-m7 -march=armv7e-m -mthumb"
```
mbavaro: `-march=armv7e-m` isn't a flag in the CubeMX `Makefile`.. should be enforced by `mcpu` https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html

PR #622, opened 2018-11-21
Msg csac telem

denniszollo:


c/include/libsbp/system.h
```diff
@@ -82,12 +82,38 @@ typedef struct SBP_ATTR_PACKED {
...
+/** Experimental telemetry message
+ *
+ * The CSAC telemetry message has an implementation defined telemetry string
+ * from a device. It is not produced or available on general Swift Products.
+ * It is intended to be a low rate message for status purposes.
+ */
+#define SBP_MSG_CSAC_TELEMETRY        0xFF04
+typedef struct SBP_ATTR_PACKED {
+  u8 id;           /**< Index representing the type of telemetry in use.  It is implemention defined. */
+  char telemetry[0]; /**< Comma separated list of values as defined by the index */
```
cbeighley: @dzollo Are all the telemetry values 8 bit values?

cbeighley: Oh, I see, comma separated.

cbeighley: This feels unnecessarily loosely typed to me - don't we already know all the telemetry values that are produced? Can we not make this much more strongly typed?

c/include/libsbp/system.h
```diff
@@ -82,12 +82,38 @@ typedef struct SBP_ATTR_PACKED {
...
+} msg_csac_telemetry_t;
+
+
+/** Experimental telemetry message labels
+ *
+ * The CSAC telemetry message provides labels for each member of the string
+ * produced by MSG_CSAC_TELEMETRY. It should be provided by a device at a lower
+ * rate than the MSG_CSAC_TELEMETRY.
+ */
+#define SBP_MSG_CSAC_TELEMETRY_LABELS 0xFF05
```
cbeighley: Do we need a message for the labels? Don't we already know what data comes out in the telemetry? If it ever changes, we could just rev the `SBP_MSG_CSAC_TELEMETRY` message.

cbeighley:
@dzollo I don't think changing this is strictly necessary if you don't feel like we have the time, but ideally I think this would be much more strongly typed struct of specific values rather than a comma separated string.

c/include/libsbp/system.h
denniszollo: we could, but this is the native data from csac and it seemed easier to just leave it as is.

c/include/libsbp/system.h
denniszollo: The labels come from the CSAC.

denniszollo:
We discussed and decided these messages are okay since they are as close as possible to what is actually on the wire from the timing device and the published timing device serial API.  This telemetry concept is very extensible for future timing devices and/or timing device firmware updates.  If we require a tighter integration feedback loop between our receiver and timing module in future I agree that the string version of the message can be deprecated and replaced with a new strongly typed message.

denniszollo:
It's a good thing I didn't assume anything about the strings either as I see sometimes the device we are using replaces something that is normally a number with a string.

PR #634, opened 2019-01-16
Front end gain message

denniszollo:
This PR adds a new SBP message to provide the gain values in RF frontends from Swift Receivers.  Its goal is to provide human friendly messages that can be used to assess receiver health and purposely divorces itself from hardware implementation.  If a future frontend doesn't have an AGC for one of the stages, we have a way to encode it as invalid or missing.

image
@ericsenant @mbavaro @nsirola @fnoble


denniszollo:
I think I didn't generate all languages after updating description.  I will fix.

switanis:
Max can be 255 not 256 I think.
The usable (valid) range will be 0 to 100. I mean no any scaling. Please confirm.
Otherwise good. I like it.

denniszollo:
@sfefan. I used a signed 8 bit integer and noted that the max value of 127 would mean channel is missing and a negative value would mean some kind of error.  valid range is 0 to 100.

switanis:
Sounds good. Thx.

PR #637, opened 2019-01-24
Compile c library with -fPIC.

martin-swift:
As I was moving things around in the repo refactor, I got an error related to this.  Seems like a good thing to have in a library and I am not aware of any downsides.

### DIFF
--- a/c/src/CMakeLists.txt
+++ b/c/src/CMakeLists.txt
@@ -30,9 +30,11 @@ set(libsbp_HEADERS
 set(libsbp_SRCS
   edc.c
   sbp.c
-)
+  )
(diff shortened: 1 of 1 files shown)


silverjam:
@martin-swift what was the error that you got?

martin-swift:
@silverjam 
```
/usr/bin/ld: sbp/src/libsbp-build/src/libsbp.a(sbp.c.o): relocation R_X86_64_PC32 against symbol `sbp_clear_callbacks' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
```

PR #638, opened 2019-01-25
Allow multiple messages be sent at once [ESD-1038]

silverjam:


python/sbp/client/framer.py
```diff
@@ -128,7 +128,7 @@ def _receive(self):
             pass
         return msg
 
-    def __call__(self, msg, **metadata):
+    def __call__(self, *msgs, **metadata):
```
tn-exafore: Please update the docstring, too.

I presume that metadata will be shared between the messages sent in one go if it ever gets implemented?

denniszollo: Is there a guaranteed ordering of the messages?

silverjam: Messages will be passed to the transport in the order that they appear in the list, beyond that I don't think this changes anything with regards to ordering guarantees (that we had prior).

silverjam:
@denniszollo This isn't really an interface change since we're changing from `send(msg, **metadata)` to `send(*msgs, **metadata)` which allows any call in the form `send(msg)` to remain unchanged.  To send multiple message the invocation looks like this `send(msg1, msg2)` or if you have a list of messages then it's: `send(*msgs)`.

python/sbp/client/framer.py
silverjam: @tn-exafore I was thinking we should just remove the metadata parameter... @denniszollo any comments on that?  Can I just remove `metadata`?

silverjam: Removing metadata was too much of a rats nest, seems some of this data is passed to logging objects at some point.

python/sbp/client/handler.py
```diff
@@ -250,9 +250,9 @@ def cb(msg, **metadata):
         event.wait(timeout)
         self.remove_callback(cb, msg_type)
 
-    def __call__(self, msg, **metadata):
+    def __call__(self, *msg, **metadata):
```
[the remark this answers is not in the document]
silverjam: fixed

python/sbp/client/framer.py
silverjam: (Also, fixed docs)

python/setup.py
```diff
@@ -31,18 +31,47 @@
...
+    if tag.endswith('-dirty'):
+        kw_params['dirty'] = True
+    return meta(version, **kw_params)
+
 
 setup(name='sbp',
       description='Python bindings for Swift Binary Protocol',
       long_description=readme,
-      use_scm_version = {'root': '..', 'relative_to': __file__},
+      use_scm_version={'root': sbp_root(), 'parse': do_parse},
```
tn-exafore: why the manual parsing? Is this a temporary thing or meant to be merged?

silverjam: Sorry, I'll pull this into a separate PR. This allows libsbp development versions to be installed with pip.  Without this, things would crash somewhere inside setuptools_scm.

tn-exafore: ok, although it's a bit of a hack, it's a useful feature. But cleaner in a separate PR for sure.

PR #642, opened 2019-02-05
Remove obsolete examples and SBP_MSG_TWEET [EX-476]

tn-exafore:


silverjam:
Super late to the party here... but technically we shouldn't remove messages... right?  I think `SBP_MSG_TWEET` might be an exception here, but future "obsoleted" or removed messages should remain in the yaml with their original message IDs but be renamed to reflect their status.  E.g. `OBSOLETE_SBP_MSG_TWEET`.

PR #645, opened 2019-02-09
Remove dependency on setuptools_scm [ESD-1038]

silverjam:
+ Remove setuptools_scm since it prevents us from installing from a git URL via pip, adapt a new setup.py from enthought's traitsui [package](https://github.com/swift-nav/traitsui/blob/swift-2019.01/setup.py).

+ Add `default.nix` for running `make all`


HOWTO.md
```diff
@@ -88,6 +88,15 @@ Ubuntu 16.04.
    libsbp version number in the C, Python, JavaScript, and LaTeX developer
    documentation are consistent.
 
+   - C: update `c/include/version.h` with new version
+
+   - Python: update `python/setup.py` with new version, change
+     `IS_RELEASED = false` to `IS_RELEASED = true`.
+
+   - JavaScript: update `package.json` and `package-lock.json`.
+
```
denniszollo: All of these version files this should happen in makefile automatically after a "make all", should we mention as such?

python/setup.py
```diff
@@ -29,28 +40,130 @@
...
-      classifiers=CLASSIFIERS,
-      packages=PACKAGES,
-      platforms=PLATFORMS,
-      install_requires=INSTALL_REQUIRES,
-      tests_require=TEST_REQUIRES,
-      use_2to3=False,
-      zip_safe=False)
+MAJOR = 2
+MINOR = 4
+MICRO = 8
```
denniszollo: Will these strings need to be updated by hand each release ^^ ?

python/setup.py
```diff
@@ -29,28 +40,130 @@
...
+        # LANGUAGE is used on win32
+        env['LANGUAGE'] = 'C'
+        env['LANG'] = 'C'
+        env['LC_ALL'] = 'C'
+        out = subprocess.Popen(
+            cmd, stdout=subprocess.PIPE, env=env,
+        ).communicate()[0]
+        return out
+    try:
+        out = _minimal_ext_cmd(['git', 'describe', '--tags'])
```
denniszollo: what will happen if someone pip installs sbp WITHOUT git in their path?

HOWTO.md
benjaminaltieri: I'm seeing `c/include/version.h`, `package.json`, and `sbp.cabal.m4` update automatically - whereas `package-lock.json` does not

python/setup.py
benjaminaltieri: looks like the answer is yes

benjaminaltieri:
I'm also a fan of unlocking this for developer ease of use, the current issues look to be centered around the downstream dependency interactions (i.e. `(piksi_tools) bash-3.2$ python setup.py install
error in piksi_tools setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Invalid requirement, parse error at "'+https:/'"`) and the added manual process of updating `python/setup.py` which just need to be properly specified in docs

python/setup.py
silverjam: It should resolve to "unknown", see `if match is None` below.

HOWTO.md
silverjam: ~I'm confused... how do these update automatically?~ Never mind... looks like the sbp generator for C updates this.  I suppose it would make sense to update the Python version too?

silverjam:
@benjaminaltieri that is/was a problem with `piksi_tools` -- not libsbp itself, see https://github.com/swift-nav/piksi_tools/pull/977

python/setup.py
silverjam: I've fixed so that python version will be automatically generated along with the other languages (that already have version info automatically generated).

python/setup.py
silverjam: Thus, if it was an unreleased tarball of libsbp, then it would show as "v2.4.9.dev0+gUnknown" as it's version.

benjaminaltieri:
awesome, this is great. :shipit:

PR #646, opened 2019-02-12
[ME-166] Add 16 bit correlator message

mbavaro:
A new message with 16-bit complex correlator outputs


denniszollo:
Looks good to me, but please split up the YAML change from the generated code change into two commits.  It makes it easier to collaborate.

mbavaro:
Is this a new practice?
@dzollo Please review again.

PR #651, opened 2019-02-14
Use twine/nix to push to PyPI, auto-mangle IS_RELEASED [ESD-1038]

silverjam:
+ Add a target for just deploying Python

+ Automatically mangle the `IS_RELEASED` flag in setup.py

+ Add nix spec for running a deployment

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -88,9 +88,6 @@ Ubuntu 16.04.
    libsbp version number in the C, Python, JavaScript, and LaTeX developer
    documentation are consistent.
 
-   - Python: In `python/setup.py` change `IS_RELEASED = False` to
-     `IS_RELEASED = True`.
-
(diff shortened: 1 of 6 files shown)


denniszollo:
Code looks good, but can you explain the human workflow for a release / deploy to help me understand?  

Will the wheel work for mac and windows users too or do those users get a tarball of the python source?

silverjam:
@denniszollo The libsbp wheels are source based (no compiled libs), so they'll work on any platform.  Release flow should be fairly similar.  To release just Python it would be:

```
make PYPI_USERNAME=xyz PYPI_PASSWORD=password dist-python 
```

Or, all together:

```
make PYPI_USERNAME=xyz PYPI_PASSWORD=password dist
```

silverjam:
@dzollo if you have concerns I'll address in a follow-up PR

PR #647, opened 2019-02-13
[WIP] win32 build for GTT

pmiettinen:
Currently GTT has SBP module as statically copied files, which is not good. Some investigations for providing SBP as actual library on win32 platform.

Initial build produces only static `sbp.lib`. See if `sbp.dll` should be build also.

How to build in 'Native Tools Command Prompt for VS 2017':

Short:
```
md build
cd build
cmake ..
msbuild libsbp.sln /p:Configuration="Release" 
```

Details:
```
c:\xoodi\github\sn\libsettings\libsbp\c>md build

c:\xoodi\github\sn\libsettings\libsbp\c>cd build

c:\xoodi\github\sn\libsettings\libsbp\c\build>cmake ..
-- Building for: Visual Studio 15 2017
-- Selecting Windows SDK version 10.0.17763.0 to target Windows 10.0.17134.
-- The C compiler identification is MSVC 19.16.27026.1
-- The CXX compiler identification is MSVC 19.16.27026.1
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/BuildTools/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x86/cl.exe
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/BuildTools/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x86/cl.exe -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/BuildTools/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x86/cl.exe
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/BuildTools/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x86/cl.exe -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- libsbp version: v2.4.9-dirty
-- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE)
CMake Warning at docs/CMakeLists.txt:7 (message):
  Doxygen not found, the documentation will not be built.


-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE)
[...]

PR #652, opened 2019-02-14
Updated CMake files

jbangelo:
Cleaned out apparently unused CMake files and variables.

Switched to setting target properties for C standard and warnings.

Exported targets to the build tree to make building as an external project easier.


c/src/CMakeLists.txt
```diff
@@ -31,10 +31,18 @@ set(libsbp_SRCS
   edc.c
   sbp.c
   )
-set(LIBSBP_CFLAGS "-fpic" CACHE STRING "Compile flags for libsbp.")
```
silverjam: Libsbp is a library that our customers definitely use, we need to make sure we're choices that reflect that.  The cflags var here we designed to allow customers to override these compile flags if needed.

jbangelo: Ok, the old use wasn't working when I added more than one flag to the string. I'll figure out the right want to do this.

Would the removed CMake files potentially be used by customers?

c/CMakeLists.txt
```diff
@@ -1,57 +1,8 @@
...
-  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
-  RESULT_VARIABLE GIT_VERSION_FOUND
-  ERROR_QUIET
-  OUTPUT_VARIABLE GIT_VERSION
-  OUTPUT_STRIP_TRAILING_WHITESPACE
-)
-if (GIT_VERSION_FOUND)
-  set(VERSION "unknown")
-else (GIT_VERSION_FOUND)
-  set(VERSION ${GIT_VERSION})
```
silverjam: Do we not need this property anymore?  CMake docs specify that's it's used when generated the library? https://cmake.org/cmake/help/v3.7/prop_tgt/VERSION.html

c/src/CMakeLists.txt
silverjam: Not sure, can we investigate the `git blame` on these files to see why they were originally introduced?

c/CMakeLists.txt
```diff
@@ -1,57 +1,8 @@
-cmake_minimum_required(VERSION 2.8.7)
+cmake_minimum_required(VERSION 3.2)
 project(libsbp)
 
-# Setup flags for Code Coverage build mode
-set(CMAKE_CXX_FLAGS_COVERAGE "${CMAKE_CXX_FLAGS_DEBUG} --coverage" CACHE STRING
```
silverjam: Can we check `git blame` on these chunks of cmake?  Is there potentially something downstream that would use this?

c/CMakeLists.txt
jbangelo: The original code wasn't ever setting the `VERSION` property, it was just making a variable called `VERSION`.

If we want to have that property set I can certainly add it in, but I'm not sure our way of getting the version number will work in most cases. It seems like CMake usually expects versions to be exactly in the format of `0.0.0.0` with no additional text before or after. Our git tags at the very least have a "v" character at the beginning.

c/src/CMakeLists.txt
jbangelo: Most of the `.cmake` files appear to have been added in this commit: https://github.com/swift-nav/libsbp/commit/062a3e3ec351303023e556a95ee880b6e11e2479 but they weren't used in that commit, and haven't been touched since they were added. The toolchain file has been updated recently so I'll bring it back in.

c/CMakeLists.txt
jbangelo: `git blame` shows all of these removed lines were added in commit https://github.com/swift-nav/libsbp/commit/062a3e3ec351303023e556a95ee880b6e11e2479 (the same one the removed `.cmake` files were added) and they haven't been modified since. TBH I'm guessing this was all copy pasta from somewhere else.

With how CMake works setting these variables will only have an effect on the stuff defined in this file and any directories it adds via `add_subdirectories()`. These variables aren't used anywhere in the libsbp repo, so removing them shouldn't effect the library itself. Anything downstream that does a `add_subdirectory(libsbp/c/)` also shouldn't see any effects of removing them due to CMake's scoping rules (unless they are doing something really strange with their CMake).

c/CMakeLists.txt
silverjam: Ok, let's not worry about it for now.

c/src/CMakeLists.txt
silverjam: Ok, SGTM.

c/CMakeLists.txt
silverjam: Sounds good, thanks.

c/src/CMakeLists.txt
```diff
@@ -31,10 +31,27 @@ set(libsbp_SRCS
...
+
+set(LIBSBP_CFLAGS "-fpic -Wall -Werror" CACHE STRING "Compile flags for libsbp.")
```
pmiettinen: Please check this conversation

https://github.com/swift-nav/libsbp/pull/647#discussion_r256539614
and
https://github.com/swift-nav/libsbp/pull/647#discussion_r256899431

c/src/CMakeLists.txt
```diff
@@ -31,10 +31,27 @@ set(libsbp_SRCS
...
 
 add_library(sbp ${libsbp_SRCS})
-target_compile_options(sbp PRIVATE "${LIBSBP_CFLAGS}")
+
+if(MSVC)
+    separate_arguments(LIBSBP_COMPILER_OPTIONS WINDOWS_COMMAND ${LIBSBP_CFLAGS})
```
pmiettinen: This doesn't compile with MSVC:
`cl : Command line error D8021: invalid numeric argument '/Werror'`

Please check https://github.com/swift-nav/libsbp/pull/647

c/src/CMakeLists.txt
jbangelo: Amazing! Thanks for the info.

c/src/CMakeLists.txt
jbangelo: Good catch! I'll add a special case for MSVC. I think the intention is for the consumers of the library that need/want specific options to set this string themselves, but it would be good to make sure the defaults work out of the box for the common compilers at least :)

c/CMakeLists.txt
```diff
@@ -1,56 +1,19 @@
...
-    CMAKE_EXE_LINKER_FLAGS_COVERAGE
-    CMAKE_SHARED_LINKER_FLAGS_COVERAGE )
-# Update the documentation string of CMAKE_BUILD_TYPE for GUIs
-set(CMAKE_BUILD_TYPE "${CMAKE_BUILD_TYPE}" CACHE STRING
-  "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel Coverage."
-   FORCE )
-
-# Set project version using Git tag and hash.
-execute_process(
-  COMMAND git describe --dirty --tags --always
```
martin-swift: so we've totally abandoned automated versioning?

c/test/CMakeLists.txt
```diff
@@ -21,6 +21,9 @@ else (CMAKE_CROSSCOMPILING)
     add_executable(test_libsbp check_main.c check_edc.c check_sbp.c ${generated_c_sources})
 
     target_link_libraries(test_libsbp ${TEST_LIBS})
+    set_property(TARGET test_libsbp PROPERTY C_STANDARD 99)
```
martin-swift: nothing wrong with this, per se, but for maximum consistency, can we pick set_target_properties or set_properties(TARGET ...)?

c/CMakeLists.txt
jbangelo: If this was being used, it can be put back in. But I couldn't see a way the `VERSION` variable could be used. It's not set to the parent scope, and it's not being used to set the `VERSION` property. It appeared to be only used to print out the version string while `cmake` was running.

c/test/CMakeLists.txt
jbangelo: Fair enough, `set_target_properties` it is.

c/CMakeLists.txt
martin-swift: Cool - I am all for terminated unused and unneeded code with extreme prejudice.  Just want to make sure we don't screw up someone's release process or something, since libsbp is submoduled in a lot of places.

kmdade:
awesome.

PR #654, opened 2019-02-26
added virtualenv instructions for the python bindings for libsbp

ekalosak:


python/README.rst
```diff
@@ -20,6 +20,23 @@ Install package from pypi::
 
   $ sudo pip install sbp
 
+Setup using a virtual environment
+-----
+
+Install the `virtualenv`::
```
pmiettinen: Rather create and activate?

ekalosak: good call

PR #655, opened 2019-02-27
[WIP] Add JIT for crc, build into existing buffer to avoid copies

silverjam:
+ Changed the behavior of Framer to stop when "iterator" (the remote connection or file) is no longer producing data.  This seems to be a more standard behavior and it allows us to create an sbp2json utility that will stop when no more data is available.  We can potentially drop these changes if we can't easily figure out how to harmonize this with the rest of the code base.

+ Framer now use an "into_buffer" function to avoid copying data

+ Use numba to accellerate the crc function, running this shows the following speed-ups (first number is with JIT, second is without):

```
$ ./bench/crc.py
Start: building random data...
End: building random data.
1.06078910828
19.5082828999
```

TODO:
+ Need to abstract `build_payload` and the creation of the `StreamPayload` object, currently there's a lot of unnecessary duplicated code between generated objects.

+ Remove http support and fix unit tests.  HTTP related tests are currently failing and I suspect many of the unit tests need to updated to account for the new behavior of the Framer iterator.


silverjam:
Split into #656 and #658

PR #663, opened 2019-03-14
Add option to omit metadata [ESD-1170]

pmiettinen:
https://github.com/swift-nav/piksi_tools/pull/989


denniszollo:
Looks good, but as far as I'm concerned without the metadata we should just be recording straight binary, not json.

PR #665, opened 2019-03-18
Add option for explicit start [ESD-1166]

pmiettinen:
If receive thread is started before all the callbacks are added and sink is ready it is highly probable that some of the messages at start are lost.


python/sbp/client/handler.py
```diff
@@ -31,9 +31,12 @@ class Handler(object):
     ----------
     source : Iterable of tuple(SBP message, {'time':'ISO 8601 str'})
       Stream of SBP messages
+    explicit_start : Boolean
```
silverjam: Maybe `autostart` and default to true?

pmiettinen: Sure, sounds better.

PR #666, opened 2019-03-18
Add SBP wrapper for NMEA GSV sentence

silverjam:


silverjam:
real support for this has been merged

PR #656, opened 2019-02-27
Python: Numba based SBP parsing [ESD-1154]

silverjam:
This is paired with the following `piksi_tools` PR: https://github.com/swift-nav/piksi_tools/pull/985/

## Design notes
Numba based parsing for SBP classes.  Adds various `numba` accelerated `get_*` functions used to parse SBP data.

## Parsing floats
Floats are parsed using a module generated by another library, `cffi` -- https://cffi.readthedocs.io/en/latest/ -- this is used to generate a Python native module that can parse and re-assemble float and double values from a stream of bytes.  This was done because `numba` was unable to generate code that could do the necessary bit banging.

To generate this module, run the following:
```
source <my/libsbp/virtualenv/bin/activate>
cd python/sbp/jit
python parse_float.py
```

This will generate `parse_float_c.so` (and `parse_float_c.c`) which will be loadable by Python.

## Next steps
The next step from here would be to make the various `MsgWhatever.parse_members` methods able to be JIT generated.  This would require some more clever use of `Tuple(( ... ))` for the return values of the `.parse_member` method.

## Testing results
- Without JIT’d `unpack`/`parse` (with JIT’d `crc` routine though):
   ```
   $ NOJIT=y time python ./sbp2json.py <sbp16m.bin  >/dev/null
      247.10 real       227.52 user         1.40 sys
   ```

- With JIT’d `crc`/`unpack`/`parse`:
   ```
   $ time python ./sbp2json.py <sbp16m.bin  >/dev/null

   real    0m11.632s 
   user    0m10.813s
   sys     0m0.338s
   ```

- Haskell `sbp2json` for reference:
   ```
   $ time sbp2json <sbp16m.bin  >/dev/null

   real    0m6.563s
   user    0m8.168s
   sys     0m0.667s
   ```

PR #667, opened 2019-03-19
Automated Travis benchmark [ESD-1158]

pmiettinen:
This is done on top of https://github.com/swift-nav/libsbp/pull/656

And paired with https://github.com/swift-nav/piksi_tools/pull/985


pmiettinen:
Test comment

pmiettinen:
Test comment

pmiettinen:
`travis.yml` based approach seems cumbersome, investigate more tox related approach.

pmiettinen:
Working locally. Some tune up TODO for docker cooperation.

pmiettinen:
Based on the runs this far, the performance ratio threshold could be set somewhere around 1.8.

python/sbp/jit/parse_float.py
```diff
@@ -0,0 +1,49 @@
+import cffi
```
silverjam: Prior to merging this we need to figure out how to get this module to generate itself automatically (for development), then build a compiled version for a PyPI wheel (this is optional, since it might be worthy of a whole separate PR to figure out to build and distribute this).

pmiettinen: How about a first phase remedy:

```
from sbp.jit import parse_float
from sbp.jit import parse_float_c
```

Also note the additions to `tox.ini` so another possibility would be to add the requirement to run `make python` before usage but that would probably go unnoticed in quite a lot of cases.

silverjam: @pmiettinen that might work, as long as the import succeeds, hopefully this doesn't require the user to have write access to wherever `sbp.jit.parse_float` is installed?

pmiettinen: Pretty sure the write access is required but I don't think this it matters at this phase as this solution is aimed only for development and the files shall reside under the repository? Do this now and later a better solution in a PR adding the `sbp.jit` package into the release PyPi package?

silverjam: Ok

PR #670, opened 2019-03-26
EX-569: new message for SV azimuths/elevations

nsirola:
Add a new message to enable NMEA GSV sentence generation from SBP data.

image


c/include/libsbp/observation.h
```diff
@@ -1046,6 +1046,28 @@ typedef struct SBP_ATTR_PACKED {
 } msg_glo_biases_t;
 
 
+/** Satellite azimuth and elevation.
+ *
+* Satellite azimuth and elevation.
+ */
+typedef struct SBP_ATTR_PACKED {
+  sbp_gnss_signal_t sid;    /**< GNSS signal identifier */
+  u8 az;     /**< Azimuth angle in [0, 358] [deg * 2] */
```
silverjam: A u8 isn't enough to hold the whole range listed here, is that an issue?

nsirola: This is in units of 2 degrees, so values range from 0 to 179.

nsirola: Hmm I wonder if the comment should list the range of the raw values in this field instead of the range in degrees..

nsirola: Updated.

nsirola:
Rebased the yaml and unit test changes into one commit and all the generated files into another. Ready to merge.

anth-cole:
lgtm

PR #673, opened 2019-04-03
py3 compat: use errno attribute for socket.error

silverjam:
In Python version 2.6 socket.error was changed to be a subclass of IOError, which supports a `.errno` attribute, in Python 3 support for treating `socket.error` as a tuple was dropped.

### DIFF
--- a/python/sbp/client/drivers/network_drivers.py
+++ b/python/sbp/client/drivers/network_drivers.py
@@ -76,9 +76,9 @@ class TCPDriver(BaseDriver):
             return data
         except socket.timeout:
             self._connect()
-        except socket.error as e:
+        except socket.error as socket_exception:
             # this is fine
-            if e[0] == errno.EINTR:
+            if socket_exception.errno == errno.EINTR:
                 return
             # we really shouldn't be doing this
             raise IOError
--- a/python/sbp/client/drivers/network_drivers.py
+++ b/python/sbp/client/drivers/network_drivers.py
@@ -76,9 +76,9 @@ class TCPDriver(BaseDriver):
             return data
         except socket.timeout:
             self._connect()
-        except socket.error as socket_exception:
+        except socket.error as socket_error:
             # this is fine
-            if socket_exception.errno == errno.EINTR:
+            if socket_error.errno == errno.EINTR:
                 return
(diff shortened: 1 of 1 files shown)


python/sbp/client/drivers/network_drivers.py
```diff
@@ -76,9 +76,9 @@ def read(self, size):
             return data
         except socket.timeout:
             self._connect()
-        except socket.error as e:
+        except socket.error as socket_exception:
```
pmiettinen: Maybe `socket_error`. Name `socket_exception` seems to me would cover all the socket exception classes.

pmiettinen:
I wonder how this is still failing

```
=================================== FAILURES ===================================
_______________________________ test_tcp_logger ________________________________
    def test_tcp_logger():
      handler = tcp_handler(MsgPrintDep(text=b'abc').to_binary())
      ip, port = tcp_server(handler)
      port = "socket://%s:%s" % (ip, port)
      baud = 115200
      t0 = time.time()
      sleep = 0.1
      timeout = 1.0
      cb_context = {'assert_logger_called': False}
      def assert_logger(s, **metadata):
        cb_context['assert_logger_called'] = True
        assert s.preamble==0x55
        assert s.msg_type==0x10
        assert s.sender==66
        assert s.length==3
        assert s.payload==b'abc'
        assert s.crc==0xDAEE
      with PySerialDriver(port, baud) as driver:
        with Handler(Framer(driver.read, None, verbose=False), autostart=False) as link:
          link.add_callback(assert_logger)
          link.start()
          while True:
            if time.time() - t0 > timeout or cb_context['assert_logger_called']:
              break
            time.sleep(sleep)
>     assert cb_context['assert_logger_called'], "SBP msg callback function was not called"
E     AssertionError: SBP msg callback function was not called
E     assert False
```

python/sbp/client/drivers/network_drivers.py
silverjam: fixed

silverjam:
@pmiettinen the failure is intermittent

pmiettinen:
Yep, I just thought that the Handler autostart disablement would have fixed this.

PR #682, opened 2019-05-03
Copyless receive [ESD-1155]

pmiettinen:
## TODO
- [x] Implement `readinto(buf)` at least with `FileDriver`?
- [x] Generated files
- [x] Implement `from_binary()` for messages without payload
- [x] Resolve why locally passing unit tests are failing on Travis


python/sbp/client/framer.py
```diff
@@ -100,52 +146,118 @@ def _readall(self, size):
...
+
+            while l < size:
+                d = self._read(size - len(data))
+                if not d or self._broken:
+                    if allow_empty:
+                        return None
+                    raise StopIteration
+                data += d
+
+            #self._rx_buf = np.roll(self._rx_buf, -1 * self._rx_buf_idx)
```
silverjam: Remove?

python/sbp/client/framer.py
```diff
@@ -100,52 +146,118 @@ def _readall(self, size):
...
         if not preamble:
             return None
-        elif ord(preamble) != SBP_PREAMBLE:
+
+        preamble = preamble[0]
+
+        if preamble != _SBP_PREAMBLE:
             if self._verbose:
-                print("Host Side Unhandled byte: 0x%02x" % ord(preamble))
+                print("Preamble failure")
```
silverjam: Why not dump the actual preamble?

pmiettinen:
Benchmarking with command (JSON not expanded)
`time PYTHONPATH=../python/ python ../../piksi_tools/piksi_tools/serial_link.py  --file -p 0001.sbp --log --logfilename ./sbp_c.json --skip-metadata`
processing the ~80MB log file used in the Travis benchmark PR.

The improvement seems to be somewhat questionable:
- with master I have clocked results between 5m0s..5m22s, usually closer to the slower end
- unknown what is occasionally giving the performance boost, maybe the OS handling the file reads sometimes treats the script favorably?
- copy free PR in greedy mode performs more steadily around 5m7s area, worst cases seen around 5m15s

Conclusions:
- in master branch the parsing of the data is clearly the dominant time consumer
- the average improvement could be said to be around 10s but the variance is also high compared to the amount of improvement

Notes:
- `sbp2json.py` with `sbp.jit` processes the same file in 40-50 seconds but it bypasses the handler-framer apparatus using it own buffering mechanism

python/sbp/client/framer.py
```diff
@@ -39,17 +47,24 @@ class Framer(six.Iterator):
...
                  write,
+                 readinto=None,
                  verbose=False,
                  dispatcher=dispatch,
                  into_buffer=True,
-                 skip_metadata=False):
+                 skip_metadata=False,
+                 greedy=False):
         self._read = read
+        self._readinto = readinto
```
silverjam: How is this triggered?  I don't see a location where `readinto` is configured to be something other than `None`?

silverjam: Assuming you did something like this in piksi_tools:
```diff
diff --git a/piksi_tools/serial_link.py b/piksi_tools/serial_link.py
index c3fdc22f..6e32bb3b 100755
--- a/piksi_tools/serial_link.py
+++ b/piksi_tools/serial_link.py
@@ -338,7 +338,8 @@ def main(args):
     with Handler(Framer(driver.read,
                         driver.write,
                         args.verbose,
-                        skip_metadata=args.skip_metadata),
+                        skip_metadata=args.skip_metadata,
+                        readinto=True),
                  autostart=False) as link:
         # Logger with context
         with get_logger(args.log,
```

pmiettinen:
```sh
pasi@pasi-linux:~/swiftnav/piksi_tools$ git stash pop
On branch silverjam/sbp2json
Your branch is up-to-date with 'origin/silverjam/sbp2json'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   piksi_tools/serial_link.py

no changes added to commit (use "git add" and/or "git commit -a")
Dropped refs/stash@{0} (a9698df8813e72a62c42b7948fc7bf20b5780ecb)
pasi@pasi-linux:~/swiftnav/piksi_tools$ git diff
diff --git a/piksi_tools/serial_link.py b/piksi_tools/serial_link.py
index c3fdc22..0c8ca8a 100755
--- a/piksi_tools/serial_link.py
+++ b/piksi_tools/serial_link.py
@@ -337,8 +337,10 @@ def main(args):
     driver = get_base_args_driver(args)
     with Handler(Framer(driver.read,
                         driver.write,
+                        driver.readinto,
                         args.verbose,
-                        skip_metadata=args.skip_metadata),
+                        skip_metadata=args.skip_metadata,
+                        greedy=True),
                  autostart=False) as link:
         # Logger with context
         with get_logger(args.log,
pasi@pasi-linux:~/swiftnav/piksi_tools$
```

pmiettinen: Currently `readinto` is implemented only in the `FileDriver`.

python/sbp/client/framer.py
```diff
@@ -100,52 +146,121 @@ def _readall(self, size):
...
+            if self._greedy:
+                to_read = len(self._rx_buf) - self._rx_buf_unprocessed
+            else:
+                to_read = size - self._rx_buf_unprocessed
+
+            # Bytes are read into the tail of the buffer. If the index is too
+            # far to fit the new data, buffer needs to be rolled.
+            free_tail = len(self._rx_buf) - self._rx_buf_idx
+            if free_tail < to_read:
+                self._rx_buf = self._shift(self._rx_buf, -self._rx_buf_idx)
```
silverjam: We should try to eliminate this since it is effectively a copy.  I think we could use a "double buffer" idea to implement this, that is:

  1. allocate two buffers for all reads (say 4k each), buffer A and buffer B
  2. buffer A is initially assigned as the "read buffer", buffer B is the "leftover" buffer
  3. "read buffer" is passed to network/file layer that supports `readinto`
  4. read buffer is returned filled with data
  5. SBP processing is done, some amount of data is potentially left over, this becomes the "left over buffer", so "read buffer" and "leftover buffer" swap
  6. It's an error if 4k of "left-over" is accumulated and the left-over buffer is flushed

python/sbp/client/framer.py
silverjam: 👍

python/sbp/client/framer.py
pmiettinen: I can try that. Although I tried with big enough buffer size to eliminate rolling altogether and didn't see improvement.

silverjam:
Closing since this didn't bear fruit

PR #683, opened 2019-05-03
Added messages for separate orbit corrections and clock corrections

jbangelo:
Added messages for separate SSR orbit corrections and SSR clock corrections.

I followed the example for the other messages, where each message contains the corrections for a single signal. I think this would imply a single RTCM message could result in multiple SBP messages.


c/include/libsbp/ssr.h
```diff
@@ -164,6 +164,53 @@ satellite being tracked.
...
+#define SBP_MSG_SSR_CLOCK             0x05DA
+typedef struct SBP_ATTR_PACKED {
+  gps_time_sec_t time;               /**< GNSS reference time of the correction */
+  sbp_gnss_signal_t sid;                /**< GNSS signal identifier (16 bit) */
+  u8 update_interval;    /**< Update interval between consecutive corrections [s] */
+  u8 iod_ssr;            /**< IOD of the SSR correction. A change of Issue Of Data
+SSR is used to indicate a change in the SSR 
+generating configuration
+ */
+  u32 iod;                /**< Issue of broadcast ephemeris data or IODCRC (Beidou) */
```
jbangelo: @lloydmaza caught that this field isn't a part of the corresponding RTCM message. Will remove this field.

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -267,3 +267,92 @@ definitions:
...
+            type: GPSTimeSec
+            desc: GNSS reference time of the correction
+        - sid:
+            type: GnssSignal
+            desc: GNSS signal identifier (16 bit)
+        - update_interval:
+            type: u8
+            units: s
+            desc: Update interval between consecutive corrections
+        - iod_ssr:
```
scarcanague: So the problem is that the rtcm message is not stateless. the iod of the 1058 message refers to the iod of the related 1057 message previously sent:
image

For SBP we would probably want to keep it as stateless as possible so I think we might want the IOD in the clock only solution. 
@benjamin0 @jbangelo  What is the context of this PR? Do we want to reduce the bandwidth and be able to send a "high rate" clock sbp message?

mfine: @scarcanague context is supporting Geo++ SSR which is delivering clock and orbit messages independently.

scarcanague: OK. I think we want to add the IOD then, that might complicate the RTCM->SBP conversion but otherwise we will run into trouble if our SBP stream does not end up being sequenced properly

jbangelo: Adding in the IOD will mean that the RTCM->SBP conversion will require some additional state and logic to properly convert the messages.

If that's the route we want to take I'd argue that we shouldn't add these new messages at all and have the gnss-converters tool combine the separate orbit and clock SSR messages into our previously defined combined orbit and clock message. I think that would require similar logic/buffering in the conversion tool anyways.

Or do we want a 1-to-1 relationship between RTCM and SBP messages?

benjamin0: The goal of this work is to unblock being able to test the Geo++ C&O stream. I'd be in favor of making the assumption that we won't lose or receive messages out of order for now. If we end up relying on this implementation, then we'll need to revisit this assumption.

jbangelo: Seems reasonable. Should this be merged now?

jbangelo:
@scarcanague Should we just close this PR?

scarcanague:
@jbangelo Let's close it for now

PR #685, opened 2019-05-10
[WIP] parse_members optimizing [ESD-1157]

pmiettinen:
Original idea is to numba.jit the `parse_member()` but it appears to be tricky or even impossible because numba isn't able to interpret the complex/mixed datatypes.

While experimenting I got sidetracked quite far but made some interesting findings.

This is a result form using numpy's `frombuffer()` function which appears to do the parsing from buffer quickly but the problem is that there appears to be no naive way to convert structured numpy arrays to dictionary or JSON.

image

Profile shows that biggest part of the time is spent unpacking the results from numpy array to dictionary.. Actual parsing takes ~1/10 compared to the unpacking. (MsgObs context). Due to the slowness of unpacking the overall result is ~20-30% worse than current numba performance. If unpacking can be solved to be quicker, this approach shows good potential though.

For comparison the numba profile with same log:
image

Time values are not comparable between the runs. Reasoning is that the much larger amount of function calls is overwhelming the profiler in the numba case. Rather concentrate on which function are most heavy inside each run.


python/sbp/jit/msg.py
```diff
@@ -171,12 +171,12 @@ def judicious_round(f):
     # NOTE: Can't this be done already on libsbp side?
     return float(round(dec.Decimal(float(f)), abs(d.as_tuple().exponent)))
 
-
+@nb.jit
```
pmiettinen: these functions are not used in this solution

silverjam:
@pmiettinen Just for future reference, can you push new generated code?

PR #686, opened 2019-05-10
Added three new messages for gridded atmospheric SSR corrections.

martin-swift:
Implementing messages from https://drive.google.com/file/d/1jMLC2cjvVk6TUDGuktVLeK3kPyn4yi7S/view and https://drive.google.com/file/d/1RKjSgyBlxBhFz7ipExKh_Z2gFU_Yb0gR/view

I tried to be fairly faithful to the spec, but I tried to avoid things like bit packing and I was also working with the constraints of SBP, particularly the 255 max message size and the fact you can only have one variable length array in a message.


spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
             desc: |
-              Signal phase discontinuity counter. 
+              Signal phase discontinuity counter.
               Increased for every discontinuity in phase.
      - bias:
             type: s32
             units: 0.1 mm
             desc: Phase bias for specified signal
 
+ - SvId:
```
mfine: Would we want to use GNSS signal here?

martin-swift: We could reuse that; the reason I didn't is because band is not used or needed for this message, so I felt that including the band was a bit of a red herring.

mfine: I meant to remove `SvId` and use `GNSSSignal` instead. If the signal is not applicable and just the satellite - is that something we've done before? If that's the case, maybe it belongs in `gnss.yaml` alongside `GNSSSignal`.

benjamin0: Let's use the `SvId` type but move it into `gnss.yaml` (worth renaming `GNSSSat`?).

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
+    short_desc: Space vehicle identifier
+    desc: |
+      A (Constellation ID, satellite ID) tuple that unique identifies
```
benjamin0: typo: `uniquely`

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
+        type: GPSTime
+        desc: GNSS time of the STEC data
+    - num_msgs:
+        type: u8
+        desc: Number of messages in the dataset
+    - seq_num:
+        type: u8
+        desc: Postion of this message in the dataset
+    - ssr_update_interval:
+        type: u16
```
benjamin0: Can this be a `u8`?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
+        type: u8
+        desc: troposphere quality indicator
+    - index:
+        type: u8
+        desc: index of the grid point
+
+ - STECSatElement:
+     short_desc: SSR code biases corrections for a particular satellite.
+     desc: |
+       Code biases are to be added to pseudorange.
```
benjamin0: I think this might be the wrong description/short_desc?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
+        type: GPSTime
+        desc: GNSS time of the STEC data
+    - num_msgs:
+        type: u8
+        desc: Number of messages in the dataset
+    - seq_num:
+        type: u8
+        desc: Postion of this message in the dataset
+    - ssr_update_interval:
+        type: u16
```
benjamin0: same here

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
+    desc: |
+      The 3GPP message contains nested variable length arrays
+      which are not suppported in SBP, so each grid point will
+      be identified by the index.
+    fields:
+    - time:
+        type: GPSTime
+        desc: GNSS time of the STEC data
+    - num_msgs:
+        type: u8
```
benjamin0: Any chance we'll need more than 256 messages for even a pretty fine grid?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
+    - ssr_update_interval:
+        type: u16
+        desc: update interval in seconds
+    - iod_ssr:
+        type: u8
+        desc: range 0 - 15
+    - tropo_quality:
+        type: u8
+        desc: troposphere quality indicator
+    - index:
```
benjamin0: Do we need this in the header?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
+    fields:
+    - region_size_inverse:
+        type: u8
+        desc: inverse of region size
+    - area_width:
+        type: u16
+        desc: area width; see spec for details
+    - lat_nw_corner_enc:
+        type: u16
+        desc: encoded lattidue of the northwest corner of the grid
```
benjamin0: nit: typo

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -267,3 +411,51 @@ definitions:
...
+        - stec_sat_list:
+            type: array
+            fill: STECSatElement
+            desc: Array of STEC information for each space vehicle
+
+ - MSG_SSR_GRIDDED_CORRECTION:
+    id: 0x05F0
+    short_desc: Gridded troposphere and STEC residuals
+    desc: |
+      STEC residuals are per spcae vehicle, tropo is not.
```
benjamin0: typo

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -267,3 +411,51 @@ definitions:
             desc: |
               Phase biases corrections for a
               satellite being tracked.
+
+
+ - MSG_SSR_STEC_CORRECTION:
+    id: 0x05EB
+    short_desc: Slant Total Electon Content
+    desc: |
+      The STEC per space vehicle, given as polynomial approimation for
```
benjamin0: typo

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
+         type: SvId
+         desc: Unique space vehicle identifier
+     - stec_quality_indicator:
+         type: u8
+         desc: quality of STEC data
+     - stec_coeff:
+         type: array
+         size: 4
+         fill: s16
+         desc: coefficents of the STEC polynomial
```
ljbade: is it worth putting the polynomial formula in the description for clarity?

ljbade: and the units?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
+        type: SvId
+        desc: space vehicle identifier
+    - residual:
+        type: s16
+        desc: STEC residual
+
+ - GridElement:
+    short_desc: Grid datum for troposphere and STEC residuals
+    desc: |
+      Contains one tropo datum, plus STEC residuals for each space
```
ljbade: I guess we are trying to keep close to the 3GPP proposals, but combining tropo and iono does mean we can't have different update rates for the two types of data if there was a desire to reduce data rate

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
+         type: array
+         fill: STECResidual
+         desc: STEC Residual for the given space vehicle
+
+ - GridDefinitionHeader:
+    short_desc: Defines the grid for STEC and tropo grid messages
+    desc: |
+      Defines the grid for STEC and tropo grid messages.
+      Also includes an RLE encoded validity list.
+    fields:
```
ljbade: One shortcoming of the qualcomm grid vs ublox's is that it be nice to have a "grid ID" in the grid definition and correction messages. Then you could generate a series of smaller grids to cover a region and use the rover's position to filter out grids that are not relevant at the caster endpoint to reduce data rate.

ljbade:
A general improvement would be to include the units of the different fields in the description, that way the SBP spec is self contained and you don't need to go look up the units from the proposal docs

spec/yaml/swiftnav/sbp/ssr.yaml
martin-swift: done.

spec/yaml/swiftnav/sbp/ssr.yaml
martin-swift: f

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
+    - time:
+        type: GPSTime
+        desc: GNSS time of the STEC data
+    - num_msgs:
+        type: u8
+        desc: Number of messages in the dataset
+    - seq_num:
+        type: u8
+        desc: Postion of this message in the dataset
```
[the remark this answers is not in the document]
martin-swift: f

spec/yaml/swiftnav/sbp/ssr.yaml
martin-swift: not if we want the units to be seconds.  (this is one of the places I deviated from the spec)

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
```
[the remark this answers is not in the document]
martin-swift: f

spec/yaml/swiftnav/sbp/ssr.yaml
martin-swift: well, you're going to be able to fit quite a few GridElements in each message (64-ish?) so this will support a 127 x 127 grid.  Do we need finer?  

Also, should I include an absolute index for the GridElement?  We can calculate it implicitly from the position in the message and the message number, but it would serve as an extra sanity check, or allow us to send only specific grids.  I think it would be prudent...

martin-swift: done. I had that already, but in the wrong place.

spec/yaml/swiftnav/sbp/ssr.yaml
martin-swift: nope!

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
+        desc: range 0 - 15
+    - tropo_quality:
+        type: u8
+        desc: troposphere quality indicator
+    - index:
+        type: u8
+        desc: index of the grid point
+
+ - STECSatElement:
+     short_desc: SSR code biases corrections for a particular satellite.
```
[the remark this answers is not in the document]
martin-swift: yup, copypasta indeed.

spec/yaml/swiftnav/sbp/ssr.yaml
martin-swift: f

spec/yaml/swiftnav/sbp/ssr.yaml
martin-swift: nah.  that gets long.  Its in the spec, and anyone who actually needs to understand that part is going to have to read the spec anyway.

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,163 @@ definitions:
...
+
+ - STECResidual:
+    desc: STEC residual
+    fields:
+    - sv_id:
+        type: SvId
+        desc: space vehicle identifier
+    - residual:
+        type: s16
+        desc: STEC residual
```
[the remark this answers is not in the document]
martin-swift: f

spec/yaml/swiftnav/sbp/ssr.yaml
martin-swift: dunno... above my pay grade here :)  @anth-cole @benjamin0 ?

spec/yaml/swiftnav/sbp/ssr.yaml
martin-swift: we could always implement both; should we?  @anth-cole @benjamin0 
if so, can you point me to the Ublox spec?

spec/yaml/swiftnav/sbp/ssr.yaml
martin-swift: f

spec/yaml/swiftnav/sbp/ssr.yaml
martin-swift: f

spec/yaml/swiftnav/sbp/ssr.yaml
martin-swift: f

spec/yaml/swiftnav/sbp/ssr.yaml
benjamin0: Let's stick with the Qualcomm proposal for now to start simple and get things up and running then we can always rev this message later.

spec/yaml/swiftnav/sbp/ssr.yaml
benjamin0: Good point, @ljbade. I'm in favor of sticking with the Compact SSR messages for now, but we can always split them out later.

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,147 @@ definitions:
...
+        type: u8
+        desc: troposphere quality indicator
+
+ - STECSatElement:
+     desc: STEC for the given satellite.
+     fields:
+     - sv_id:
+         type: SvId
+         desc: Unique space vehicle identifier
+     - stec_quality_indicator:
```
martin-swift: So the spec tells us how to calculate "Q" from the this value, but what are the units of Q?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,147 @@ definitions:
...
+
+ - STECResidual:
+    desc: STEC residual
+    fields:
+    - sv_id:
+        type: SvId
+        desc: space vehicle identifier
+    - residual:
+        type: s16
+        desc: STEC residual (Scale factor 0.04 TECU)
```
martin-swift: and what's "U" here?  Total Electron Content...? @benjamin0

martin-swift:
Everyone happy?  I want to push the button on this today..

spec/yaml/swiftnav/sbp/ssr.yaml
scarcanague: Unit

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -47,19 +47,155 @@ definitions:
...
+        type: u16
+        units: coded value - 0 means not present (Swift convention)
+        desc: area width; see spec for details
+    - lat_nw_corner_enc:
+        type: u16
+        desc: encoded latitude of the northwest corner of the grid
+    - lon_nw_corner_enc:
+        type: u16
+        desc: encoded longitude of the northwest corner of the grid
+    - num_msgs:
```
benjamin0: Will there be more than 1 GridDefinitionHeader messages?

martin-swift: that was done to support the RLE validity list.  I think that thing is pretty janky.  Why would we ever send invalid data?  I'd be more than happy to snip that whole RLE list out of the message, and if we really need it, include validity flag in the correction/STEC messages.  Thoughts?

martin-swift: I think I see where they are going with this;  because it's broadcast, everyone is consuming it.  If the rover is in some corner case, we might want to say "hey, don't use this, because you are out of bounds from the model/whatever"

martin-swift: It would be a minor deviation from spec, but I could chop out the multi message support, since I think is it unlikely we'll see an RLE list that long, and you said these aren't prod messages.

benjamin0: Yeah I forgot about the RLE validity list, let's keep it.

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -267,3 +404,51 @@ definitions:
...
+      message to get the state space representation of the atmospheric
+      delay.
+    fields:
+        - header:
+            type: STECHeader
+            desc: Header of a STEC message
+        - stec_sat_list:
+            type: array
+            fill: STECSatElement
+            desc: Array of STEC information for each space vehicle
```
silverjam: The STECHeader defines a `index/total_messages` scheme for create a sequence of MSG_SSR_STEC_CORRECTION messages, in that case does this need to be an array?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -267,3 +404,51 @@ definitions:
             desc: |
               Phase biases corrections for a
               satellite being tracked.
+
+
+ - MSG_SSR_STEC_CORRECTION:
+    id: 0x05EB
+    short_desc: Slant Total Electon Content
```
silverjam: Typo?

```suggestion
    short_desc: Slant Total Electron Content
```

spec/yaml/swiftnav/sbp/ssr.yaml
silverjam: Or, I guess the idea is to put as many `STECSatElement` into one message as are possible?

martin-swift: yup!

silverjam:
@martin-swift Travis failure is the same semi-transient error we've seen for a while, I think this is good to merge.

PR #691, opened 2019-05-20
Conditional use of pthread library.

ipswift:
Library pthread should be used only when PLATFORM_IMPLEMENTATION_NO_PTHREAD is ON. It was found while building starling for Android where pthread is included into libandroid.

### DIFF
--- a/c/test/CMakeLists.txt
+++ b/c/test/CMakeLists.txt
@@ -7,7 +7,7 @@ else (CMAKE_CROSSCOMPILING)
     message(STATUS "Skipping unit tests, Check library not found!")
   else (NOT CHECK_FOUND)
 
-    if(LIBSBP_BUILTIN_PTHREADS) # In Android's libc implementation (bionic), pthreads is not a separate library
+    if (LIBSBP_BUILTIN_PTHREADS) # In Android's libc implementation (bionic), pthreads is not a separate library
       set(TEST_LIBS ${TEST_LIBS} ${CHECK_LIBRARIES} sbp m)
     else()
       set(TEST_LIBS ${TEST_LIBS} ${CHECK_LIBRARIES} pthread sbp m)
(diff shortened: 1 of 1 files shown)


c/test/CMakeLists.txt
```diff
@@ -7,7 +7,11 @@ else (CMAKE_CROSSCOMPILING)
     message(STATUS "Skipping unit tests, Check library not found!")
   else (NOT CHECK_FOUND)
 
-    set(TEST_LIBS ${TEST_LIBS} ${CHECK_LIBRARIES} pthread sbp m)
+    if(PLATFORM_IMPLEMENTATION_HAS_PTHREAD)
```
silverjam: Where is this defined?

silverjam: If this is a Swift specific option that we're enabling in an NDK build (since Bionic has pthread built-in), can we name this something like `LIBSBP_BUILTIN_PTHREADS`.

silverjam:
```suggestion
    if(LIBSBP_BUILTIN_PTHREADS)
```

c/test/CMakeLists.txt
```diff
@@ -7,7 +7,11 @@ else (CMAKE_CROSSCOMPILING)
     message(STATUS "Skipping unit tests, Check library not found!")
   else (NOT CHECK_FOUND)
 
-    set(TEST_LIBS ${TEST_LIBS} ${CHECK_LIBRARIES} pthread sbp m)
+    if(LIBSBP_BUILTIN_PTHREADS) # In Android's libc implementation (bionic), pthreads is not a separate library
```
silverjam:
```suggestion
    if (LIBSBP_BUILTIN_PTHREADS) # In Android's libc implementation (bionic), pthreads is not a separate library
```

Issue #681, opened 2019-04-17
MSG_INIT_BASE / Hotstart with known Position


michaelpantic:
Hi,

We're using and developing the https://github.com/ethz-asl/ethz_piksi_ros package for use on our Robots. Know we've come across a use case were a GPS receiver might loose view to the horizon for some time, but we still have relatively accurate position estimation in ENU frame through other means ( Visual-Inertial Odometry). 
Additionally, we want RTK-GPS to restore as fast as possible (every second counts!) after loosing it.

Is there a way to hotstart the Piksi Multi with a known baseline? I've found the MSG_INIT_BASE message, is that still in use on the Piksi Multi and how is the baseline sent to the piksi?

If this is not implemented, it would definitely be a very useful extension for us robotics people.

Thanks much for your help!

swiftnav-sarika:
Hello @michaelpantic 
 
Piksi Multi does not support hot start currently and it is currently not on our roadmap for future releases. The MSG_INIT_BASE is no longer supported in SBP messages. Please refer to SBP document here, https://support.swiftnav.com/customer/en/portal/articles/2492810-swift-binary-protocol. 

We appreciate you reaching out to us with this feature request and feedback. Please feel free to request additional support from the website here, https://support.swiftnav.com/. 

Thanks!

silverjam:
Added an internal issue to deprecate this message and update the sbp docs: https://swift-nav.atlassian.net/browse/ESD-1286

michaelpantic:
Thanks much for the quick response!

Too bad, it really would be a large advantage over other RTK systems, as many applications do use RTK GPS as a drift-free additional estimation to an existing state estimation stack. Especially for transitioning indoor-outdoor and likewise use cases it just would make sense to use a baseline prior to hot-start IAR.

At least in our experience the unpredictable/uninfluencable "time-to-fix" after loss is probably the largest impediment for tighter integration into our state estimation/mapping stack for autonomous UAVs.
If you plan to introduce that feature / or willing to provide lower level access in exchange for some testing of the feature, don't hesitate to reach out to us!

michaelpantic:
@marco-tranzatto @kholtmann FYI

PR #698, opened 2019-06-07
Deprecate MSG_INIT_BASE [ESD-1286]

silverjam:
Per #681 we don't plan to support this any time soon, so we're going to deprecate the message to avoid confusion.

**Note**: there's some additional change to `sbp.jit` folded into this since it wasn't (until now) part of the `all` target for updating generated code.


silverjam:
Travis failure looks like a (hopefully) transient issue with the sbp2json benchmark

PR #701, opened 2019-06-12
Don't use hardcoded include/link paths

woodfell:
/usr/local/(include|lib) are hardcoded in to the cmake system. This throws an error when trying to cross compile using buildroot. These paths should not be needed anyway.

### DIFF
--- a/c/test/CMakeLists.txt
+++ b/c/test/CMakeLists.txt
@@ -8,7 +8,7 @@ endif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
 FILE(GLOB generated_c_sources auto*.c)
 add_executable(test_libsbp check_main.c check_edc.c check_sbp.c ${generated_c_sources})
 
-target_link_libraries(test_libsbp ${TEST_LIBS})
+target_link_libraries(test_libsbp PRIVATE ${TEST_LIBS})
 set_target_properties(test_libsbp PROPERTIES
         C_STANDARD 99
         C_STANDARD_REQUIRED ON)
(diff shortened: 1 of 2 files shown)


woodfell:
Ah ok, in that case we can just wrap it up in an if(APPLE). Unfortunately the buildroot crosscompiler throws an error when passing -L/usr/local/lib

c/test/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ set_target_properties(test_libsbp PROPERTIES
 if(MSVC)
   target_include_directories(test_libsbp PRIVATE ${PROJECT_SOURCE_DIR}/include/libsbp/)
 else()
-  target_include_directories(test_libsbp PRIVATE ${PROJECT_SOURCE_DIR}/include/libsbp/ /usr/local/include/)
+  target_include_directories(test_libsbp PRIVATE ${PROJECT_SOURCE_DIR}/include/libsbp/)
```
benjaminaltieri:
```suggestion
  target_include_directories(test_libsbp PRIVATE ${PROJECT_SOURCE_DIR}/include/libsbp/)
  if (APPLE)
    target_include_directories(test_libsbp PRIVATE /usr/local/include/)
  endif ()
```

c/test/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ set_target_properties(test_libsbp PROPERTIES
 if(MSVC)
```
benjaminaltieri: this would likely solve the issue on PBR(P) side
```suggestion
if(MSVC OR CROSS_COMPILING)
```

woodfell: I'm not sure whether any other compiler would also break, so I've made a change to only add /usr/local/* on apple platforms

PR #703, opened 2019-06-20
Change ssr_update_interval to u8.

martin-swift:


martin-swift:
@silverjam - can you help me with the autogenerated code again?  Did you ever manage to dockerize that?

ljbade:
@martin-swift looks good to me. Only other improvement I can think of (but not really critical) is you could use GpsTimeSec instead of GpsTime, the former has whole second resolution latter has nanosecond resolution, I noticed our other SSR messages only use the second resolution type.

PR #704, opened 2019-06-21
SSR STEC and gridded correction message refinements

ljbade:
Further refinements to the SSR STEC and gridded correction messages for consistency with the existing SSR messages.

I also fleshed out the description and unit fields to make it easier for someone to use the messages without having to cross reference back to the CLAS spec.

### DIFF
--- a/spec/yaml/swiftnav/sbp/ssr.yaml
+++ b/spec/yaml/swiftnav/sbp/ssr.yaml
@@ -20,7 +20,7 @@ definitions:
      short_desc: SSR code biases corrections for a particular satellite.
      desc: |
       Code biases are to be added to pseudorange.
-      The corrections are conform with typical RTCMv3 MT1059 and 1065.
+      The corrections conform with typical RTCMv3 MT1059 and 1065.
      fields:
      - code:
             type: u8
@@ -33,7 +33,7 @@ definitions:
      short_desc: SSR phase biases corrections for a particular satellite.
      desc: |
       Phase biases are to be added to carrier phase measurements.
-      The corrections are conform with typical RTCMv3 MT1059 and 1065.
+      The corrections conform with typical RTCMv3 MT1059 and 1065.
      fields:
      - code:
             type: u8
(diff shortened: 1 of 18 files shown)


spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -405,15 +422,14 @@ definitions:
...
  - MSG_SSR_STEC_CORRECTION:
     id: 0x05EB
     short_desc: Slant Total Electron Content
     desc: |
       The STEC per space vehicle, given as polynomial approximation for
-      a given grid.  This should be combined with SSR-GriddedCorrection
+      a given grid area.  This should be combined with MSG_SSR_GRIDDED_CORRECTION
       message to get the state space representation of the atmospheric
-      delay.
+      delay. It is typically equivalent to the QZSS CLAS Sub Type 8 messages
```
martin-swift: ?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -405,15 +422,14 @@ definitions:
...
               satellite being tracked.
 
-
  - MSG_SSR_STEC_CORRECTION:
     id: 0x05EB
     short_desc: Slant Total Electron Content
     desc: |
       The STEC per space vehicle, given as polynomial approximation for
-      a given grid.  This should be combined with SSR-GriddedCorrection
+      a given grid area.  This should be combined with MSG_SSR_GRIDDED_CORRECTION
```
martin-swift: not sure why we need "area".  We have a grid that has been defined.  Of course a grid covers an area; I feel like this is tautological and thus more confusing.

ljbade: sure i can drop that

spec/yaml/swiftnav/sbp/ssr.yaml
ljbade: I wanted to put a reference to the CLAS message since you have to use the correction algorithm from that document. This way someone not familiar will know where to go find the details

PR #706, opened 2019-06-25
Modify to work with Rust's bindgen

silverjam:
This change allows me to write an "entrypoint" header for bindgen so that I can automatically generate FFI bindings.


silverjam:
FYI, going to merge this, seems low risk

c/include/libsbp/common.h
```diff
@@ -28,6 +30,7 @@
  * `int` which can lead to portability issues between different platforms.
  * \{ */
 
+#ifndef _RUSTC_BINDGEN_
```
jbangelo: Does bindgen define it's own `s8` etc types? Do we actually need to not include the stdlib headers?

silverjam: I manually defined equivalent types to avoid the other includes that this generates.  The lack of includes here is to limit the what bindgen tries to generate.  Bindgen will attempt to generate bindings for everything it the `.h`-- so we want to avoid trying to generate bindings for everything in the c stdlib.

PR #712, opened 2019-07-10
Support running w/o Numba/Numpy installed

silverjam:


python/sbp/msg.py
```diff
@@ -21,44 +20,84 @@
...
-
-SENDER_ID = _SENDER_ID
-SBP_PREAMBLE = _SBP_PREAMBLE
-
-crc_buffer = np.zeros(512, dtype=np.uint8)
+    try:
+      from sbp.jit import parse
+      parse.compile()
+      parse_jit = importlib.import_module('sbp.jit.' + parse_jit_name)
+    except ImportError:
```
denniszollo: Maybe if user gets here we can add some helpful hints to a user.

1. Indicate which libraries / files were missing so that if someone is messing with pyinstaller they might realize that they forget to include the right stuff and won't need jit if they just get the right files. 
1. Refer a user to the last version of libsbp that didn't require jit compilation / numba.  In that way people on odd systems who can't get these dependencies easily can just fallback to a simpler (but slower) libsbp.

silverjam: Item #2 should not be necessary with this change, we should automatically just fallback to parsing without Numba/Numpy.

silverjam:
@denniszollo I added some documentation around expected failure cases

python/sbp/msg.py
silverjam: FYI, for item 1 I haven't been able to figure out how to get PyInstaller to work with the Numba/JIT stuff yet (I've been able to get PyInstaller to package the library, but once the app starts we can't find it), as you saw recent versions of the library broke things like the Windows build, so this "graceful fallback" is necessary for these cases for now.

silverjam: I think instead of a warning, setting a property within the library might be appropriate?  So if someone really needs the speed from the JIT code, they can assert on that property.

PR #719, opened 2019-08-06
ORI-594 Fix up SSR atmospherics documentation

ljbade:


spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -113,7 +113,7 @@ definitions:
         type: u8
         desc: |
           Quality of the troposphere data. Encoded following RTCM DF389
-          specifcation but as TECU instead of m.
+          specifcation in units of m.
```
martin-swift: is it meters?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -125,7 +125,7 @@ definitions:
         type: u8
         desc: |
           Quality of the STEC data. Encoded following RTCM DF389 specifcation
-          but as TECU instead of m.
+          but in units of TECU instead of m.
```
martin-swift: I think it's some weird coded value that isn't TECU or meters.  still not quite how to interpret that spec...

spec/yaml/swiftnav/sbp/ssr.yaml
ljbade: From the CLAS spec:
image

From RTCM:
image

So they are basically reusing the SSR URA formula which takes the index (in the u8) and returns a value in metres.

We could try copying that formula into the YAML but it wouldn't look pretty.

I'm open to alternative wordings if you think there is something clearer we could use.

spec/yaml/swiftnav/sbp/ssr.yaml
ljbade: From the CLAS spec:
image

From RTCM:
image

In this case my reading is that they want you to use the SSR URA formula, but just replace metres with TECU, so it keeps the same numerical value, just interpreted differently.

I'm open to alternative wordings if you think there is something clearer we could use.

ljbade:
@silverjam thank you!

ljbade:
@silverjam after merging this I have discovered the c code didn't generate

PR #723, opened 2019-09-03
Add c api to call user function for every successfully deframed SBP message. [ESD-998]

denniszollo:
This allows a user to use sbp_process_generic and call the user function for every sbp message.  It doesn't change our API at all and could simplify many things like logging SBP messages to a file without needing to copy over the sbp deframing code. 

[ESD-998]

[ESD-998]: https://swift-nav.atlassian.net/browse/ESD-998

### DIFF
--- a/c/include/libsbp/sbp.h
+++ b/c/include/libsbp/sbp.h
@@ -28,6 +28,9 @@ extern "C" {
 #define SBP_OK_CALLBACK_EXECUTED 1
 /** Return value indicating message decoded with no associated callback in sbp_process. */
 #define SBP_OK_CALLBACK_UNDEFINED 2
+/** Return value indicating message decoded and only an all_msg function was
+ * executed. */
+#define SBP_OK_ALLMSG_FN_EXECUTED_ONLY 3
 /** Return value indicating an error with the callback (function defined). */
 #define SBP_CALLBACK_ERROR -1
 /** Return value indicating a CRC error. */
@@ -46,6 +49,8 @@ extern "C" {
 
 /** SBP callback function prototype definition. */
 typedef void (*sbp_msg_callback_t)(u16 sender_id, u8 len, u8 msg[], void *context);
+typedef void (*sbp_allmsg_fn_t)(u16 sender_id, u16 msg_type, u8 len, u8 msg[],
+                                void *context);
 
 /** SBP callback node.
  * Forms a linked list of callbacks.
@@ -76,12 +81,20 @@ typedef struct {
   u8 msg_buff[256];
   void* io_context;
   sbp_msg_callbacks_node_t* sbp_msg_callbacks_head;
+  sbp_allmsg_fn_t allmsg_fn;
+  void *allmsg_context;
 } sbp_state_t;
 
 /** \} */
 
+/* registers a callback to be called on a particular msg_type */
 s8 sbp_register_callback(sbp_state_t* s, u16 msg_type, sbp_msg_callback_t cb, void* context,
                          sbp_msg_callbacks_node_t *node);
+/* sets a function to be called on ANY deframed message */
+s8 sbp_set_allmsg_fn(sbp_state_t *s, sbp_allmsg_fn_t fn, void *context);
+/* gets the function to be called on ANY deframed message */
+sbp_allmsg_fn_t sbp_get_allmsg_fn(sbp_state_t *s);
+s8 sbp_remove_allmsg_fn(sbp_state_t *s);
 s8 sbp_remove_callback(sbp_state_t *s, sbp_msg_callbacks_node_t *node);
 void sbp_clear_callbacks(sbp_state_t* s);
 void sbp_state_init(sbp_state_t *s);
--- a/c/include/libsbp/version.h
+++ b/c/include/libsbp/version.h
@@ -25,7 +25,7 @@
 /** Protocol minor version. */
 #define SBP_MINOR_VERSION 6
 /** Protocol patch version. */
-#define SBP_PATCH_VERSION 4
+#define SBP_PATCH_VERSION 6
 
 /** \} */
 
--- a/c/src/sbp.c
+++ b/c/src/sbp.c
@@ -53,6 +53,21 @@
  *
  * where `SBP_MY_MSG_TYPE` is the numerical identifier of your message type.
  *
+ * If you'd like to call a function for EVERY sbp message, setup an allmsg function
+ * with the type sbp_allmsg_fn_t.  It must be of the form:
+ *
+ * ~~~
+ * void my_allmsg_fn(u16 sender_id, u16 msg_type, u8 len, u8 msg[], void *context)
+ * {
+ *   // Process msg.
+ * }
+ * ~~~
+ *
+ * Then register the allmsg callback with the SBP library as follows:
+ * ~~~
+ * sbp_register_allmsg_fn(&sbp_state, &my_allmsg_fn, &context);
+ * ~~~
+ *
  * You must now call sbp_process() periodically whenever you have received SBP
  * data to be processed, e.g. from the serial port. Remember sbp_process() may
  * not use all available data so keep calling sbp_process() until all the
@@ -227,6 +242,44 @@ s8 sbp_remove_callback(sbp_state_t *s, sbp_msg_callbacks_node_t *node)
   return SBP_CALLBACK_ERROR;
 }
 
+/** Set a function to be called for ANY message type.
+ *
+ * \param s        pointer to sbp_state_t struct in use
+ * \param fn       Pointer to message callback function
+ * \param context  Pointer to context for callback function
+ * \return `SBP_OK` (0) if successful, `SBP_NULL_ERROR` if fn is NULL
+ *
+ */
+s8 sbp_set_allmsg_fn(sbp_state_t *s, sbp_allmsg_fn_t fn, void *context) {
+  /* Check our callback function pointer isn't NULL. */
+  if (0 == fn) {
+    return SBP_NULL_ERROR;
+  }
+  s->allmsg_fn = fn;
+  s->allmsg_context = context;
+  return SBP_OK;
+}
+
+/** Retrieve the callback for ANY message type.
+ *
+ * \param s        pointer to sbp_state_t struct in use
+ * \return sbp_allmsg_fn_t* to allmsg fn pointer or NULL if none 
+ *
+ */
+sbp_allmsg_fn_t sbp_get_allmsg_fn(sbp_state_t *s) {
+  return s->allmsg_fn;
+}
+
+/** Remove allmsg function pointer and context.
+ *
+ * \return `SBP_OK` (0) if successful
+ */
+s8 sbp_remove_allmsg_fn(sbp_state_t *s) {
+  s->allmsg_fn = 0;
+  s->allmsg_context = 0;
+  return SBP_OK;
+}
+
 /** Clear all registered callbacks.
  * This is probably only useful for testing but who knows!
  */
@@ -252,6 +305,9 @@ void sbp_state_init(sbp_state_t *s)
 
   /* Clear the callbacks, if any, currently in s */
   sbp_clear_callbacks(s);
+
+  /* Clear the allmsg_fn, if any, currently set on s */
+  sbp_remove_allmsg_fn(s); 
 }
 
 
@@ -302,7 +358,8 @@ void sbp_state_set_io_context(sbp_state_t *s, void *context)
  * \return `SBP_OK` (0) if successful but no complete message yet,
  *         `SBP_OK_CALLBACK_EXECUTED` (1) if message decoded and callback executed,
  *         `SBP_OK_CALLBACK_UNDEFINED` (2) if message decoded with no associated
- *         callback, and `SBP_CRC_ERROR` (-2) if a CRC error
+ *         callback, SBP_OK_ALLMSG_FN_EXECUTED_ONLY (3) if only the generic
+ *         callback for every msg was called, and `SBP_CRC_ERROR` (-2) if a CRC error
  *         has occurred. Thus can check for >0 to ensure good processing.
  */
 s8 sbp_process(sbp_state_t *s, s32 (*read)(u8 *buff, u32 n, void *context))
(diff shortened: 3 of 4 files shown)


benjaminaltieri:
had a backlog item to add this feature (https://swift-nav.atlassian.net/browse/ESD-998). thanks for picking this up!

c/src/sbp.c
```diff
@@ -306,6 +306,12 @@ void sbp_state_set_io_context(sbp_state_t *s, void *context)
  *         has occurred. Thus can check for >0 to ensure good processing.
  */
 s8 sbp_process(sbp_state_t *s, s32 (*read)(u8 *buff, u32 n, void *context))
+{
+return sbp_process_generic(s, read, &sbp_process_payload);
```
benjaminaltieri: indent

c/src/sbp.c
```diff
@@ -306,6 +306,12 @@ void sbp_state_set_io_context(sbp_state_t *s, void *context)
...
+}
+
+s8 sbp_process_generic(sbp_state_t *s, s32 (*read)(u8 *buff, u32 n, void *context),
+    s8 (*sbp_process_payload) (sbp_state_t *s, u16 sender_id, u16 msg_type, u8 msg_len, u8 payload[]))
```
benjaminaltieri: probably want to avoid aliasing `sbp_process_payload` here with the arg name. also you might be able to reuse `sbp_msg_callback_t` if you pass the `sbp_state_t` as the `context *` arg

c/include/libsbp/sbp.h
```diff
@@ -87,6 +87,8 @@ void sbp_clear_callbacks(sbp_state_t* s);
 void sbp_state_init(sbp_state_t *s);
 void sbp_state_set_io_context(sbp_state_t *s, void* context);
 s8 sbp_process(sbp_state_t *s, s32 (*read)(u8 *buff, u32 n, void* context));
+s8 sbp_process_generic(sbp_state_t *s, s32 (*read)(u8 *buff, u32 n, void *context),
```
jbangelo: Nit: This is fairly generic name. Maybe something like `sbp_process_msg_catchall` or something like that? As an alternative I think a comment explaining what this is meant for would help. I know we don't have any comments for the other functions, but they are sorta-kinda self explanatory while this one isn't.

c/include/libsbp/sbp.h
```diff
@@ -87,6 +87,8 @@ void sbp_clear_callbacks(sbp_state_t* s);
...
+    s8 (*sbp_process_payload) (sbp_state_t *s, u16 sender_id, u16 msg_type, u8 msg_len, u8 payload[]));
```
jbangelo: I'd be hesitant to have the generic callback take in `sbp_state_t *` directly, the user's code shouldn't be using that. Instead maybe have a `void *context` as an arg to the generic callback and pass it into `sbp_process_generic` as well?

c/src/sbp.c
```diff
@@ -383,7 +389,7 @@ s8 sbp_process(sbp_state_t *s, s32 (*read)(u8 *buff, u32 n, void *context))
       if (s->crc == crc) {
 
         /* Message complete, process it. */
-        s8 ret = sbp_process_payload(s, s->sender_id, s->msg_type, s->msg_len,
+        s8 ret = (*sbp_process_payload)(s, s->sender_id, s->msg_type, s->msg_len,
```
jbangelo: If you give this arg a different name then you shouldn't need to deference the function pointer here.

c/src/sbp.c
denniszollo: I renamed the existing sbp_process_payload to sbp_process_payload_cbs to notate that it goes through the linked list of callbacks.  This one, I renamed to sbp_process_payload_fn to indicate it is a function pointer.

c/include/libsbp/sbp.h
denniszollo: I renamed to sbp_process_all and added some comments per your suggestion.

c/include/libsbp/sbp.h
denniszollo: Agreed, I think a context was better and more flexible.  For the legacy case, the context is the head of the linked list.

c/include/libsbp/sbp.h
```diff
@@ -86,9 +86,18 @@ s8 sbp_remove_callback(sbp_state_t *s, sbp_msg_callbacks_node_t *node);
...
 void sbp_state_set_io_context(sbp_state_t *s, void* context);
+/* sbp_process has existed since dawn of sbp. It fires registered callbacks. */
 s8 sbp_process(sbp_state_t *s, s32 (*read)(u8 *buff, u32 n, void* context));
-s8 sbp_process_payload(sbp_state_t *s, u16 sender_id, u16 msg_type, u8 msg_len,
-    u8 payload[]);
+/* sbp_process_all allows a user's process_payload_fn to be called for every
+ * deframed sbp message. This is useful for logging or forwarding any message
+ * regardless of type. Optionally, it can call sbp_process_payload_cbs itself
+ * to invoke matching callbacks in the linked list as well.*/
+s8 sbp_process_all(sbp_state_t *s, s32 (*read)(u8 *buff, u32 n, void *context),
```
benjaminaltieri: I definitely prefer this naming scheme, but I agree with @silverjam that we shouldn't change the public API

benjaminaltieri: if need be we can marshall `sbp_process_payload` directly to `sbp_process_payload_cbs` with a comment stating that this api is deprecated.

denniszollo:
I've updated one more time in such a way that there is no change to existing API.  I've also added unit tests.

c/src/sbp.c
```diff
@@ -227,13 +242,45 @@ s8 sbp_remove_callback(sbp_state_t *s, sbp_msg_callbacks_node_t *node)
...
+ * \param msg_type Message type associated with callback
+ * \param fn       Pointer to message callback function
+ * \param context  Pointer to context for callback function
+ * \return `SBP_OK` (0) if successful, `SBP_NULL_ERROR` if fn is NULL
+ *
+ */
+s8 sbp_register_allmsg_callback(sbp_state_t *s, sbp_allmsg_callback_fn_t fn,
+                                void *context) {
+  /* Check our callback function pointer isn't NULL. */
+  if (0 == fn) {
```
benjaminaltieri: `NULL` preferred for pointer checks

c/src/sbp.c
```diff
@@ -227,13 +242,45 @@ s8 sbp_remove_callback(sbp_state_t *s, sbp_msg_callbacks_node_t *node)
...
+  s->allmsg_context = context;
+  return SBP_OK;
+}
+
+/** Remove allmsg function pointer.
+ *
+ * \return `SBP_OK` (0) if successful
+ */
+s8 sbp_remove_allmsg_callback(sbp_state_t *s) {
+  s->allmsg_cb = 0;
```
benjaminaltieri: `NULL` preferred

c/src/sbp.c
```diff
@@ -227,13 +242,45 @@ s8 sbp_remove_callback(sbp_state_t *s, sbp_msg_callbacks_node_t *node)
...
+
 /** Clear all registered callbacks.
  * This is probably only useful for testing but who knows!
  */
 void sbp_clear_callbacks(sbp_state_t *s)
 {
   /* Reset the head of the callbacks list to NULL. */
   s->sbp_msg_callbacks_head = 0;
+  /* Reset the all_cb to 0. */
+  s->allmsg_cb = 0;
```
benjaminaltieri: I would argue this shouldn't be included, `sbp_remove_allmsg_callback` already covers clearing this

c/src/sbp.c
```diff
@@ -423,6 +473,12 @@ s8 sbp_process_payload(sbp_state_t *s, u16 sender_id, u16 msg_type, u8 msg_len,
       ret = SBP_OK_CALLBACK_EXECUTED;
     }
   }
+  if (0 != s->allmsg_cb) {
```
benjaminaltieri: use `NULL`

c/src/sbp.c
```diff
@@ -423,6 +473,12 @@ s8 sbp_process_payload(sbp_state_t *s, u16 sender_id, u16 msg_type, u8 msg_len,
...
+    (*s->allmsg_cb)(sender_id, msg_type, msg_len, payload, s->allmsg_context);
```
benjaminaltieri:
```suggestion
    *(s->allmsg_cb)(sender_id, msg_type, msg_len, payload, s->allmsg_context);
```

benjaminaltieri: or omit deref, it isn't technically necessary

c/src/sbp.c
```diff
@@ -227,13 +242,45 @@ s8 sbp_remove_callback(sbp_state_t *s, sbp_msg_callbacks_node_t *node)
...
+ * Register a callback that is called when any
+ * valid message is received.
+ *
+ * \param msg_type Message type associated with callback
+ * \param fn       Pointer to message callback function
+ * \param context  Pointer to context for callback function
+ * \return `SBP_OK` (0) if successful, `SBP_NULL_ERROR` if fn is NULL
+ *
+ */
+s8 sbp_register_allmsg_callback(sbp_state_t *s, sbp_allmsg_callback_fn_t fn,
```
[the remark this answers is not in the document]
denniszollo: that is good point; I will rename to "sbp_set_allmsg_callback".  Do you like the "allmsg" wording?

c/src/sbp.c
denniszollo: I agree, but in order to do this it requires adding a new #include that wasn't here when the original author wrote.  I thought perhaps there was a portability reason to not include stddefs.   I'm happy to follow up with a separate PR to add in NULL for null pointer and the new include, but for now I wanted to follow the existing pattern.

c/src/sbp.c
denniszollo: I was debating that as well.  If I don't use the word "callback" for the "allmsg_fuction", then this would be very clear and I wouldn't need to change sbp_clear_callbacks.

c/src/sbp.c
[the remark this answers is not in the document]
denniszollo: Now I am thinking to call it sbp_set_allmsg_fn() or something similar as to semantically differentiate between this feature, and the linked list of callbacks for particular message types.

c/src/sbp.c
denniszollo: after renaming I agree and changed.

c/src/sbp.c
denniszollo: done.

c/src/sbp.c
denniszollo: as discussed, separate PR to be up today

c/src/sbp.c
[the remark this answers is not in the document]
denniszollo: agree I renamed

c/include/libsbp/version.h
```diff
@@ -25,7 +25,7 @@
 /** Protocol minor version. */
 #define SBP_MINOR_VERSION 6
 /** Protocol patch version. */
-#define SBP_PATCH_VERSION 4
+#define SBP_PATCH_VERSION 6
```
silverjam: Should need to touch this, it's bumped automatically during a release.

c/include/libsbp/sbp.h
```diff
@@ -76,12 +81,20 @@ typedef struct {
...
+  void *allmsg_context;
 } sbp_state_t;
 
 /** \} */
 
+/* registers a callback to be called on a particular msg_type */
 s8 sbp_register_callback(sbp_state_t* s, u16 msg_type, sbp_msg_callback_t cb, void* context,
                          sbp_msg_callbacks_node_t *node);
+/* sets a function to be called on ANY deframed message */
+s8 sbp_set_allmsg_fn(sbp_state_t *s, sbp_allmsg_fn_t fn, void *context);
```
silverjam: How about `sbp_register_allmsg_callback` to match `sbp_register_callback`?

c/include/libsbp/sbp.h
```diff
@@ -76,12 +81,20 @@ typedef struct {
   u8 msg_buff[256];
   void* io_context;
   sbp_msg_callbacks_node_t* sbp_msg_callbacks_head;
+  sbp_allmsg_fn_t allmsg_fn;
```
silverjam: No a huge fall of `allmsg`, would prefer `all_msg`, but I can live with it.

c/src/sbp.c
```diff
@@ -227,6 +242,43 @@ s8 sbp_remove_callback(sbp_state_t *s, sbp_msg_callbacks_node_t *node)
...
+sbp_allmsg_fn_t sbp_get_allmsg_fn(sbp_state_t *s) {
+  return s->allmsg_fn;
+}
+
+/** Remove allmsg function pointer.
+ *
+ * \return `SBP_OK` (0) if successful
+ */
+s8 sbp_remove_allmsg_fn(sbp_state_t *s) {
+  s->allmsg_fn = 0;
```
silverjam: Clear `allmsg_context` too?

c/include/libsbp/sbp.h
benjaminaltieri: I believe I suggested `set` since this is not registered, as in place in a registry of callbacks along with other callbacks. there is only one `allmsg` function to either set or clear

c/include/libsbp/sbp.h
denniszollo: I just didn't want anyone to be confused that all_msg meant all messages.

benjaminaltieri: agree on this point, considering one of the use cases of this function (hence the different signature) is to receive the entire raw frame so the binary blob can be written to file. from this perspective I would almost prefer a name like `sbp_raw_frame_handler` or something to that effect

c/include/libsbp/sbp.h
denniszollo: I was worried about calling it a "callback".  If I do, then I think it makes sense that sbp_clear_callbacks clears this allmsg callback as well, which is kind of a change in behavior for that part of the API that Ben wasn't too fond of.

c/src/sbp.c
denniszollo: yes I agree.

denniszollo:
Superceded by #725

PR #724, opened 2019-09-06
JSONLogIterator - Fix the iteration if the JSON is already unwrapped

ebethon:


denniszollo:
Seems good and is an improvement.  It might be nice to print out useful error messages explaining what flavors of JSON the tool supports if none of the keys are there.

PR #728, opened 2019-09-19
Make a msg_buff pointer for backwards compatibility

denniszollo:


c/src/sbp.c
```diff
@@ -325,6 +325,7 @@ void sbp_state_init(sbp_state_t *s)
 
   /* Set the IO context pointer, passed to read and write functions, to NULL. */
   s->io_context = 0;
+  s->msg_buff = &(s->frame_buff[6]);
```
benjaminaltieri: let's use the macro 🙂

silverjam:
```suggestion
 s->msg_buff = SBP_FRAME_MSG_PAYLOAD(s->frame_buff);
```

PR #730, opened 2019-09-19
libsbp version 2.7.0

silverjam:


silverjam:
@jbangelo FYI, the rust generator doesn't appear to produce stable output, the mod.rs in this release is jumbled from the previous time rust was generated.

jbangelo:
Odd. Looks like I'm able to reproducible locally. Will look into it when I have time. Do the other generators need to do anything special to keep their output consistent? Seems like it could be inconsistent ordering of the message definitions.

silverjam:
I think the java generator had similar issues, a fix is probably as simple just sorting the list of keys you get back from a dictionary.

PR #734, opened 2019-10-02
Add option to annotate with timestamp (simulate Swift Console output)

silverjam:
Attempting to add an option to the Python sbp2json to take in SBP binary and output something similar to what the console outputs.  Timestamps are somewhat synthetic since they follow a GPS epoch rather than the current system time.  The UUID is completely synthetic and fixed, with the intent that it could be used to identify a file that was converted using this tool rather than coming from the Swift Console.

### DIFF
--- a/python/sbp/sbp2json.py
+++ b/python/sbp/sbp2json.py
@@ -11,7 +11,6 @@ from sbp import msg as msg_nojit
 from sbp.navigation import SBP_MSG_UTC_TIME
 from sbp.table import dispatch as dispatch_nojit
 
-CURRENT_MSG_INDEX = 0
 LAST_UTC_TIME = datetime.fromtimestamp(0)
 
 NORM = os.environ.get('NOJIT') is not None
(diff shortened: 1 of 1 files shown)


denniszollo:
I understand that this is fulfilling some kind of need, but isn't there a way to modify the consumer of the json to not require the timestamp?

python/sbp/sbp2json.py
```diff
@@ -140,10 +149,34 @@ def floatstr(o, allow_nan=self.allow_nan,
             self.skipkeys, _one_shot)
         return _iterencode(o, 0)
 
-def get_jsonable(res):
-    if hasattr(res, 'to_json_dict'):
-        return res.to_json_dict()
-    return res
+
```
denniszollo: What if the log has no MSG_UTC_TIME?  I feel like you might want to warn if LAST_UTC_TIME is never set by having received the message.

silverjam:
I think I'm find with ditching this, the right answer was what you suggested, just change the client to use the "non-annotated" JSON

PR #739, opened 2019-10-21
Add C++ support

jbangelo:
This PR adds a few lightweight C++ classes that wrap around the existing libsbp C interface. The main item of interest here is the `MessageHandler`. This class automates the infrastructure for registering a callback to have a member function called upon receipt of a sbp message.

Here is an example of how it would be used
```
class ECEFHandler : private sbp::MessageHandler<msg_gps_time_t, msg_pos_ecef_t, msg_pos_ecef_cov_t> {
 public:
  ECEFHandler(sbp::State *state) : private sbp::MessageHandler<msg_gps_time_t, msg_pos_ecef_t, msg_pos_ecef_cov_t>(state)
  {
    /* Do other constructor stuff */
  }

 private:
  void handle_sbp_msg(uint16_t sender_id, uint8_t message_length, const msg_gps_time_t& msg)
  {
    /* handle GPS time message */
  }

  void handle_sbp_msg(uint16_t sender_id, uint8_t message_length, const msg_pos_ecef_t& msg)
  {
    /* handle pos ECEF message */
  }

  void handle_sbp_msg(uint16_t sender_id, uint8_t message_length, const msg_pos_ecef_cov_t& msg)
  {
    /* handle pos ecef cov message */
  }
};
```

The instantiation and registration of the `sbp_msg_callbacks_node_t` objects are all taken care of automatically by the `sbp::MessageHandler` constructor, and are unregistered in the destructor. The derived class is able to publicly or privately inherit from `sbp::MessageHandler`, and the `handle_sbp_msg` member functions can be public or private. Each `handle_sbp_msg` is a virtual member function so additional levels of inheritance can be employed, though we do incur the overhead of the vtable lookup at run time. The mapping of message struct to message ID is achieved via a set of type traits that are automatically generated from the specification YAML. Currently only the message ID is included in the message type traits, but additional information could be added in the future if deemed useful.

The other classes are thin wrappers around the existing C structs, and probably don't need much explanation.

### DIFF
--- a/c/include/libsbp/cpp/message_handler.h
+++ b/c/include/libsbp/cpp/message_handler.h
@@ -1,0 +1,174 @@
+/**
+ * Copyright (C) 2019 Swift Navigation Inc.
+ * Contact: Swift Navigation <dev@swiftnav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef SBP_CPP_MESSAGE_HANDLER_H_
+#define SBP_CPP_MESSAGE_HANDLER_H_
+
+#include <cassert>
+#include <array>
+
+#include <libsbp/cpp/state.h>
+#include <libsbp/cpp/message_traits.h>
+
+namespace sbp {
+
+/**
+ * A convenience type alias for class member functions that accept SBP message callbacks
+ */
+template<typename ClassT, typename ArgT>
+using CallbackMemFn = void (ClassT::*)(uint16_t, uint8_t, const ArgT &);
+
+/**
+ * A helper function for calling a C++ object member function from a libsbp callback.
+ *
+ * This function is registered with libsbp as a callback and calls a specific member function
+ * of an object from that callback with the SBP message decoded. The member function pointer
+ * is encoded into the template parameter, and the instance of the object is passed in via
+ * the `context` parameter.
+ *
+ * @tparam MsgT The message type to decode the payload as
+ * @tparam ClassT The class type to call the function on
+ * @tparam func Pointer to the member function to call
+ *
+ * @param sender_id The decoded sender ID, is forwarded on to `func`
+ * @param len The length of the message, is forwarded on to `func`
+ * @param msg The raw message payload
+ * @param context Pointer to an instance of `ClassT` to call `func` on
+ */
+template<typename MsgT, typename ClassT, CallbackMemFn<ClassT, MsgT> func>
+inline void sbp_cb_passthrough(uint16_t sender_id, uint8_t len, uint8_t msg[], void *context) {
+  assert(nullptr != context);
+
+  auto instance = static_cast<ClassT *>(context);
+  auto val = reinterpret_cast<MsgT *>(msg);
+  ((*instance).*(func))(sender_id, len, *val);
+}
+
+namespace details {
+
+/**
+ * Recursive interface type for defining the interface functions for `MessageHandler`.
+ *
+ * These types define a virtual `operator()` for handling a specific SBP message type,
+ * as well as a function for registering it as a SBP callback.
+ *
+ * @note These types aren't meant to be used directly by application code.
+ *
+ * @tparam MsgType The type of SBP message to define an interface for
+ * @tparam OtherTypes Other types to recursively define interfaces for
+ */
+template<typename MsgType, typename... OtherTypes>
+class CallbackInterface : CallbackInterface<OtherTypes...> {
+ public:
+  CallbackInterface() = default;
+  virtual ~CallbackInterface() = default;
+
+  using CallbackInterface<OtherTypes...>::handle_sbp_msg;
+  virtual void handle_sbp_msg(uint16_t sender_id, uint8_t message_length, const MsgType& msg) = 0;
+
+ protected:
+  void register_callback(sbp_state_t *state, sbp_msg_callbacks_node_t nodes[]) {
+    sbp_register_callback(state,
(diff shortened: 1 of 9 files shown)


c/include/libsbp/cpp/message_handler.h
```diff
@@ -0,0 +1,146 @@
...
+template<typename... MsgTypes>
+ class MessageHandler : public details::CallbackInterface<MsgTypes...> {
+    static constexpr size_t kMsgCount = sizeof...(MsgTypes);
+
+    State &state_;
+    std::array<sbp_msg_callbacks_node_t, kMsgCount> callback_nodes_;
+
+  public:
+
+    explicit MessageHandler(State *state) : details::CallbackInterface<MsgTypes...>(), state_(*state), callback_nodes_() {
```
kmdade: Is there a reason the ctor argument isn't `State&`?

jbangelo: The `State` needs to be mutable, and our C++ guidelines say any non-`const` parameter should be passed as a pointer.

generator/sbpg/targets/resources/sbp_message_traits_template.h
```diff
@@ -0,0 +1,37 @@
...
+#define SBP_CPP_MESSAGE_TRAITS_H
+
+((*- for i in includes *))
+#include <libsbp/(((i))).h>
+((*- endfor *))
+
+namespace sbp {
+
+template<typename>
+struct MessageTraits { };
```
RReichert: is it allowed to have someone manually specify `MessageTraits<double> a;`? If not than static assure to make sure it never gets called unless specialized

c/include/libsbp/cpp/message_handler.h
```diff
@@ -0,0 +1,146 @@
...
+ * Classes that derive from `MessageHandler` need to implement
+ *   void handle_sbp_msg(uint16_t sender_id, uint8_t message_length, const MsgType& msg) const;
+ * for each `MsgType` in the list of message types given as template parameters.
+ *
+ * @note It should not matter if the class derives publicly or privately from `MessageHandler`
+ *
+ * @tparam MsgTypes List of SBP message types to register callbacks for
+ */
+template<typename... MsgTypes>
+ class MessageHandler : public details::CallbackInterface<MsgTypes...> {
```
RReichert: might I suggest adding a quick code snippet on how to use this class with the state class within the doxygen doc, with a quick snippet you get a quick grasp of how to use the class.

c/include/libsbp/cpp/state.h
```diff
@@ -0,0 +1,82 @@
...
+};
+
+class IWriter {
+ public:
+  virtual ~IWriter() = default;
+
+  virtual s32 write(const u8 *buffer, u32 buffer_length) = 0;
+};
+
+class State {
```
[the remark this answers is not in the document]
jbangelo: The name came from the `sbp_state_t` which it wraps into a C++ object. I'm open to alternate suggestions though.

c/include/libsbp/cpp/message_traits.h
```diff
@@ -0,0 +1,1026 @@
+/**
```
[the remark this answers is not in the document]
jbangelo: I don't think there's a double check in the CI. I think it's assumed as part of the manual review that only the relevant changes are made to the auto generated files.

c/include/libsbp/cpp/message_handler.h
jbangelo: Good suggestion. I don't think any documentation is generated for the libraries. I'll also add a real quick example in the example section.

generator/sbpg/targets/resources/sbp_message_traits_template.h
jbangelo: I'll remove the definition of this class so it's impossible to use it without specializing it. There are a couple of more complicated ways to prevent their use but this seems the most strait forward approach.

c/include/libsbp/cpp/message_handler.h
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * A helper function for calling a C++ object member function from a libsbp callback.
+ *
+ * This function is registered with libsbp as a callback and calls a specific member function
+ * of an object from that callback with the SBP message decoded. The member function pointer
+ * is encoded into the template parameter, and the instance of the object is passed in via
+ * the `context` parameter.
+ *
+ * @tparam MsgT The message type to decode the payload as
+ * @tparam ClassT The class type to call the function on
```
denniszollo: Is there a way to call the same function on multiple class types?  I find this is a common pattern for me (i.e do the same thing for any ephemeris message received)

jbangelo: Not with how this PR currently stands. The problem with that all of the message types are distinct and there's little to no information relating the content of similar messages. The closest thing we could get is defining a base class that handles each of the distinct messages transforms them into a common type and then calls another member function that works on the common type.

Something like this:
```
class CommonEphemerisHandler : public sbp::MessageHandler<EphemerisMessage1, EphemerisMessage2> {
  public:
    struct CommonEphemerisType { ... };

    virtual void handle_common_ephemeris(const CommonEphemerisType& eph) = 0;

    void handle_sbp_msg(uint16_t sender_id, uint8_t message_length, const EphemerisMessage1& msg) {
      CommonEphemerisType eph = convertEph1(msg);
      handle_common_ephemeris(eph);
    }

    void handle_sbp_msg(uint16_t sender_id, uint8_t message_length, const EphemerisMessage2& msg) {
      CommonEphemerisType eph = convertEph2(msg);
      handle_common_ephemeris(eph);
    }
};

class CustomEphemerisHandler : public CommonEphemerisHandler {
  public:
    void handle_common_ephemeris(const CommonEphemerisType& eph) { ... }
};

class OtherEphemerisHandler : public CommonEphemerisHandler {
  public:
    void handle_common_ephemeris(const CommonEphemerisType& eph) { ... }
};
```

c/include/libsbp/cpp/message_traits.h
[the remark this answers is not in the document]
RReichert: I'd probably say it would be worth adding that in so there is no misalignment between one source and its C++ representation

c/include/libsbp/cpp/state.h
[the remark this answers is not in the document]
RReichert: I would actually call it `SBP` since this class is how I think of SBP but it might be inappropriate. I'll leave it up to you, my comment was merely a suggestion

PR #740, opened 2019-10-21
Use common test targets module

woodfell:


c/test/CMakeLists.txt
```diff
@@ -6,34 +6,35 @@ if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
...
 if(MSVC)
-  target_include_directories(test_libsbp PRIVATE ${PROJECT_SOURCE_DIR}/include/libsbp/)
+  set(TEST_INCLUDES ${PROJECT_SOURCE_DIR}/include/libsbp/)
 else()
-  target_include_directories(test_libsbp PRIVATE ${PROJECT_SOURCE_DIR}/include/libsbp/)
+  set(TEST_INCLUDES ${PROJECT_SOURCE_DIR}/include/libsbp/)
   if(APPLE)
     # Some libraries are available in non-standard places on apple.
-    target_include_directories(test_libsbp PRIVATE /usr/local/include)
+    set(test_libsbp /usr/local/include)
```
RReichert: `test_libsbp` isn't being used elsewhere that I can tell

RReichert: `test_libsbp` isn't being used elsewhere that I can tell

c/test/CMakeLists.txt
```diff
@@ -6,34 +6,35 @@ if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
...
 
 FILE(GLOB generated_c_sources auto*.c)
-add_executable(test_libsbp check_main.c check_edc.c check_sbp.c ${generated_c_sources})
-
-target_link_libraries(test_libsbp PRIVATE ${TEST_LIBS})
-set_target_properties(test_libsbp PROPERTIES
-        C_STANDARD 99
-        C_STANDARD_REQUIRED ON)
 
 if(MSVC)
```
RReichert: Would this statement be equal to the one written below?
```
set(TEST_INCLUDES ${PROJECT_SOURCE_DIR}/include/libsbp/)

if (APPLE)
    # This is not a great way of doing this, but the proper cmake function
    # target_link_directories() was introduced in version 3.13 and we need to support
    # older versions for the moment. We can use target_link_libraries to pass arbitrary
    # linker flags, making sure that this instance is well protected and only applies
    # to a single platform. Don't use the extant link_directories() function since
    # that leaks the path to other targets.
    set(TEST_INCLUDES "-L/usr/local/lib")
endif()
```
Seems like the same things with extra steps.

RReichert: The other thing that doesn't seem just right is shouldn't it be:

```
list (APPEND TEST_INCLUDES "-L/usr/local/lib")
````

c/test/CMakeLists.txt
woodfell: Yep, that should be updating `TEST_INCLUDES`

c/test/CMakeLists.txt
woodfell: Yes, pretty much the same as the above

PR #741, opened 2019-10-25
Remove extra \\ from non-printable escaping

silverjam:


denniszollo:
I may have made a mistake on the escape sequences.

On other news, I think we need to replace the platform dependent "llu" and similar things with the portable format specifiers in <inttypes.h>

silverjam:
@denniszollo yeah, we can fix on the main PR

jkretzmer:
This looks good to me, but obviously will be a breaking change for Syrinx (which is fine).

PR #745, opened 2019-10-30
Add unit tests for Rust

jbangelo:
This PR adds generation of the test cases for Rust. The tests are built and run as a part of the typical build process for the rust bindings. I took the route of generating the tests instead of writing tests that read and process the YAML directly because of Rust's lack of reflection it seemed difficult to do a member wise comparison of the values without either learning to write a macro or autogenerating at least some part of the tests.

A couple of additional minor changes were made to fix issues found while implementing these tests:
- The `SBPMessage` was not marked as public, so it's fields could not be accessed from outside the crate.
- Several of the YAML specifications for tests contained values without a decimal point for float values. This meant that the generators for the tests would emit the literal values as integers. Rust doesn't allow these implicit type conversions, so the YAML was modified to contain values with a decimal point so the test generator knew to emit the correct floating point comparison.

The template for the tests were based on the existing C test template. Besides the syntactic changes the floating point comparisons were changed significantly to be more strict. The difference between the literal value and the unpacked value are compared against a a scaled multiple of the Epsilon of the floating point type. This should make sure that our unpacked floating point values are correct to within about 5 counts of the original value specified in the YAML. In addition the test generator now emits the expected floating point literals to 17 decimal places to ensure the literal is precise enough.


generator/sbpg/targets/resources/sbp_tests_template.rs
```diff
@@ -0,0 +1,76 @@
+//
+// Copyright (C) 2015-2018 Swift Navigation Inc.
```
silverjam: 2019?

generator/sbpg/targets/test_rust.py
```diff
@@ -0,0 +1,92 @@
...
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+"""
+Generator for c tests target.
```
silverjam: Rust?

generator/sbpg/targets/test_rust.py
```diff
@@ -0,0 +1,92 @@
...
+"""
+
+from sbpg.targets.templating import *
+from sbpg.targets.rust import *
+import base64
+
+TEST_TEMPLATE_NAME = "sbp_tests_template.rs"
+#CHECK_SUITES_TEMPLATE_NAME = "sbp_c_suites.h.j2"
```
silverjam: Delete if these aren't used

generator/sbpg/targets/test_rust.py
```diff
@@ -0,0 +1,92 @@
...
+  py_template = JENV.get_template(TEST_TEMPLATE_NAME)
+  with open(destination_filename, 'w') as f:
+    f.write(py_template.render(s=package_spec,
+                               description=package_spec.description,
+                               pkg_name=package_spec.package,
+                               include=package_spec.package.split('.')[1],
+                               filepath="/".join(package_spec.filepath) + ".yaml"))
+
+
+#def render_check_suites(output_dir, all_package_specs):
```
silverjam: Delete?

generator/sbpg/targets/test_rust.py
```diff
@@ -0,0 +1,92 @@
...
+def strEscape(value):
+    return "\"{}\".to_string()".format(value)
+
+def toStr(value):
+    return str(value)
+
+def modName(value):
+    return value.split('.')[1]
+
+JENV.filters['b64_decode'] = b64_decode
```
silverjam: Let's try to normalize the naming if these template functions

generator/sbpg/targets/resources/sbp_tests_template.rs
```diff
@@ -0,0 +1,76 @@
...
+use sbp::messages::SBPMessage;
+
+trait AlmostEq {
+    fn almost_eq(self, rhs: Self) -> bool;
+}
+
+impl AlmostEq for f32 {
+    fn almost_eq(self, rhs: Self) -> bool {
+        const ULP: f32 = 5.0;
+        ((self - rhs).abs() <= (std::f32::EPSILON * (self + rhs).abs() * ULP)) || ((self - rhs).abs() < std::f32::MIN)
```
silverjam: Not familiar with this formula, is it a standard?

generator/sbpg/targets/resources/sbp_tests_template.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+}
+
+impl AlmostEq for f64 {
```
silverjam: Any easy way to not have to repeat this in each test?

generator/sbpg/targets/test_rust.py
```diff
@@ -0,0 +1,92 @@
...
+TEST_TEMPLATE_NAME = "sbp_tests_template.rs"
+#CHECK_SUITES_TEMPLATE_NAME = "sbp_c_suites.h.j2"
+#CHECK_MAIN_TEMPLATE_NAME = "sbp_c_main.c.j2"
+
+def b64_decode(field):
+    print("Decoding '{}'".format(field))
+    b = base64.standard_b64decode(field)
+    return [ str(ord(ch)) for ch in b ]
+
+def stringType(value):
```
silverjam: If these are copied from the c test generator, can we move them to a common lib?

generator/sbpg/targets/resources/sbp_tests_template.rs
jbangelo: Fixed

generator/sbpg/targets/test_rust.py
jbangelo: Done here and in the C test generator.

generator/sbpg/targets/resources/sbp_tests_template.rs
jbangelo: Not sure how standard it is, I think I first found in on cppreference (https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon).

`EPSILON` is the distance between `1.0` and the next largest value that can be represented, we then scale it by a value that's on the same order of magnitude as the values (I've seen variations here where there is no scaling done, or the scaling is done by multiplying by the larger of the two values). The idea is that the scaled epsilon value is now about the same size as large as the smallest change in value at the same order of magnitude as the values. We then scale it again by some fudge factor (5 in this case). The final comparison is between the difference between the given values and this fudged smallest representable value.

This isn't always a good way to compare floating point values, but in this case we're trying to verify that the value has been decoded properly from the binary stream (and we aren't performing any mathematical operation on them). I think this comparison against the unit of least precision (ULP) makes sense.

generator/sbpg/targets/resources/sbp_tests_template.rs
jbangelo: Moved them into a `common.rs` file which is then used in each test case.

generator/sbpg/targets/test_rust.py
jbangelo: Moved into a `common.py` file and reused in both generators. I didn't look through all of the other modules, but the couple I did check didn't look like they had similar functions to de-duplicate.

generator/sbpg/targets/resources/sbp_tests_template.rs
silverjam: cool, thanks for the explanation

PR #747, opened 2019-10-31
Added optional serialize feature to rust crate

jbangelo:
Added a compile time feature, called `serialize`, which pulls in the base serde crate and derives all message types from the serde serialize and deserialize traits. In addition I updated the sbp crate to use the 2018 edition and fixed a few minor issues in the rust file templates.

### DIFF
--- a/generator/sbpg/targets/resources/sbp_messages_mod.rs
+++ b/generator/sbpg/targets/resources/sbp_messages_mod.rs
@@ -36,8 +36,8 @@ pub enum SBP {
 }
 
 impl SBP {
-    pub fn parse(msg_id: u16, sender_id: u16, payload: &mut &[u8]) -> Result<SBP, ::Error> {
-        let x: Result<SBP, ::Error> = match msg_id {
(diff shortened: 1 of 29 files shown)


jbangelo:
One thing that came to mind was to add tests of the JSON serialization feature. The YAML tests contain JSON payload of the message, we could add tests that serialize a message object and make sure the resultant JSON is equivalent to the pre-generated JSON.

jbangelo:
@silverjam Agreed that it would significantly increase the size/complexity of this PR. Was just bringing it up in case anyone thought it would be prudent to include from the get go. I'll consider it a "nice to have" for the future.

jbangelo:
Merging as is. It looks like travis somehow missed the update. There's one failure in the travis build, but it's the pythong/haskell speed comparison which occasionally fails anyways.

jbangelo:
@benjaminaltieri There isn't an example at the moment. I'll have to resurrect my `sbp2json` in rust.

PR #750, opened 2019-11-01
Added check for a couple of needed commands to build rust.

jbangelo:
Also added a basic README for the rust subdirectory with installation instructions.


jbangelo:
The Travis failure is the python vs Haskell speed comparison. Merging despite that

PR #751, opened 2019-11-04
Can't compare int and NoneType by default in python3.

martin-swift:


martin-swift:
gah. let me pull master.  must have moved.

PR #752, opened 2019-11-04
Cannot compare int and NoneType in python3

martin-swift:


martin-swift:
ok; what is package-lock.json?

silverjam:
@martin-swift it's the javascript package "lock" for npm, in order to create reproducible builds, npm "locks" the packages it actually installs with this file.  So, we should only commit this file if we add new deps to `package.json` (the npm package spec).

martin-swift:
> @martin-swift it's the javascript package "lock" for npm, in order to create reproducible builds, npm "locks" the packages it actually installs with this file. So, we should only commit this file if we add new deps to `package.json` (the npm package spec).

So what changes it?

silverjam:
@martin-swift when we gen javascript, we invoke npm, it looks like npm (because it's newer than when this `package-lock.json` was generated) is adjusting the syntax of the lock file to be more modern.  At some point we should commit these syntax changes with a recent version of npm, but probably as a separate PR?

PR #758, opened 2019-11-16
Example: rust message serialization plus trait to_frame method

benjaminaltieri:
just putting this up for fun, curious if it makes sense or if we should just stick with using `sbp::framer::into_frame` manually

### DIFF
--- a/generator/sbpg/targets/resources/sbp_messages_mod.rs
+++ b/generator/sbpg/targets/resources/sbp_messages_mod.rs
@@ -25,11 +25,13 @@ use self::unknown::Unknown;
 #[cfg(feature = "serialize")]
 use serde::{Serialize, Deserialize};
 use crate::serialize::SbpSerialize;
+use crate::framer::FramerError;
 
 pub trait SBPMessage: SbpSerialize {
     fn get_message_type(&self) -> u16;
     fn get_sender_id(&self) -> Option<u16>;
     fn set_sender_id(&mut self, new_id: u16);
+    fn to_frame(&self) -> std::result::Result<Vec<u8>, FramerError>;
 }
 
 #[cfg_attr(feature = "serialize", derive(Serialize, Deserialize))]
--- a/generator/sbpg/targets/resources/sbp_messages_template.rs
+++ b/generator/sbpg/targets/resources/sbp_messages_template.rs
@@ -92,6 +92,11 @@ impl super::SBPMessage for (((m.identifier|camel_case))) {
     fn set_sender_id(&mut self, new_id: u16) {
         self.sender_id = Some(new_id);
     }
+
+    fn to_frame(&self) -> std::result::Result<Vec<u8>, crate::framer::FramerError> {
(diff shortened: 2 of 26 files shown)


generator/sbpg/targets/resources/sbp_messages_mod.rs
```diff
@@ -25,11 +25,13 @@ use self::unknown::Unknown;
 #[cfg(feature = "serialize")]
 use serde::{Serialize, Deserialize};
 use crate::serialize::SbpSerialize;
+use crate::framer::FramerError;
 
 pub trait SBPMessage: SbpSerialize {
     fn get_message_type(&self) -> u16;
     fn get_sender_id(&self) -> Option<u16>;
     fn set_sender_id(&mut self, new_id: u16);
+    fn to_frame(&self) -> std::result::Result<Vec<u8>, FramerError>;
```
benjaminaltieri: might be able to make this consume the message as well if the original version of that behavior is preferred

rust/sbp/src/lib.rs
```diff
@@ -154,14 +154,16 @@ mod tests {
...
+
         let msg = crate::messages::system::MsgStartup {
             sender_id: Some(250),
             cause: 1,
             startup_type: 45,
             reserved: 0,
         };
 
-        let frame = crate::framer::to_frame(&msg).unwrap();
+        let frame = msg.to_frame().unwrap();
```
benjaminaltieri: the whole point I guess is just this little bit of syntactic sugar 🤷‍♂

benjaminaltieri: the whole point I guess is just this little bit of syntactic sugar 🤷‍♂

generator/sbpg/targets/resources/sbp_messages_mod.rs
jbangelo: I'd be in favor of leaving this as is. According to this page of the rust style guide I think an `into_` function like this would be acceptable, but I think the intention is that `into_` should be relatively lightweight if possible. https://doc.rust-lang.org/1.0.0/style/style/naming/conversions.html

generator/sbpg/targets/resources/sbp_messages_template.rs
```diff
@@ -92,6 +92,11 @@ impl super::SBPMessage for (((m.identifier|camel_case))) {
     fn set_sender_id(&mut self, new_id: u16) {
         self.sender_id = Some(new_id);
     }
+
+    fn to_frame(&self) -> std::result::Result<Vec<u8>, crate::framer::FramerError> {
+        let trait_object: &dyn super::SBPMessage = self as &dyn super::SBPMessage;
```
jbangelo: I think this is a bit cleaner?
```suggestion
        let trait_object = self as &dyn super::SBPMessage;
```

rust/sbp/src/lib.rs
jbangelo: Yup! I'm all for sugar that simplifies code!

PR #759, opened 2019-11-19
Add variances to SSR atmospheric messages. 

martin-swift:
In order to widelane fix in SSR mode, we need the variances for iono and tropo atmo delays (much like we have in OSR).  The only important file to look at is ssr.yaml; all the others are auto-generated language bindings (unless you are interested in verifying the generation happened correctly).


spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -142,6 +142,12 @@ definitions:
         type: s8
         units: 4 mm (add 0.252 m to get actual vertical wet delay)
         desc: Wet vertical delay
+    - stddev:
+        type: u8
+        units: |
+          modified DF389 scale; class is upper 3 bits, value is lower 5
+          stddev <= (3^class * (1 + value/16) - 1) mm
```
martin-swift: @pgrgich

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -153,6 +159,12 @@ definitions:
         type: s16
         units: 0.04 TECU
         desc: STEC residual
+    - stddev:
+        type: u8
+        units: |
+          modified DF389 scale; class is upper 3 bits, value is lower 5
+          stddev <= (3^class * (1 + value/16) - 1) / 10 TECU
```
martin-swift: @pgrgich

jbangelo:
Do we want to burn this message ID and use a new one? Any code that processes the old messages with this definition isn't going to be happy.

martin-swift:
> Do we want to burn this message ID and use a new one? Any code that processes the old messages with this definition isn't going to be happy.

? I don't think so; I have starling pull too that will fix all that code.

benjamin0:
@jbangelo Great point, we should make it a separate message number.

martin-swift:
oh, I see.  old versions....

jbangelo:
@martin-swift Exactly. I should have been more clear.

c/include/libsbp/ssr.h
```diff
@@ -114,8 +114,11 @@ but in units of TECU instead of m.
  * Troposphere vertical delays at the grid point.
  */
 typedef struct SBP_ATTR_PACKED {
-  s16 hydro;    /**< Hydrostatic vertical delay [4 mm (add 2.3 m to get actual vertical hydro delay)] */
-  s8 wet;      /**< Wet vertical delay [4 mm (add 0.252 m to get actual vertical wet delay)] */
+  s16 hydro;     /**< Hydrostatic vertical delay [4 mm (add 2.3 m to get actual vertical hydro delay)] */
```
martin-swift: why did an extra space get added here?

jbangelo: The generator inserts enough spaces so that there's 4 spaces between the longest variable and the comments, then aligns the remaining comments to that. https://github.com/swift-nav/libsbp/blob/791bf79fa092fe160a56b089dab9266a6417407c/generator/sbpg/targets/resources/sbp_messages_template.h#L48

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -435,7 +447,7 @@ definitions:
             desc: Array of STEC information for each space vehicle
 
  - MSG_SSR_GRIDDED_CORRECTION:
-    id: 0x05F0
+    id: 0x05FA
```
martin-swift: @jbangelo @silverjam - should I re-order the messages so they are monotonically increasing in the yaml?  If there is another message with that ID, will the build step throw an error?

martin-swift: if the answer is (no, yes), I am ready to push bindings

martin-swift: pushed...

jbangelo: I originally meant to deprecate the original message definition and add a new definition with the new fields. That would involve leaving the original definition in place, changing its name to be `MSG_SSR_GRIDDED_CORRECTION_DEP_A` and then adding the new message definition. That'll make sure that we don't end up reusing the same ID for something else down the line. It's a pretty remote chance of it happening but it could cause weird issues if it happens.

As for your question, I think one of the python tests will fail when it tries to store all the messages in a map by message ID and the map's length isn't what it expects.

jbangelo: And I wouldn't worry too much about reordering the messages unless you've got free time on your hands 👍

martin-swift: ah, yea - if I make a new message, certainly.  I remember there was some tool to print out all the IDs, no?
ok, I'll throw in the placeholder.

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -434,20 +446,13 @@ definitions:
...
-      It is typically equivalent to the QZSS CLAS Sub Type 9 messages
-    fields:
-        - header:
-            type: GriddedCorrectionHeader
-            desc: Header of a Gridded Correction message
-        - element:
-            type: GridElement
-            desc: Tropo and STEC residuals for the given grid point
-
+      This is here to make sure no one reuses this message ID.
```
martin-swift: @jbangelo

c/include/libsbp/ssr.h
```diff
@@ -114,8 +114,11 @@ but in units of TECU instead of m.
...
+  s8 wet;       /**< Wet vertical delay [4 mm (add 0.252 m to get actual vertical wet delay)] */
+  u8 stddev;    /**< stddev [modified DF389 scale; class is upper 3 bits, value is lower 5
+stddev <= (3^class * (1 + value/16) - 1) mm
```
pgrgich: This looks good.

c/include/libsbp/ssr.h
```diff
@@ -126,6 +129,9 @@ typedef struct SBP_ATTR_PACKED {
 typedef struct SBP_ATTR_PACKED {
   sv_id_t sv_id;       /**< space vehicle identifier */
   s16 residual;    /**< STEC residual [0.04 TECU] */
+  u8 stddev;      /**< stddev [modified DF389 scale; class is upper 3 bits, value is lower 5
+stddev <= (3^class * (1 + value/16) - 1) * 10 TECU
```
pgrgich: Why `* 10`? 1 TECU is ~16cm on L1, so we probably want resolution down to about 100th of a TECU. I think this means we should divide by 100, right?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -153,6 +159,12 @@ definitions:
         type: s16
         units: 0.04 TECU
         desc: STEC residual
+    - stddev:
+        type: u8
+        units: |
+          modified DF389 scale; class is upper 3 bits, value is lower 5
+          stddev <= (3^class * (1 + value/16) - 1) * 10 TECU
```
pgrgich: I guess this should be where I posted the question above. :)

c/include/libsbp/ssr.h
martin-swift: so this is the resolution the formula gives you (in TECU):
```
0
0.0625
0.125
0.1875
0.25
0.3125
0.375
0.4375
0.5
...
```
So divide that by 10, and you should have 1 mm resolution.  Seems like enough, ya?  The tropo is way overkill, but unless we think we need more range, I figured I would stick to mm for the units.

pgrgich: Ok, so maybe divide by 10 is more appropriate than divide by 100, but you've got a multiplication in there. If `class` and `value` are the encoded message and `stddev` is the decoded amount, I think we need to divide in order to get more resolution (e.g. 1, 2 -> 0.1, 0.2 with 0.1 resolution; if we multiple, 1, 2 -> 12, 20 with 10 resolution). Isn't that the case?

martin-swift: I disagree; let's say we calculate that our stddev is 0.06 TECU.  That's the smallest value the encoding will allow.   So in orion, I multiply by 10 before sending it; now the raw value that gets encoded is 0.6.   When I unpack, I divide by 10 again to get back to TECU.  So what is coming down in the message is TECU * 10, ya?  Or should i call it deci-TECU to eliminate this confusion?

pgrgich: All of what you said in that previous comment is correct. Either I've misunderstood the code or the code doesn't correspond, though. Isn't `stddev <= (3^class * (1 + value/16) - 1) * 10 TECU` the code that does the unpacking? I.e. shouldn't it do as you say: `When I unpack, I divide by 10 again to get back to TECU` and not multiply by 10?

martin-swift: ah, no - that is just a comment describing what is in that field.  The code that does the unpacking is here:
https://github.com/swift-nav/starling/blob/1bd3efca582e5b15761809e572ea03db5845e3a4/src/pvt_engine/ssr_unpackers.cc#L163

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -434,9 +446,23 @@ definitions:
...
     short_desc: Gridded troposphere and STEC residuals
+    desc: |
+      This message was deprecated when variances (stddev)
+      were added.
+    fields:
+        - header:
+            type: GriddedCorrectionHeader
+            desc: Header of a Gridded Correction message
+        - element:
+            type: GridElement
```
jbangelo: I didn't realize that the changes were in subtypes. To properly deprecate a message using subtypes you need to also change the names of the types and define a new type with the new fields.

So you'll need to deprecate the `GridElement`, `STECResidual`, and `TroposphericDelayCorrection` types by appending something like `_Dep_A` to their names and then change the names used here to use the deprecated types. Does that make sense?

martin-swift: Look good now?

jbangelo: Perfect! 🎉 

I'll stop making more busy work for you 😄

c/include/libsbp/ssr.h
martin-swift: alright - going ahead here since I'm out for the rest of the week.  Can always fix the comments later....

PR #760, opened 2019-11-19
Moved the rust example inside of the sbp crate

jbangelo:
Just noticed that this was a possibility. It seemed like a good idea to keep the sbp crate more or less self contained.


jbangelo:
@benjaminaltieri Good suggestion. I added a little note in this PR about that.

PR #763, opened 2019-11-27
Changes needed to libsbp for road runner [INFRA-35]

silverjam:
+ sbp2json: tools like ubx2sbp produce small amount of output (because of the small amount of data processed), in order for sbp2json to be responsive we need to disable buffering of stdin and stdout

+  In the handler code, since we spawn a thread, there's currently no way to "percolate" an exception that happens in the handler thread back to caller... the prevents code from doing things like reconnecting when a tcp connect dies... for example:
    - https://github.com/swift-nav/swiftcar/pull/42/files#diff-60beabf3ee14b8a53421a34aa6f05cdbR238 
    - or: https://github.com/swift-nav/swiftcar/pull/42/files#diff-60beabf3ee14b8a53421a34aa6f05cdbR678
   .
   These functions handle exceptions in order to know that a device is not responding and retry the connection as needed.

### DIFF
--- a/haskell/main/SBP2JSON.hs
+++ b/haskell/main/SBP2JSON.hs
@@ -28,7 +28,9 @@ encodeLine :: SBPMsg -> ByteString
 encodeLine v = toStrict $ encode v <> "\n"
 
 main :: IO ()
-main =
+main = do
+  hSetBuffering stdin NoBuffering
+  hSetBuffering stdout NoBuffering
   runResourceT $
     sourceHandle stdin
       =$= conduitDecode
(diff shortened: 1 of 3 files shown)


silverjam:
@denniszollo 

I don't think this will affect piksi_tools or the console for a couple of reasons:
- existing failures on this path would've cause the tool to hang and stop working, now it'll at least crash and stop
- the console uses the reconnect feature of the TCPDriver

We probably do need to test how this interacts with the serial port support though-- I added https://swift-nav.atlassian.net/browse/INFRA-51 to cover this.

PR #767, opened 2020-01-03
Added classes for handling specific frame types and any frame

jbangelo:
It was recently brought up that it might be helpful to have a way to handle any SBP message using the C++ classes. This PR adds the missing wrappers for handling raw frames of specific classes (corresponding to the `sbp_register_frame_callback()` C function), and for handling all incoming frames/messages (corresponding to the `sbp_register_all_msg_callback()` C function).

The interface for these were meant to be similar to the existing `MessageHandler` class. They differ in that they only define a single virtual function to be implemented instead of one for each message type. This was done since these classes would typically be used for lower level access than is typical, and the such application code would probably want access to the raw bytes of data instead of a type safe reference.

In addition to the code and example usage, a minor issue with the CMake files for the example was fixed as a part of this PR to help test the example.

### DIFF
--- a/c/include/libsbp/cpp/frame_handler.h
+++ b/c/include/libsbp/cpp/frame_handler.h
@@ -1,0 +1,191 @@
+/**
+ * Copyright (C) 2020 Swift Navigation Inc.
+ * Contact: Swift Navigation <dev@swiftnav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef SBP_CPP_FRAME_HANDLER_H_
+#define SBP_CPP_FRAME_HANDLER_H_
+
+#include <cassert>
+#include <array>
+
+#include <libsbp/cpp/state.h>
+#include <libsbp/cpp/message_traits.h>
+
+namespace sbp {
+
+/**
+ * A helper function for calling a C++ object member function from a libsbp frame callback.
+ *
+ * This function is registered with libsbp as a callback and calls the `handle_sbp_frame`
+ * member function of an object from that callback passing through all of the arguments.
+ * The instance of the object is passed in via the `context` parameter.
+ *
+ * @tparam ClassT The class type to call the function on
+ *
+ * @param sender_id The decoded sender ID, is forwarded
+ * @param msg_type The message type, is forwarded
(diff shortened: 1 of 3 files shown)


c/include/libsbp/cpp/frame_handler.h
```diff
@@ -0,0 +1,191 @@
...
+class FrameHandler {
+    static constexpr size_t kMsgCount = sizeof...(MsgTypes);
+
+    State &state_;
+    std::array<sbp_msg_callbacks_node_t, kMsgCount> callback_nodes_;
+
+  public:
+
+    explicit FrameHandler(State *state) : state_(*state), callback_nodes_() {
+        static constexpr std::array<uint16_t, kMsgCount> ids = { (0, sbp::MessageTraits<MsgTypes>::id)... };
```
jbangelo: Thanks to @kmdade for telling me about this little trick!!

c/include/libsbp/cpp/frame_handler.h
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef SBP_CPP_FRAME_HANDLER_H_
```
RReichert: trailing `_`, I believe this is a CLion thing

c/include/libsbp/cpp/frame_handler.h
```diff
@@ -0,0 +1,191 @@
...
+ *
+ *     void handle_sbp_frame(uint16_t sender_id, uint16_t msg_type,
+ *                         uint8_t payload_len, uint8_t payload[],
+ *                         uint16_t frame_len, uint8_t frame[]) {
+ *       // handle either GPS_TIME or POS_ECEF message types
+ *     }
+ * };
+ *
+ */
+template<typename ...MsgTypes>
```
RReichert: missing `@tparam for MsgTypes`

c/include/libsbp/cpp/frame_handler.h
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * Base type for defining classes that handle specific SBP Frames
+ *
+ * Application classes should derive from this class if they wish to handle
+ * entire frames of specific SBP messages with a member function. `FrameHandler`
+ * instantiates all of the callback nodes, and registers the member function with
+ * the given `sbp_state_t`.
+ *
+ * Classes that derive from `FrameHandler` need to implement
+ *   void handle_sbp_frame(uint16_t sender_id, uint16_t msg_type,
```
RReichert: you can enclose the following in a `@code` `@endcode` and the doxygen will look nice

jbangelo: Nice! I didn't know about those tags

PR #770, opened 2020-01-09
Changed the CMake to install all headers

jbangelo:
@lkloh noticed that we don't install the C++ libsbp headers, this PR is meant to fix that. I tried simply adding the headers to the existing list but the install command didn't put them in a `cpp/` subdirectory. This approach seemed cleaner than making a second variable and second install command.


lkloh:
> @lkloh noticed that we don't install the C++ libsbp headers, this PR is meant to fix that. I tried simply adding the headers to the existing list but the install command didn't put them in a `cpp/` subdirectory. This approach seemed cleaner than making a second variable and second install command.

thanks for fixing!

PR #772, opened 2020-01-14
Dockerfile for building libsbp binding.  Uses debian buster.

martin-swift:
cc: @jbangelo @ljbade @woodfell @schniedergers @mfine @benjamin0

### DIFF
--- a/Dockerfile-build
+++ b/Dockerfile-build
@@ -1,0 +1,77 @@
+# This is an image that should been able to generate Swift libsbp bindings.
+#
+# Developers should be in the top level directory of a clean clone of
+# libsbp, then type:
+#
+# docker run -v `pwd`:/mnt/workspace -i -t <this image>
+#
+# where <this image> is
+# 571934480752.dkr.ecr.us-west-2.amazonaws.com/libsbp-build:<tag>
+
+FROM debian:buster-slim
+
+# @Klaus - what is this?  
+ARG DEBIAN_FRONTEND=noninteractive
+RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
+
+RUN apt-get update && apt-get install -y \
+    git \
+    sudo \
+    software-properties-common \
+    apt-utils \
+    wget \
+    curl \
+    && mkdir /usr/share/man/man1
+
+# Toolchains
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    pandoc \
+    llvm \
+    nodejs \
+    npm \
+    mocha \
+    gradle \
+    texlive-science \
+    texlive-fonts-extra \
+    haskell-stack \
+    haskell-platform \
+    check \
+    cmake \
+    ccache \
+    pkg-config \
+    doxygen \
+    graphviz \
+    texlive-latex-base \
+    imagemagick \
+    libudev-dev \
+    uuid-dev
+
+RUN echo "deb http://ftp.debian.org/debian buster-backports main" >> /etc/apt/sources.list.d/buster-backports.list
+
+RUN apt-get update && apt-get install -y \
+    python-pip \
+    python3 \
+    python3-pip \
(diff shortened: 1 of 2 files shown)


martin-swift:
same file from this pull:
https://github.com/swift-nav/docker-recipes/pull/30

Dockerfile-build
```diff
@@ -0,0 +1,77 @@
...
+# libsbp, then type:
+#
+# docker run -v `pwd`:/mnt/workspace -i -t <this image>
+#
+# where <this image> is
+# 571934480752.dkr.ecr.us-west-2.amazonaws.com/libsbp-build:<tag>
+
+FROM debian:buster-slim
+
+# @Klaus - what is this?
```
benjamin0: Please remove comments before merging.

Dockerfile-build
```diff
@@ -0,0 +1,77 @@
+# This is an image that should been able to generate Swift libsbp bindings.
```
benjamin0: Does this need the standard header?

Dockerfile-build
```diff
@@ -0,0 +1,77 @@
...
+#
+# Developers should be in the top level directory of a clean clone of
+# libsbp, then type:
+#
+# docker run -v `pwd`:/mnt/workspace -i -t <this image>
+#
+# where <this image> is
+# 571934480752.dkr.ecr.us-west-2.amazonaws.com/libsbp-build:<tag>
```
benjamin0: These instructions look slightly different than the ones in the readme - can we have only one set?

Dockerfile-build
martin-swift: I don't know, but I can add one....

Dockerfile-build
martin-swift: Was testing the instructions and I think @schniedergers needs to add a new entry to ECR first so I can publish this.

Dockerfile-build
martin-swift: orion dockerfiles don't have headers, so I'm going to say "not needed" unless advised otherwise.

martin-swift: added

Dockerfile-build
silverjam: You can add it here: https://github.com/swift-nav/cloud-infra/blob/master/terraform/ci/ecr/ci_ecr.tf

martin-swift: I am looking at the comment, and I am confused about what I am supposed to do:
```
// To import an existing ECR repo, log into the AWS build account,
// then run 'terraform import 'aws_ecr_repository.build-ecr["hello-swift"]' hello-swift' as example for
// the hello-swift repo.
//
// If you need to remove state for a repo, use 'terraform state rm aws_ecr_repository.build-ecr["hello-swift"]'.
```
I don't think there is a ECR repo for libsbp-build yet;  also, I don't have authorization to assume the build role, and this is not for CI, per se.  This is for devs so they don't have to create the docker image from scratch (which takes a while).

I just want to create the libsbp-build repo in ECR so tag pushes under `docker-recipes` for libsbp-build will generate images that we can access.

silverjam: How about we push this to our public repo then?  https://hub.docker.com/u/swiftnav  The `docker-recipes` repo has support for this (see the `buildroot-base` docker image).  This was we don't need to deal with ECR at all.

silverjam: See https://github.com/swift-nav/docker-recipes/blob/master/Jenkinsfile#L121 -- particularly the `dockerhub` parameter should be `true` and the `ecr` param `false`:
```
                        imageBuild(
                            context: context,
                            name: "buildroot-base",
                            dockerfile: "buildroot-base/piksi_buildroot/scripts/Dockerfile.base",
                            dockerhub: true,
                            ecr: false,
                            tags: tagInfo.imageTags,
                            variants: tagInfo.publishVariants,
                        )
```

PR #773, opened 2020-01-14
Disallow copying of State

lkloh:
To support the coding guidelines that ban copying into functions.

RE: https://github.com/swift-nav/interleaver/pull/2/files#r366537476

### DIFF
--- a/c/include/libsbp/cpp/state.h
+++ b/c/include/libsbp/cpp/state.h
@@ -37,6 +37,10 @@ class State {
   IReader *reader_;
   IWriter *writer_;
 
+  State(const State&);
+  
+  State& operator=(const State&);
+
   static s32 read_func(u8 *buff, u32 n, void *ctx) {
     State *instance = static_cast<State *>(ctx);
     return instance->reader_->read(buff, n);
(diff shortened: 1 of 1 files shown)


c/include/libsbp/cpp/state.h
```diff
@@ -37,6 +37,10 @@ class State {
   IReader *reader_;
   IWriter *writer_;
 
+  State(const State&);
```
RReichert: The code you've added doesn't actually achieve the goal of disabling the copying of the class, it just tells the compiler that there should be copy constructor and assignment constructor defined somewhere, in the code. The compiler won't realize that there is an error until it performs the linking state of the build process, which is one of the last steps. The best way to achieve the goal of disabling of a copy constructor/assignment is to do the following.

```
State(const State&) = delete;
State& operator=(const State&) = delete;
```

After you make the change, try and write a unit test to verify that it works (you won't be able to commit the work because it will error out the compiler, but at least it is a way to check that your code works).

jbangelo: Agreed. Eventually we might want to enable move operations, but disallowing copies should implicitly disallow moves for now.

lkloh:
Thanks for finishing it off! Gonna merge the submodule updates since they passed

PR #775, opened 2020-02-10
Record total bytes read/written in python sbp driver

denniszollo:
I added this feature to enable a bits in bits out view for the console (https://github.com/swift-nav/piksi_tools/pull/1131)


silverjam:
Per slack disscusion we'll roll the counter at 32-bits and move the record keeping logic to the base class.

PR #776, opened 2020-02-19
Add JSON log iterator with functionality matching Framer class

pmiettinen:
Enabler for https://github.com/swift-nav/piksi_tools/pull/1137

### DIFF
--- a/python/sbp/client/handler.py
+++ b/python/sbp/client/handler.py
@@ -62,15 +62,15 @@ class Handler(object):
 
         messages = gen_messages()
         while True:
-            msg_and_metatdata = None
+            msg_and_metadata = None
             try:
-                msg_and_metatdata = next(messages, None)
+                msg_and_metadata = next(messages, None)
             except Exception as exc:
                 self._exception = exc
                 break
-            if msg_and_metatdata is None:
+            if msg_and_metadata is None:
                 break
-            msg, metadata = msg_and_metatdata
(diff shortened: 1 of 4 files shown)


python/sbp/client/handler.py
```diff
@@ -62,15 +62,15 @@ def gen_messages():
 
         messages = gen_messages()
         while True:
```
denniszollo: there are no changes to this file, so can we try and avoid changing whitespace?

denniszollo:
Can you comment as to how this is different than the existing JsonLogIterator and the reasoning behind needing a new iterator rather than updating the existing JsonLogIterator?  I couldn't tell by inspecting the code.

I think we tried to make the existing iterator capable of handling all three types of json serialized SBP; but perhaps that was a bad idea.

python/sbp/client/handler.py
denniszollo: oh I see now that you renamed the variable.

pmiettinen:
@denniszollo the main problem is the `yield` inside of `__next__`. It leads to a "we put an iterator into your iterator" kind of deal. I think it's somewhat mentioned in the class documentation also, see "Notes" here https://github.com/swift-nav/libsbp/blob/master/python/sbp/client/loggers/json_logger.py#L128

denniszollo:
replace with #779 due to merge conflict

silverjam:
closing since #779 was merged

PR #777, opened 2020-02-19
[SEN-138] Add message for accumulated wheeltick count

reimerix:
I'm still trying to generate the bindings but here's my proposal for a wheel tick message. While trying to implement the message that we had agreed upon, I noticed it has a fatal flaw: we really need timestamps for the latest tick count for every single wheel, especially if we don't have many ticks per revolution.

There are still some reserved bits in the `flags` field which could be used to communicate additional information, but IMHO the producing side should handle a lot of what was in our initial proposal, e.g. it should provide the logic to reverse the counting direction if the direction signal changes polarity.

### DIFF
--- a/python/tests/sbp/test_table.py
+++ b/python/tests/sbp/test_table.py
@@ -44,7 +44,7 @@ def test_table_count():
   Test number of available messages to deserialize.
 
   """
-  number_of_messages = 183
+  number_of_messages = 184
   assert len(_SBP_TABLE) == number_of_messages
 
 def test_table_unqiue_count():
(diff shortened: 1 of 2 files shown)


denniszollo:
If we want to go with this route, we could also consider adding a ENUM into the flags field into the existing MSG_ODOMETRY to indicate whether the units are accumulated distance or velocity (mm/s).  I was basically trying to map the typical auto ABS message directly into SBP.

GuillaumeDec:
> I'm still trying to generate the bindings but here's my proposal for a wheel tick message. While trying to implement the message that we had agreed upon, I noticed it has a fatal flaw: we really need timestamps for the latest tick count for every single wheel, especially if we don't have many ticks per revolution.
> 
> There are still some reserved bits in the `flags` field which could be used to communicate additional information, but IMHO the producing side should handle a lot of what was in our initial proposal, e.g. it should provide the logic to reverse the counting direction if the direction signal changes polarity.

I'm good with that. Any "status" (valid / invalid etc.) flag could be stuffed into the flags [2-7].

PR #780, opened 2020-02-24
Add library support to read and write SBP files

lkloh:


c/test/cpp/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+`json2sbp`, which can be installed by following the instructions [here](https://github.com/swift-nav/libsbp/tree/master/haskell#install-from-hackage).
+
+## Test data
+
+The json test data is stored and edited inside the `json_data` folder.
+
+The unit tests run on the `.sbp` files stored inside the `sbp_data` folder, which are directly generated from the `.json` files by running the following commands:
+```
+cd interleaver/tests/unit/
```
silverjam: typo, should be `c/test/cpp`?

lkloh: Yeah copy and paste thanks!

c/include/libsbp/cpp/sbp_stdio.h
```diff
@@ -0,0 +1,71 @@
...
+
+#ifndef SBP_CPP_STDIO_H
+#define SBP_CPP_STDIO_H
+
+#include <fstream>
+
+#include <libsbp/cpp/message_handler.h>
+
+namespace sbp {
+
```
lkloh: Following example of [state.h](https://github.com/swift-nav/libsbp/blob/master/c/include/libsbp/cpp/state.h) have implementation in header file too.

jbangelo: My original intention was that all of the C++ wrappers would be header only. I think that's fine to continue here, especially since these are nice small classes.

c/test/cpp/generate_sbp_data.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
jbangelo: Do we need to be able to regenerate the binary file?

c/test/cpp/test_sbp_stdio.cc
```diff
@@ -0,0 +1,107 @@
...
+        handler.write_message();
+        handler.write_message();
+      }
+    }
+  }
+};
+
+
+TEST_F(SbpStdioTest, ReadsSbpFiles) {
+  EXPECT_EQ(num_entries_in_file("gnss_data.sbp"), 3);
```
jbangelo: It looks like there's already a `test_data` directory already in the root of the repo. Could we reuse one of the files there to test this?

c/include/libsbp/cpp/sbp_stdio.h
```diff
@@ -0,0 +1,71 @@
...
+class SbpFileReader : public sbp::IReader {
+  public:
+    SbpFileReader(const std::string &file_path)
```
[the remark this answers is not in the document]
lkloh: Is the general policy to use `const char *` where possible (unless you need to string manipulation functions that `std::string` offers?) 

I was reading `Cases where you might prefer char* over std:string` here: https://www.geeksforgeeks.org/char-vs-stdstring-vs-char-c/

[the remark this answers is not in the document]
jbangelo: For general C++ programming I'd say `std::string` is almost always preferred. But since we're attempting to keep libsbp code more or less portable `const char *` is the lowest common denominator. Like I said before, it's difficult to imagine a situation where libsbp would be used where we're using `std::fstream` without concern but where we wouldn't want to use `std::string`.

c/test/cpp/test_sbp_stdio.cc
lkloh: So I looked into reusing one of the files there. Unfortunately, both the `short.sbp` and `long.sbp` files gave me an error when I read them in. This looks like a data error, not a problem with my code, as `sbp2json` also gave me this problem:
```
(py3) laykuanloh@Lays-MBP ~/Desktop$ cat long.sbp | sbp2json
  File "/Users/laykuanloh/py3/bin/sbp2json", line 6, in <module>
    module_main()
  File "/Users/laykuanloh/py3/lib/python3.8/site-packages/sbp/sbp2json.py", line 235, in module_main
    sbp_main(args)
  File "/Users/laykuanloh/py3/lib/python3.8/site-packages/sbp/sbp2json.py", line 211, in sbp_main
    m = msg_nojit.SBP.unpack(b)
  File "/Users/laykuanloh/py3/lib/python3.8/site-packages/sbp/msg.py", line 233, in unpack
    assert p.preamble == SBP_PREAMBLE, "Invalid preamble 0x%x." % p.preamble
AssertionError: Invalid preamble 0x0.
```

c/test/cpp/generate_sbp_data.sh
lkloh: Yes, it's helpful to have the json version (and use it to generate binary) if we ever want to change/add tests.

c/test/cpp/test_sbp_stdio.cc
silverjam: This is bizarre, the data files there work fine on my machine (and on the CI for sbp2json).

lkloh: Talked, there's a bug in Python 3.8 working with sbp2json. Filed as https://swift-nav.atlassian.net/browse/INFRA-114.

Unfortunately the files in `test_data` still don't work - nothing gets read in at all. At this point I think its better to keep the files I added instead, which have the added benefit of allowing user to easily make changes to the test data if they desire.

PR #782, opened 2020-02-26
File playback driver

pmiettinen:
## Motivation
Dennis' comment here https://github.com/swift-nav/piksi_tools/pull/1137#pullrequestreview-361312591

## Related
https://github.com/swift-nav/piksi_tools/pull/1138

## In practice
Playing back from SBP JSON log:
image

### DIFF
--- a/python/sbp/client/drivers/file_driver.py
+++ b/python/sbp/client/drivers/file_driver.py
@@ -10,6 +10,10 @@
 
 from .base_driver import BaseDriver
 
+import datetime
+import json
+import time
+
 
 class FileDriver(BaseDriver):
     """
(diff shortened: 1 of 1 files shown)


pmiettinen:
Not production worthy yet but good for initial comments if it makes any sense.

pmiettinen:
At least the console/serial_link needs some rearranging to enable the different combinations of the options.

pmiettinen:
If the timestamp metadata is not present, maybe TOW could be utilized instead. Would need to account for rollovers though.

silverjam:
@pmiettinen I believe we're fine to merge this

PR #785, opened 2020-03-10
Update haskell installation instructions

lkloh:
The README for Haskell currently instructs users to install a very old version of SBP version that causes [this bug](https://swift-nav.atlassian.net/browse/INFRA-117). Installing a higher version fixes the bug.

Comments about other places in the codebase/confluence to update instructions are welcome.

### DIFF
--- a/haskell/README.md
+++ b/haskell/README.md
@@ -10,13 +10,15 @@ Haskell client for Swift Binary Protocol (SBP).
 
 Available on [Hackage as `sbp`](http://hackage.haskell.org/package/sbp).
 
-The library supports building against Stackage LTS-15. To
-install from Hackage using `stack`:
+The library supports building against Stackage LTS-10. To install from Hackage using `stack`:
 
-    $ stack install --resolver lts-15.3 sbp # (LTS-15)
-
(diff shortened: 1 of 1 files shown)


mfine:
You can choose the SBP version to install without having to change the LTS version:

```
$ stack install sbp-2.6.3
```
Just annoying to continuously update the README to indicate that. Can rev the LTS here, but note we've generally been pinned to LTS-10 everywhere, so like publishing against LTS-15 would potentially cause problems elsewhere.

lkloh:
> You can choose the SBP version to install without having to change the LTS version:
> 
> ```
> $ stack install sbp-2.6.3
> ```
> 
> Just annoying to continuously update the README to indicate that. Can rev the LTS here, but note we've generally been pinned to LTS-10 everywhere, so like publishing against LTS-15 would potentially cause problems elsewhere.

I updated the README to recommend installing the latest sbp version listed on the [snapshots](https://www.stackage.org/package/sbp/snapshots) page, with an example on how to do so. Reverted LTS to LTS-10 recommendation.

mfine:
:+1:

PR #788, opened 2020-03-23
rust: nom parsing fix-up for eof error [INFRA-125]

silverjam:
Apparently nom has two cases were it can return "not enough data", there's an `Incomplete` return type and an EOF error.  I'm not sure in what cases it would return one vs the other, but the EOF error here would cause us the drop messages in the Rust version of sbp2json (which was surfaced while testing the Rust variant of json2sbp).

TODO:
- [x] Add unit test


silverjam:
> Good catch! Would we want to up the version number for this?

You mean make a new release? Yeah, I'll do that sometime this week.

PR #789, opened 2020-03-24
INFRA-137: wrap integer gets in parens

mfine:
Wrap the getting of integer in parens (puts were already wrapped). The first (used) integer array showed up and had problems with precedence around getting the array of integers.


silverjam:
:metal:

PR #792, opened 2020-04-10
add tile set id and grid id to ssr atmo messages

martin-swift:
Will need these to support multiple tiles. 
cc: @benjamin0


ljbade:
Should we look at aligning the grid definition to be closer to the final 3GPP message? (perhaps without the list option for the moment)

Specifically having a reference corner lat/lon, number of steps wide/high and the step distance in each dimension.

And ditch the RLE stuff with a bitmask.

Considering the hassle to change libsbp everywhere it's probably worth doing it at same time.

ljbade:
Also we might need to rev the message IDs so that things don't crash when people try to run this on old logs...

martin-swift:
for sure - this was just meant to get the ball rolling.  I didn't even generate the bindings yet.

martin-swift:
yes, 100% in favor of re-doing the grid def. message and the RLE stuff.

ljbade:
The change of the SBP SSR messages will be put on hold until we finish out the other internal components needed for multiple tiles. We will then have a better idea of exactly what changes we will need to make and can carefully roll it out. Otherwise we will end up with multiple iterations of these messages which is a lot of effort.

PR #803, opened 2020-05-12
rust: derive clone

silverjam:
So we can have a way to easily detach message objects from their source lifetimes, implement clone.
This enables us to implement something like:

https://gist.github.com/silverjam/43b543c58f37cb39575b01487f10779e


silverjam:
@jbangelo nothing jumps out at me, except maybe we should be `Copy` instead of `Clone`?  I picked `.clone()` since it's more explicit, and SBP structs can be 256 bytes (not sure what the threshold should be for picking copy over clone).

jbangelo:
@silverjam I think the log message contains a string (others might as well), which would prevent deriving `Copy` for it. I think only deriving `Clone` is probably good enough. Given the variable size of the messages I don't think it's in the spirit of the `Copy` trait too potentially be copying 255 bytes.

PR #807, opened 2020-05-21
Support building with ARMCC

gsmcmullin:


gsmcmullin:
Depends on merging https://github.com/swift-nav/cmake/pull/58 and updating cmake submodule pointer.

PR #812, opened 2020-06-01
README typos, pin deps for sbp2json in Cargo.lock

silverjam:


.gitignore
```diff
@@ -70,7 +70,7 @@ sbp_out.*
 .build/
 
 # rust
-Cargo.lock
+/Cargo.lock
```
jbangelo, line 73: Is this needed? Aren't you copying the `Cargo.lock` file into the `rust/` sub directory in the benchmark script?

silverjam: There's a top-level `Cargo.lock` that we need to ignore (I think) so that we don't have a `Cargo.lock` if libsbp is pulled in as a dependency from git.  Then there's a `Cargo.lock` that's in `rust/sbp2json/Cargo.lock` that we've checked in to pin deps, this is what gets copied to the benchmark build script's build directory.  The benchmark build script is a little misnamed at the moment... since it's used for building the binaries we distribute for Linux, as well as the benchmark process.

PR #813, opened 2020-06-08
Dzollo/python sender id whitelist

denniszollo:
required for https://github.com/swift-nav/piksi_tools/pull/1162

### DIFF
--- a/python/sbp/client/framer.py
+++ b/python/sbp/client/framer.py
@@ -51,7 +51,7 @@ class Framer(six.Iterator):
                  dispatcher=dispatch,
                  into_buffer=True,
                  skip_metadata=False,
-                 sender_id_filter=[]):
+                 sender_id_filter_list=[]):
         self._read = read
         self._write = write
         self._verbose = verbose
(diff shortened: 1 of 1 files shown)


python/sbp/client/framer.py
```diff
@@ -135,12 +137,15 @@ def _receive(self):
             if self._verbose:
                 print("crc mismatch: 0x%04X 0x%04X" % (msg_crc, crc))
             return None
-        msg = SBP(msg_type, sender, msg_len, data, crc)
-        try:
-            msg = self._dispatch(msg)
-        except Exception as exc:
-            warnings.warn("SBP dispatch error: %s" % (exc,))
-        return msg
+        if (len(self._sender_id_filter) == 0 or sender in self._sender_id_filter):
```
benjaminaltieri, line 140:
```suggestion
        if (not self._sender_id_filter or sender in self._sender_id_filter):
```

denniszollo: I could make that change; and agree it is a little clearer. 

To do I would have to change the semantics of passing an empty list if not filter to passing None if no filter, and also change the default value to sender_id_filter=None for the keyword arg.

denniszollo: Decided to rename the arg to include list and think that showing an empty list helps people know what to do should they choose to add a filter.

PR #815, opened 2020-06-30
Update BDS eph to include AODE/AODC parameters.

pgrgich:
One part of a change that will involve adding a new SBP message for BDS ephs. The problem is that the current one round trips by storing the AODE/AODC in the IODE/IODC parameters. This is confusing and we also want to use the IODE/IODC as a unique key (in a later PR).

### DIFF
--- a/c/include/libsbp/observation.h
+++ b/c/include/libsbp/observation.h
@@ -340,6 +340,40 @@ typedef struct SBP_ATTR_PACKED {
 } msg_ephemeris_qzss_t;
 
 
+/** Deprecated
+ *
+ * This observation message has been deprecated in favor of
+ * an ephemeris message with AODE/AODC.
+ */
+#define SBP_MSG_EPHEMERIS_BDS_DEP_A            0x0089
+typedef struct SBP_ATTR_PACKED {
+    ephemeris_common_content_t common;      /**< Values common for all ephemeris types */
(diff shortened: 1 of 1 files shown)


c/include/libsbp/observation.h
```diff
@@ -340,39 +340,75 @@ typedef struct SBP_ATTR_PACKED {
 } msg_ephemeris_qzss_t;
```
pgrgich, line 342: I'm sure I've done something against protocol in this PR. :)

mfine: Only a generated header file is changed here - did you change that by hand? Look at past PR's around new deprecated messages for guidance here.

pgrgich:
Since this PR changes `rtcm_msg_eph`, we'll have to pass that on to `msg_ephemeris_bds_t` and because that's in `msg_ephemeris_t`, `ephemeris_t` will also need to change. We'll also need to ensure that SIP is aligned (https://github.com/swift-nav/sip/blob/master/proto/sip.proto#L230-L278 and https://github.com/swift-nav/sip/blob/master/src/rtcm_decoder.cc#L485, at least).  https://github.com/swift-nav/libsbp/blob/master/HOWTO.md should be adhered to and the PR posted on the `libsbp` channel.

As a result of all these complications, I think I'll put it on the backburner for now.

PR #824, opened 2020-07-17
Guillaume/so 7/add leader group message

GuillaumeDec:
Add sbp definition of new leader/leading message for Solution Groups. 

SBP_MSG_GROUP_META.

As per Design Doc of new Starling Output MVP.
The point of this message is to facilitate consumption of output data,
in particular that of the fuzed Solution Group, aka "Fuzed Wagon".

Could be used for the GNSS (ARP) Wagon too. 

Had to update parsers for Rust & Javascript b/c a new type (array of u16 for the group_msgs field) was introduced.


spec/yaml/swiftnav/sbp/system.yaml
```diff
@@ -281,4 +281,44 @@ definitions:
...
+                   - 1: GNSS only
+                   - 2: GNSS+INS (Fuzed)
+                   - 3: Reserved
+             - 0-1:
+                 desc: Time source
+                 values:
+                   - 0: None (invalid)
+                   - 1: Reference epoch is start of current GPS Week
+                   - 2: Reference epoch is last PPS
+                   - 3: Reference epoch is Sensor Time Init
```
reimerix, line 320: Could we align this to the convention we use for `MSG_IMU_RAW`?
```suggestion
                   - 0: Reference epoch is start of current GPS week
                   - 1: Reference epoch is Sensor Time Init
                   - 2: Reference epoch is unknown
                   - 3: Reference epoch is last PPS
```

reimerix: Also, a time since last PPS probably won't ever be used to time-tag outputs of the fusion engine

spec/yaml/swiftnav/sbp/system.yaml
```diff
@@ -281,4 +281,44 @@ definitions:
...
+       It also lists the atomic contents (i.e. types of messages included) of the Solution Group.
+     fields:
+       - wn:
+           type: u16
+           units: weeks
+           desc: GPS Week Number
+       - tow:
+           type: u32
+           units: ms
+           desc: GPS time of week rounded to the nearest millisecond
```
reimerix, line 300: Needs different wording if we support "time since startup" timestamps

spec/yaml/swiftnav/sbp/system.yaml
```diff
@@ -281,4 +281,44 @@ definitions:
...
+     short_desc: Solution Group Metadata
+     public: true
+     desc: |
+       This leading message lists the time metadata of the Solution Group.
+       It also lists the atomic contents (i.e. types of messages included) of the Solution Group.
+     fields:
+       - wn:
+           type: u16
+           units: weeks
+           desc: GPS Week Number
```
reimerix, line 296: Needs different wording if we support "time since startup" timestamps

GuillaumeDec:
> Left a couple of comments - I think it would be good if the time source flags field would follow one convention throughout the SBP protocol, and since it was there first in `MSG_IMU_RAW`, I would tend to follow that convention, even if it admittedly makes less sense than your proposal.

After much consideration, I agree with your suggestion, let's stick to the legacy. Not ideal perhaps, but the `0` meaning "invalid" isn't necessarily a fundamental law either: one could argue that `0` means "default" state. And GNSS-referenced tow/tom are quite the default in this house :-)

PR #825, opened 2020-07-17
python: remove generation of the sbp.jit _payload_size method

silverjam:
This method was not used, and the genreated code wasn't correct when the
message included an unsized member like a string or a variable sized
array.


silverjam:
@GuillaumeDec FYI, noticed this while reviewing your PR

silverjam:
@GuillaumeDec since this is merged you'll need to merge and regenerate Python prior to merging your PR (#824)

PR #826, opened 2020-07-20
[HS-56] Add time offset message

reimerix:
This PR adds the `MSG_GNSS_TIME_OFFSET` to the SBP protocol. It contains the offset between a local time (unique to the sender ID emitting the messages) and the GNSS time scale.


reimerix:
@lloydmaza If a device has multiple sensors connected, it should be using the same clock to timestamp them. I don't think we would need more than one time offset by sender. If a device happens to timestamp different sensors with different clocks, it could also send out these messages with different sender ids.

@GuillaumeDec No other untracked files 🙂

PR #832, opened 2020-08-05
[SO-6] Add SOLN_META message for Fuzed Wagon

GuillaumeDec:
https://swift-nav.atlassian.net/browse/SO-6


spec/yaml/swiftnav/sbp/solution_meta.yaml
```diff
@@ -0,0 +1,222 @@
...
+                 desc: The type of sensor
+                 values:
+                   - 0: GNSS Position (see GNSSInputType)
+                   - 1: GNSS Velocity Displacement (see GNSSInputType)
+                   - 2: GNSS Velocity Doppler (see GNSSInputType)
+                   - 3: Odometry Ticks (see OdoInputType)
+                   - 4: Odometry Speed (see OdoInputType)
+                   - 5: IMU Sensor (see IMUInputType)
+                   - 6: Reserved
+                   - 7: Reserved
```
scarcanague, line 146: 2 more sensors seems a little tight here, Im wondering if we should rather to 0-3 for type of sensor, 4-5 for description and 6-7 for reserved

spec/yaml/swiftnav/sbp/solution_meta.yaml
```diff
@@ -0,0 +1,222 @@
...
+     short_desc: Solution Sensors Metadata
+     public: true
+     desc: |
+       This message contains all metadata about the sensors received and/or used in computing the Fuzed Solution.
+       It focuses primarly, but not only, on GNSS metadata.
+     fields:
+       - pdop:
+           type: u16
+           units: 0.01
+           desc: Position Dilution of Precision, as per last received DOPS,
```
scarcanague, line 171: I would change the phrasing a bit and keep it short:
`Position Dilution of Precision, as per last available DOPS from Starling GNSS engine`

spec/yaml/swiftnav/sbp/solution_meta.yaml
GuillaumeDec: Good point, Done!

java/src/com/swiftnav/sbp/client/MessageTable.java
```diff
@@ -173,6 +173,10 @@
 import com.swiftnav.sbp.settings.MsgSettingsReadByIndexDone;
 import com.swiftnav.sbp.settings.MsgSettingsRegister;
 import com.swiftnav.sbp.settings.MsgSettingsRegisterResp;
+import com.swiftnav.sbp.solution_meta.Gnssinputtype;
```
silverjam, line 176: We need to fix the java generator so that it correctly camel cases these embedded types.

GuillaumeDec: Good catch @silverjam ! Thanks!

GuillaumeDec: Done! 👍

PR #834, opened 2020-08-13
Update SSR atmo correction messages to allow multiple tiles.

ljbade:


spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -474,61 +391,335 @@ definitions:
...
+        - c1:
+            type: s32
+            units: 0.001 mm/s
+            desc: C1 polynomial coefficient for correction of broadcast satellite clock
+        - c2:
+            type: s32
+            units: 0.00002 mm/s^-2
+            desc: C2 polynomial coefficient for correction of broadcast satellite clock
+
+ - STECHeaderDep:
```
jbangelo, line 586: It might be pessimistic, but we should probably call this `DepA` just in case so we don't have to rename it if there's another revision to the header

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -474,61 +391,335 @@ definitions:
...
+        type: u8
+        desc: |
+            Update interval between consecutive corrections. Encoded
+            following RTCM DF391 specification.
+    - iod_atmo:
+        type: u8
+        desc: |
+            IOD of the SSR atmospheric correction
+
+ - GriddedCorrectionHeaderDep:
```
jbangelo, line 612: Same here, `DepA`

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -474,61 +391,335 @@ definitions:
...
+        type: u8
+        desc: |
+            IOD of the SSR atmospheric correction
+    - tropo_quality_indicator:
+        type: u8
+        desc: |
+          Quality of the troposphere data. Encoded following RTCM DF389
+          specifcation in units of m.
+
+ - GridDefinitionHeaderDep:
```
jbangelo, line 643: `DepA`

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -474,61 +391,335 @@ definitions:
...
+        - header:
+            type: STECHeaderDep
             desc: Header of a STEC message
         - stec_sat_list:
             type: array
             fill: STECSatElement
             desc: Array of STEC information for each space vehicle
 
- - MSG_SSR_GRIDDED_CORRECTION_NO_STD:
+ - MSG_SSR_GRIDDED_CORRECTION_NO_STD_DEP:
```
jbangelo, line 687: `DepA`

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -474,61 +391,335 @@ definitions:
...
         - header:
-            type: GriddedCorrectionHeader
+            type: GriddedCorrectionHeaderDep
             desc: Header of a Gridded Correction message
         - element:
             type: GridElementNoStd
             desc: Tropo and STEC residuals for the given grid point
 
- - MSG_SSR_GRIDDED_CORRECTION:
+ - MSG_SSR_GRIDDED_CORRECTION_DEP:
```
jbangelo, line 700: `DepA`

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -474,61 +391,335 @@ definitions:
...
+            type: GriddedCorrectionHeaderDep
             desc: Header of a Gridded Correction message
         - element:
             type: GridElement
             desc: |
               Tropo and STEC residuals for the given grid point (mean
               and standard deviation)
 
- - MSG_SSR_GRID_DEFINITION:
+ - MSG_SSR_GRID_DEFINITION_DEP:
```
jbangelo, line 715: `DepA`

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -81,12 +88,18 @@ definitions:
...
+    short_desc: Header for the MSG_SSR_GRIDDED_CORRECTION message.
     desc: |
-      The 3GPP message contains nested variable length arrays
+      The LPP message contains nested variable length arrays
       which are not suppported in SBP, so each grid point will
       be identified by the index.
     fields:
+    - tile_set_id:
+        type: u8
+        desc: Unique identifier of the tile set this tile belongs to.
```
jbangelo, line 99: Do we need to include the tile set ID in these messages?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -81,12 +88,18 @@ definitions:
...
+    - tile_id:
+        type: u8
+        desc: Unique identifier of this tile in the tile set.
```
jbangelo, line 102: Should this be `u16` to be closer to the LPP spec?

spec/yaml/swiftnav/sbp/ssr.yaml
ljbade: good idea

spec/yaml/swiftnav/sbp/ssr.yaml
ljbade: @jbangelo I was thinking about this last week. in LPP there is a single 16 bit value. However in SIP we have a 8 bit tile ID, and 8 bit tile group ID.

I was thinking we would combine the two to create a 16 bit ID for LPP?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -474,61 +391,335 @@ definitions:
...
+        - cols:
+            type: u16
+            desc: |
+              Number of steps in the longtitude direction.
+
+              See GNSS-SSR-CorrectionPoints field numberOfStepsLongtitude.
+        - bitmask_valid:
+            type: u8
+            desc: |
+              If 1 then bitmask indicates valid correction points, otherwise if
```
mfine, line 503: @ljbade so this is like a `boolean` on whether to use the bitmask?

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -474,61 +391,335 @@ definitions:
...
+              0 then all correction points are valid.
+        - bitmask:
+            type: u64
+            desc: |
+              Specifies the availability of correction data at the
```
mfine, line 508: @ljbade so this means we can only support a 64 points or a 8x8 grid at most? Or the other points are not true?

spec/yaml/swiftnav/sbp/ssr.yaml
benjamin0: @ljbade Is this resolved with `tile_set_id` and `tile_id`?

spec/yaml/swiftnav/sbp/ssr.yaml
ljbade: yeah I was trying to emulate what is in LPP, where the `bitmask` is an optional value...

alternatively I could copy what we do in SIP and remove this but invert the logic of bitmask, e.g. 0 = valid, 1 = invalid

spec/yaml/swiftnav/sbp/ssr.yaml
ljbade: This is the name size as in LPP. The spec does not specifically say but I presume if you don't set bitmask (e.g. bitmask_valid = false) then you can have more points

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -474,61 +391,335 @@ definitions:
...
+          region_size (deg) = 10 / region_size_inverse
+          0 is an invalid value.
+    - area_width:
+        type: u16
+        desc: |
+          grid height (deg) = grid idth (deg) = area_width / region_size
+          0 is an invalid value.
+    - lat_nw_corner_enc:
+        type: u16
+        desc: North-West corner latitdue (deg) = region_size * lat_nw_corner_enc - 90
```
benjamin0, line 662: nit: `latitude`

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -474,61 +391,335 @@ definitions:
...
+    - region_size_inverse:
+        type: u8
+        units: inverse degrees
+        desc: |
+          region_size (deg) = 10 / region_size_inverse
+          0 is an invalid value.
+    - area_width:
+        type: u16
+        desc: |
+          grid height (deg) = grid idth (deg) = area_width / region_size
```
benjamin0, line 658: nit: `width`

spec/yaml/swiftnav/sbp/ssr.yaml
ljbade: decided its easier to match SIP for the moment

spec/yaml/swiftnav/sbp/ssr.yaml
jbangelo: In SIP the IDs are 32 bits wide. I really think we should make these at least 16 bits. The tile set ID won't be changing much, just when/if we make a new tile layout.

PR #839, opened 2020-08-25
[SSR-17] Update SSR Tile IDs

jbangelo:
Update the SSR Tile IDs to be `u16`s instead to better match LPP.


ljbade:
@jbangelo is it worth going to full 32 bits? that would then match SIP. Unless we want to drop SIP to 16 bits, that way we won't have to worry about overflow when translating

jbangelo:
I don't see much need for a full 32-bits here, 8 bits is a bit small for long term growth but 16 bits would seem to give us more than enough room. I'm not sure if there's a real need for the addition 4 bytes in the message if they probably aren't going to be used.

ljbade:
@jbangelo fair enough

do you mind fixing up this bit of the SSR docs I just noticed was wrong for `STECSatElement.stec_coeff`
`units: C00 = 0.05 TECU, C01/C10 = 0.02 TECU/deg, C11 0.02 TECU/deg^2`

jbangelo:
Done @ljbade

ljbade (approved):

PR #841, opened 2020-08-27
generate bitfield macros [INFRA-265]

denniszollo:


c/include/libsbp/orientation.h
```diff
@@ -36,6 +36,17 @@ SBP_PACK_START
  * that time-matched RTK mode is used when the base station is moving.
  */
 #define SBP_MSG_BASELINE_HEADING 0x020F
+#define SBP_BASELINE_HEADING__FIX_MODE_MASK (0x7)
```
benjaminaltieri, line 39: any reason a double underscore is needed? I can see an argument for readability, but it doesn't work on my eyes 👀

denniszollo:
I was trying to come up with a way to separate message name from field
name.  Happy to remove if you think it fails at that goal.

silverjam:
> I was trying to come up with a way to separate message name from field name. Happy to remove if you think it fails at that goal.

The goal makes sense, but it's enough of a departure from standard naming conventions that I would prefer not to include this.  So would prefer just `_` instead of `__`.

c/include/libsbp/ext_events.h
```diff
@@ -35,6 +35,33 @@ SBP_PACK_START
...
 #define SBP_MSG_EXT_EVENT 0x0101
+#define SBP_EXT_EVENT_TIME_QUALITY_MASK (0x1)
+#define SBP_EXT_EVENT_TIME_QUALITY_SHIFT (1u)
+#define SBP_EXT_EVENT_TIME_QUALITY_GET(flags) \
+                             (((flags) >> SBP_EXT_EVENT_TIME_QUALITY_SHIFT) \
+                             & SBP_EXT_EVENT_TIME_QUALITY_MASK)
+#define SBP_EXT_EVENT_TIME_QUALITY_SET(flags, val) \
+                             do {(flags |= \
+                             ((val & SBP_EXT_EVENT_TIME_QUALITY_MASK) \
+                             << SBP_EXT_EVENT_TIME_QUALITY_SHIFT));} while(0);
```
silverjam, line 46: Per CERT C recommendations: https://wiki.sei.cmu.edu/confluence/display/c/PRE11-C.+Do+not+conclude+macro+definitions+with+a+semicolon -- a macro should not conclude with a semicolon

denniszollo: I think that was an accident; thanks for noticing!

denniszollo:
@jason; removed trailing whitespace and rebased a bit for clean rebase and merge on GH.

denniszollo:
I'm waiting to collect a bit more feedback; once we put these in we can't really rename them or take them out so I want to get it right.

c/include/libsbp/bootload.h
```diff
@@ -50,6 +50,29 @@ SBP_PACK_START
  * protocol version number.
  */
 #define SBP_MSG_BOOTLOADER_HANDSHAKE_RESP  0x00B4
+#define SBP_BOOTLOADER_HANDSHAKE_RESP_SBP_MAJOR_PROTOCOL_VERSION_NUMBER_MASK (0xff)
```
silverjam, line 53: We should think about alternate ways to generate these names since these generated names are really long.  We could consider on of the following:
- prefix the name with `FIELD_` or `BOOTLOAD_FIELD_`, and only generate the longer name if there's a conflict -- this makes the names not as discover-able via completion but much shorter to type, so not sure I really like this idea
- generate a enum with the bitfield values, the name would still be long, but it'd at least be split over the enum type e.g. `sbp_bootloader_handshake_resp.sbp_major_protocol_version_number_mask` -- the downside of this would be the need to cast to an appropriate type during use
- add field to the spec to specify a symbolic name for the field instead of generating it from the description

Additionally, we could improve the typing of this with the following change:
- generate a typedef'd enum that identifies each bitfield
- generate an (inlined) function which takes the bitfield, a bitfield enum, then returns the requested value (instead of using GET/SET macros here)

PR #843, opened 2020-08-28
json2json: maintain fields other than "data" [INFRA-266]

silverjam:
If we're running json2json and there are fields in the input struct
other than "data" we need to maintain those fields.  The use case here
is for HITL, where there's metadata (a "time" field) that needs to
be kept.


silverjam:
Merging, this seems correct, and no apparent regressions in the benchmark

PR #844, opened 2020-09-03
Improve SBP protocol PDF documentation [INFRA-280]

denniszollo:
I made quite a few changes to try to get an improved PDF document.

1. Add to the schema to avoid XXXInputType and all being called real messages according to docs and generated code
2. rename some linux draft messages so they look better on the PDF (not the best idea, but they were draft anyway!)
3. allow ctrl-f for MSG_ including underscore character
4. Fixup table widths to avoid running off of page
5. Fixup table widths to avoid columns running into eachother.


spec/yaml/swiftnav/sbp/solution_meta.yaml
```diff
@@ -124,13 +124,12 @@ definitions:
  #############################################
  # for GNSS sensors
  - GNSSInputType:
-     id: 0xFFE7
-     short_desc: Flags for a given GNSS sensor used as input for the fuzed solution.
+     short_desc: Instruments the physical type of GNSS sensor input to the fuzed solution.
      public: true
+     description_only: true
```
silverjam, line 129: How about `embedded_type: true` or `generate_docs: false`?

silverjam (changes requested):
It looks like the new `public` field caused the messages to be excluded from the Python SBP message table, so a unit test failed.

I don't think we should rename the Linux messages, this will cause build failures in dependent libraries piksi_buildroot_private.  Instead we can just drop the Linux process state messages since the feature was never officially released.

Otherwise, we'll need a companion change for PBRP, see [resource_monitor](https://github.com/swift-nav/piksi_apps/blob/master/app/resource_monitor/query_process_file_descriptor.c#L448) code for an example.

silverjam:
> 3. allow ctrl-f for MSG_ including underscore character

clapping-academy

cc @jaredw42

spec/yaml/swiftnav/sbp/solution_meta.yaml
denniszollo: Thanks so much for the review Jason.  

Yes I agree now that renaming linux is bad.  I will drop those messages from PDF for now instead.

I like "embedded_type" a lot and will go with that instead.  I couldn't think of a good name so went with an obviously bad one hoping a better one would come to me.

The unit tests are wrong since they call a message a message that isn't, so I will update them.

spec/yaml/swiftnav/sbp/linux.yaml
```diff
@@ -23,6 +23,7 @@ definitions:
  - MSG_LINUX_CPU_STATE:
     id: 0x7F00
     short_desc: List CPU state on the system
+    public: false
```
silverjam, line 26: One last tweak... there seems to be some cognitive dissonance between the meaning of "public" at the top level of the module and here on an individual message.  For the top level it just puts the message in a different location in the docs, for the message level it removes it from the docs.  To resolve this I would suggest we rename this to "hidden" and invert the logic, so `public: false` would become `hidden: true`.

denniszollo: that is a good idea, but I think I'd like to do a separate PR that does that all at once if this looks good as is.

silverjam: Ok, logged a new ticket to track: https://swift-nav.atlassian.net/browse/INFRA-281

silverjam (approved):

spec/yaml/swiftnav/sbp/linux.yaml
denniszollo: I am thinking "publish:" (same logic) or "hide_from_docs:" (opposite logic).

PR #845, opened 2020-09-03
Add bits for ins motion state

GuillaumeDec:
- Add bits in MSG_INS_STATUS for INS motion state

Sorry @denniszollo, forgor that it would go in after your PDF changes... how can I prevent the pdf from becoming ugly again with overlapping text etc.? Is the generator itself updated so I don't have to worry about it, or do I need to tweak a bit the generated latex before pdf-ization? 

(~ignore the checked-in pdf for now~ -- Surprisingly, I've re-built twice from scratch and the PDF doesn't seem to pick up my new flags in MSG_INS_STATUS 🤔 🤯 )
is there some kind of 'lock' in place?
@denniszollo mind giving it a spin? Basically just trying to `make docs` ... 🙏 
[update] looks like i found the culprit:
```
!pdfTeX error: pdflatex (file ebmr8): Font ebmr8 at 375 not found
```
gonna rebuild my Docker image with the latest Dockerfile and retry
This `texlive-fonts-extra` will hopefully do the trick (it contains the `ebmr8` font..)
image

[update2]
Got my whole new Docker image built after 3hrs 😅  but still doesn't work :-(
image


denniszollo (approved):
Looks good to me.

reimerix (approved):
Looks good

denniszollo:
It should be okay, we just need to generate the PDF.

reimerix:
@denniszollo I just tried to do that using the docker image, but the recently added `\usepackage[T1]{fontenc}` in swiftnav.sty causes the error messages that @GuillaumeDec was mentioning earlier. We might need to update the docker image to include whatever is missing?

silverjam:
@reimerix @denniszollo @GuillaumeDec I was able to fix the pdf generation in the docker container with this change https://github.com/swift-nav/libsbp/pull/845/commits/e0686bc2bef4792e788a9a5d89704a8e7198a8f9 but it introduces some layout issues with the new bits:

Screenshot from 2020-09-11 23-14-15

silverjam:
Merged the PDF generation fix into master and merged master to this branch.

@GuillaumeDec FYI, you typically don't need to build the docker container yourself, you can use the prebuilt version with `docker run  -v $PWD:/mnt/workspace -i -t swiftnav/libsbp-build:2020.08.28` -- if you need to play around with package modifications you can launch the pre-built container then just use `sudo apt update` and `sudo apt install ...` to install new packages or otherwise modify the system (which is faster than rebuilding the whole thing) -- once you've figured out the new changes then you can build the container locally, or you can use the `docker-recipes` repo to make jenkins build it for you.

PR #851, opened 2020-09-12
python: add reconnect logic to network write

silverjam:
Addresses [EN-12]

[EN-12]: https://swift-nav.atlassian.net/browse/EN-12


switanis:
Tested yesterday. Works well on Windows.

PR #853, opened 2020-09-14
fix tox tests for Python 3.5

silverjam:
We support Python 3.5 (this is what the console uses) but our tests don't currently succeed on 3.5, fix this...


silverjam:
@denniszollo merging so I can unblock testing on #858 -- let me know if you see any issues

PR #856, opened 2020-09-15
Dzollo/test new msgs

denniszollo:
Use the old build_test_data script to generate test for new messages.  Thank @mookerji !

Given we weren't testing new messages, new tests failing is probable and I see that on my desktop.  Either our spec in this library is wrong, or we aren't generating the new messages according to spec.

TODO: investigate and fix any failing tests either in the generator, in the auto-test generator, in the client libraries, or in the producer of the messages.


denniszollo:
Update: This PR is ready to go.  Adds some more unit tests to avoid issues with new messages.

PR #859, opened 2020-09-22
rust: add explicit to_string for SbpString

silverjam:
There seems to be a to_string method that gets picked up by default and this converts the SbpString to it's debug representation which is not what we want... instad add an explicit to_string implementation.


betog25 (approved):

silverjam:
Merging with failing tests since Haskell stackage servers are down

PR #860, opened 2020-09-26
Fix example SBP message in SBP Specification

dgburr:
The example message shown in Chapter 4 of the SBP Specification is incorrect - the message ID for MSG_BASELINE_ECEF is 523 = 20Bh, not 514 = 202h.

### DIFF
--- a/generator/sbpg/targets/resources/sbp_base.tex
+++ b/generator/sbpg/targets/resources/sbp_base.tex
@@ -189,10 +189,10 @@ contents.
  \par As an example, consider this framed series of bytes read from a
  serial port:
 \begin{verbatim}
-55 02 02 cc 04 14 70 3d d0 18 cf ef ff ff ef e8 ff ff f0 18 00 00 00 00 05 00 43 94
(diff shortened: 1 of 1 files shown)


dgburr:
Test code in case you want to double-check my checksum calculation:

```
import sbp.navigation
import binascii

msg = sbp.navigation.MsgBaselineECEF(sender=1228, tow=416300400, x=-4145, y=-5905, z=6384, accuracy=0, n_sats=5, flags=0)
b = binascii.hexlify(msg.to_binary())
pairs = [str(b[i:i+2]) for i in range(0, len(b), 2)]
print('\n'.join(pairs))
```

denniszollo (approved):
LGTM.  I'm surprised we didn't catch this after deprecating MSG_POS_ECEF during SBP 2.0 revamp.  Thanks so much for catching and fixing!

dgburr:
> LGTM. I'm surprised we didn't catch this after deprecating MSG_POS_ECEF during SBP 2.0 revamp. Thanks so much for catching and fixing!

I didn't catch it - the problem was actually reported by a customer

PR #864, opened 2020-09-30
[SR-146] Updates to solution meta message

reimerix:
This PR changes the solution meta message according to @switanis inputs:
- Remove the `n_sats` field
- Remove `alignment_status` field (the plan is to start using some bits of `MSG_INS_STATUS` for this in the future)
- Add `tow` field
- Replace TOW of last used GNSS pos/velocity fields with an `age_gnss` field.


denniszollo:
LGTM. I prefer that we squash down to two commits, one with change to yaml, one with the generated files.

GuillaumeDec (approved):
LGTM! 🎉

dgburr:
Which starling release will this changed message be introduced in?

dgburr:
The TOW field is needed to be able to match information from the fusion stream with information from the PVT stream (especially important if n_sats is going to be removed)

dgburr:
Is the new 'tow' field guaranteed to have a corresponding epoch in the PVT stream?  Or is this possible:

SBP_MSG_GPS_TIME_GNSS with tow = 999 (due to ublox rounding)
SBP_SOLN_META with tow = 1000

switanis:
I hope SOLN META TOW is the same as TOW in all other messages in the BEST POS group.
xxx_GNSS may have different TOW.

PR #866, opened 2020-10-03
c_bindings: for variable length arrays try [] arrays rather than [0]

denniszollo:


benjaminaltieri:
you beat me to it #867, but unfortunately doesn't look like a viable fix

PR #868, opened 2020-10-03
Replace zero is variable arrays with macro for custom replacements [DO NOT MERGE]

benjaminaltieri:
This would allow a user to define `SBP_VARIABLE_ARRAY_SIZE` before including `sbp.h`. This would allow for variable arrays to have either `0` or nothing to be used.

### DIFF
--- a/c/include/libsbp/file_io.h
+++ b/c/include/libsbp/file_io.h
@@ -52,7 +52,7 @@ typedef struct SBP_ATTR_PACKED {
   u32 sequence;      /**< Read sequence number */
   u32 offset;        /**< File offset [bytes] */
   u8 chunk_size;    /**< Chunk size to read [bytes] */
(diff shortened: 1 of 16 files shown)


c/include/libsbp/file_io.h
```diff
@@ -142,8 +142,8 @@ typedef struct SBP_ATTR_PACKED {
 typedef struct SBP_ATTR_PACKED {
   u32 sequence;    /**< Write sequence number */
   u32 offset;      /**< Offset into the file at which to start writing in bytes [bytes] */
-  char filename[0]; /**< Name of the file to write to */
```
silverjam, line 145: FWIW we could completely drop the data field from this struct, it's not used in our code... and I don't think it'd possible to use it?  Why does GCC even allow this to compile?

PR #869, opened 2020-10-04
Optional variable sized arrays

benjaminaltieri:
Introduces `SBP_DISABLE_VARIABLE_SIZED_ARRAYS` define to allow for variable sized arrays to be excluded.

In order to exclude, a user must define the above macro before including any header from `include/libsbp` either in source or via compiler define.

### DIFF
--- a/c/include/libsbp/bootload.h
+++ b/c/include/libsbp/bootload.h
@@ -75,7 +75,9 @@ SBP_PACK_START
 
 typedef struct SBP_ATTR_PACKED {
   u32 flags;      /**< Bootloader flags */
+#ifndef SBP_DISABLE_VARIABLE_SIZED_ARRAYS
   char version[0]; /**< Bootloader version number */
+#endif
 } msg_bootloader_handshake_resp_t;
 
 
@@ -125,10 +127,12 @@ on the right.
 * Deprecated.
  */
 #define SBP_MSG_BOOTLOADER_HANDSHAKE_DEP_A 0x00B0
+#ifndef SBP_DISABLE_VARIABLE_SIZED_ARRAYS
 
 typedef struct SBP_ATTR_PACKED {
   u8 handshake[0]; /**< Version number string (not NULL terminated) */
 } msg_bootloader_handshake_dep_a_t;
+#endif
 
 
 /** \} */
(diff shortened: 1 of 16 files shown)


c/include/libsbp/bootload.h
```diff
@@ -84,10 +86,12 @@ typedef struct SBP_ATTR_PACKED {
  * The host initiates the bootloader to jump to the application.
  */
 #define SBP_MSG_BOOTLOADER_JUMP_TO_APP     0x00B1
+#ifndef SBP_DISABLE_VARIABLE_SIZED_ARRAYS
```
silverjam, line 89: This isn't a variable sized array?

c/include/libsbp/bootload.h
```diff
@@ -112,23 +116,27 @@ typedef struct SBP_ATTR_PACKED {
  * and not related to the Piksi's serial number.
  */
 #define SBP_MSG_NAP_DEVICE_DNA_RESP        0x00DD
+#ifndef SBP_DISABLE_VARIABLE_SIZED_ARRAYS
 
 typedef struct SBP_ATTR_PACKED {
   u8 dna[8]; /**< 57-bit SwiftNAP FPGA Device ID. Remaining bits are padded
```
silverjam, line 117: Nor is this

c/include/libsbp/bootload.h
benjaminaltieri: Yeah looks like the generator expression isn't working as intended, will take a look

benjaminaltieri: moved the logic into a filter method since it was too hard to suss out in the template. should be fixed now

c/include/libsbp/bootload.h
benjaminaltieri: 🔨

silverjam:
Let's add a compile with `-Wpendantic -Werror` to the travis matrix?

benjaminaltieri:
> Let's add a compile with `-Wpendantic -Werror` to the travis matrix?

yea I don't think CI compiles when the `SBP_DISABLE_VARIABLE_SIZED_ARRAYS` define is enabled. this is mostly a proof of concept but if it looks worth pursuing we will need to add a step to specifically test this configuration

mfine:
Nice job!

PR #870, opened 2020-10-08
Per graphic design rec., switch to Roboto

silverjam:


switanis:
Can you generate the document using Roboto so we can see if it looks correctly?

silverjam:
> Can you generate the document using Roboto so we can see if it looks correctly?

https://github.com/swift-nav/libsbp/blob/57e4cc1076e6fb8f22a7434ded6d0a3bb0f904ee/docs/sbp.pdf

switanis:
I like it. Looks much better.

But the font size is too big for me.
image

What's the size there?
I'll also send this to Colin if he has any other suggestions.

silverjam:
@switanis adjusted the font size, let me know if this looks better: https://github.com/swift-nav/libsbp/blob/c283e73c69eb04077300fd5310b094ccfd35ca0b/docs/sbp.pdf

PR #880, opened 2020-11-05
[SEN-466] new flags for fixed fuzed rate

GuillaumeDec:
- Use highest two bits of `age_gnss` in `msg_soln_meta` to indicate whether or not its value can be used to retrive the time of measurement of the fuzed solution. Useful in the context of pre-alignment, where the latest GNSS solution is forwarded at fuzed rate and timestamped from the IMU samples. 
- Use the 5th bit of pos/vel flags to indicate whether the tow is time of measurement (== time of validity) or not. In almost all but the pre-alignment scenario, this is the case, so default value is `0`, which also preserves all backward compatibilities.


reimerix (approved):
LGTM

denniszollo:
I will review; I think we can reconsider the wording a bit

GuillaumeDec:
> I will review; I think we can reconsider the wording a bit

@denniszollo  I'm getting antsy at merging 😅 😉 😁 ...

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -335,8 +335,13 @@ definitions:
             type: u8
             desc: Status flags
             fields:
-              - 5-7:
+              - 6-7:
                   desc: Reserved
+              - 5-5:
+                  desc: Type of Reported tow
```
denniszollo, line 341: how about "TOW type"

spec/yaml/swiftnav/sbp/solution_meta.yaml
```diff
@@ -123,6 +123,10 @@ definitions:
      desc: |
        This message contains all metadata about the sensors received and/or used in computing the sensorfusion solution.
        It focuses primarly, but not only, on GNSS metadata.
+       Regarding the age of the last received valid GNSS solution, the highest two bits are Time status, indicating
```
denniszollo, line 126: Remove capital T

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -335,8 +335,13 @@ definitions:
...
+                  values:
+                      - 0: Time of Measurement
+                      - 1: Other
```
denniszollo, line 344: I feel like somewhere we need to describe what "Other" could possible be.  I suggest here in the intro paragraph. https://github.com/swift-nav/libsbp/pull/880/files#diff-1669e1e4c34cf7571c33296b08862bedd2a1ce937039283eb73a89b59631fcb7L13

denniszollo (changes requested):
1. Rename "Type of Reported tow" to "tow type"
2. Add exposition about the field at the top of the "navigation" yaml where we orient people about what is going on and what "Time of measurement" means and what "other" means.  Time of Measurement means only ONE thing, "other" could be a couple of things.

spec/yaml/swiftnav/sbp/navigation.yaml
GuillaumeDec: Thanks for your feedback Dennis! 🎉 🙏

GuillaumeDec:
> 1. Rename "Type of Reported tow" to "tow type"
> 2. Add exposition about the field at the top of the "navigation" yaml where we orient people about what is going on and what "Time of measurement" means and what "other" means.  Time of Measurement means only ONE thing, "other" could be a couple of things.

1. & 2. done ✅

denniszollo:
LGTM.

denniszollo (approved):

PR #884, opened 2020-11-25
Add async rust parser [INFRA-324]

notoriaga:
The `json2sbp`/`json2json` tests aren't passing, but I think that's because this version outputs the json in a different order than the old version. This version outputs the common fields first, then the actual message. I'm not sure what the order of the original is, it seems like the two get mixed somehow. I think that's because the default implementation of the `serde_json::Value::Object` is technically unordered. The original version would convert the messages into `Value`s then insert the extra fields. This version just directly serializes a struct, which should respect the order of the fields defined in the struct


generator/sbpg/targets/resources/sbp_messages_mod.rs
```diff
@@ -22,19 +22,19 @@ use self::(((p.identifier|mod_name)))::(((m.identifier|camel_case)));
...
     fn get_message_type(&self) -> u16;
     fn get_sender_id(&self) -> Option<u16>;
     fn set_sender_id(&mut self, new_id: u16);
-    fn to_frame(&self) -> std::result::Result<Vec<u8>, FramerError>;
+    fn to_frame(&self) -> std::result::Result<Vec<u8>, crate::FramerError>;
+    fn write_frame(&self, buf: &mut Vec<u8>) -> std::result::Result<(), crate::FramerError>;
 }
 
-#[cfg_attr(feature = "sbp_serde", derive(Serialize, Deserialize))]
+#[enum_dispatch::enum_dispatch(SbpSerialize, SBPMessage)]
```
notoriaga, line 36: `enum_dispatch` lets you call functions defined in `SbpSerialize` and `SBPMessage` directly on `SBP` objects without having to convert to a trait object, which is more convenient and it's static dispatch not dynamic

generator/sbpg/targets/resources/sbp_messages_mod.rs
```diff
@@ -22,19 +22,19 @@ use self::(((p.identifier|mod_name)))::(((m.identifier|camel_case)));
...
+#[cfg_attr(feature = "sbp_serde", derive(serde::Serialize), serde(untagged))]
```
notoriaga, line 37: Only derive `Serialize` because parsing of sbp json is done with manually based on the base64 payload

generator/sbpg/targets/resources/sbp_messages_template.rs
```diff
@@ -35,11 +34,12 @@ use super::(((i)))::*;
 (((m.desc|commentify)))
 ///
 ((*- endif *))
-#[cfg_attr(feature = "sbp_serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "sbp_serde", derive(serde::Serialize))]
 #[derive(Debug, Clone)]
 #[allow(non_snake_case)]
 pub struct (((m.identifier|camel_case))) {
     ((*- if m.is_real_message *))
+    #[cfg_attr(feature = "sbp_serde", serde(skip_serializing))]
```
notoriaga, line 42: Skip the `sender_id` because the json encoders add it as apart of the `common fields` added to all messages

rust/sbp/src/codec.rs
```diff
@@ -0,0 +1,354 @@
...
+        R: AsyncRead,
+        W: AsyncWrite,
+    {
+        let source = super::sbp::SbpCodec::decode_reader(input);
+        let sink = JsonEncoder::encode_writer(output);
+        source.forward(sink).await
+    }
+
+    #[derive(Debug, Deserialize)]
+    #[serde(untagged)]
```
notoriaga, line 119: `untagged` enums de/serialize without the inner names (`Input`/`Nested`) included

silverjam:
> The `json2sbp`/`json2json` tests aren't passing, but I think that's because this version outputs the json in a different order than the old version. This version outputs the common fields first, then the actual message. I'm not sure what the order of the original is, it seems like the two get mixed somehow. I think that's because the default implementation of the `serde_json::Value::Object` is technically unordered. The original version would convert the messages into `Value`s then insert the extra fields. This version just directly serializes a struct, which should respect the order of the fields defined in the struct

Order of fields would make sense for json2json to fail (I think?)-- the comparison there should probably use `jq` or something that doesn't a non-order sensitive compare.  For `json2sbp` I think it's comparing the binary SBP data that's produced?  If that's the case, order shouldn't matter there.

notoriaga:
I changed the tests to use serde to compare the json values, so order won't matter. it looks like they are equal now except for the treatment of floats that I mentioned earlier - https://travis-ci.org/github/swift-nav/libsbp/jobs/747307118#L1557

rust/sbp/src/codec.rs
```diff
@@ -344,11 +344,40 @@ pub mod json {
...
+        where
+            W: std::io::Write,
+        {
+            if value == 0.0 || value.abs() >= 0.1 && value.abs() <= 9_999_999.0 {
+                write!(writer, "{}", value)
+            } else {
+                write!(writer, "{:e}", value)
+            }
+        }
+    }
```
notoriaga, line 382: Not sure if there is a way to avoid this duplication without defining a float trait (or pulling one in)

silverjam: You could define a macro

generator/sbpg/targets/resources/sbp-cargo.toml
```diff
@@ -18,23 +18,33 @@ keywords = ["encoding", "parsing"]
 
 [features]
 sbp_serde = ["serde"]
-sbp2json = ["sbp_serde", "serde_json", "base64", "float-pretty-print", "dtoa"]
+json = ["serde_json", "serde", "base64"]
+blocking = ["futures/executor", "tokio-util/compat"]
```
silverjam, line 22: Note from Steve: the / syntax here enables a feature for that library

generator/sbpg/targets/resources/sbp_messages_mod.rs
```diff
@@ -22,19 +22,19 @@ use self::(((p.identifier|mod_name)))::(((m.identifier|camel_case)));
 ((*- endfor *))
 use self::unknown::Unknown;
 
-#[cfg(feature = "sbp_serde")]
-use serde::{Serialize, Deserialize};
```
silverjam, line 26: Note: need to import this is removed because we're referring to the full path instead

rust/sbp/src/parser.rs
```diff
@@ -0,0 +1,366 @@
...
+
+        crc.get() == *crc_in
+    }
+
+    fn is_not_preamble(b: u8) -> bool {
+        b != 0x55
+    }
+
+    fn parse_preamble(i: &[u8]) -> IResult<&[u8], &[u8], Error> {
+        bytes::streaming::is_a("\x55")(i)
```
silverjam, line 164: Note: code here is mostly the same with the exception that we're not using the streaming interface from nom, this adds "incomplete" as a possible return value, which integrates better with the tokio codec code

rust/sbp/src/codec.rs
```diff
@@ -0,0 +1,405 @@
...
+                    Ok(None)
+                }
+                ParseResult::Incomplete => {
+                    src.reserve(MAX_FRAME_LENGTH);
+                    Ok(None)
+                }
+            }
+        }
+
+        fn decode_eof(&mut self, buf: &mut BytesMut) -> Result<Option<Self::Item>> {
```
silverjam, line 54: This gets called when the source runs out of data

rust/sbp/src/codec.rs
```diff
@@ -0,0 +1,405 @@
...
+            match self.decode(buf) {
+                Ok(Some(frame)) => Ok(Some(frame)),
+                _ => Ok(None),
```
silverjam, line 57: This might affect the "round-trip" binary compatibility relative to the Haskell sbp2json version

rust/sbp/src/codec.rs
```diff
@@ -0,0 +1,405 @@
...
+            let mut ser = Serializer::with_formatter(dst.writer(), HaskellishFloatFormatter {});
+            output.serialize(&mut ser)?;
+            dst.put_slice(b"\n");
+
+            Ok(())
+        }
+    }
+
+    /// Provide Haskell style formatting. Output should be similar to: https://hackage.haskell.org/package/base-4.8.2.0/docs/Numeric.html#v:showFloat
+    struct HaskellishFloatFormatter {}
```
silverjam, line 376: Need to make this optional with a command line flag again

rust/sbp/src/codec.rs
```diff
@@ -0,0 +1,405 @@
...
+            )
+        }
+    }
+
+    impl Encoder<SBP> for SbpEncoder {
+        type Error = Error;
+
+        fn encode(&mut self, msg: SBP, dst: &mut BytesMut) -> Result<()> {
+            self.frame.clear();
+            match msg.write_frame(&mut self.frame) {
```
notoriaga, line 82: If the methods on `SbpSerialize` took `&mut BytesMut` or maybe `&mut impl bytes::BufMut` we could write the message directly to `dst`

jbangelo (commented):
Looks really good!

I'm a little confused by the inclusion of `async` code being the default and the blocking code path being an optional feature. I see why it's done that way, to reduce code reuse, but it still seems a little backwards. If I'm making a very simple tool to parse SBP and want to use the blocking interface, why should I be pulling in `async` specific libraries and paying for the size and compile time of those? Is this a typical approach in other libraries that have both blocking and `async` interfaces, to have the `async` interface be the default?

notoriaga:
> I see why it's done that way, to reduce code reuse, but it still seems a little backwards. 

Yeah that was the reasoning. If you build async-first you have [block_on_stream](https://docs.rs/futures/0.3.8/futures/executor/fn.block_on_stream.html) to convert to an iterator and [stream::iter](https://docs.rs/futures/0.3.8/futures/stream/fn.iter.html) to go the other direction.

I think either way you are going to pay some cost going one direction to the other. If you start async and expose a sync interface built on top, you're asking sync users to pay for the extra compile time like you mentioned. You also ask them to pay the cost of running an executor. In this case it's the [LocalPool](https://docs.rs/futures/0.3.8/futures/executor/struct.LocalPool.html) executor from the futures crate. That overhead should be pretty minimal, it's a small single threaded executor `... appropriate to poll strictly I/O-bound futures that do very little work in between I/O actions`. 

Starting sync first and building an async implementation on top is a little different. Imagine you have a natively sync -

```
pub fn iter_messages<R: Read>(input: R) -> impl Iterator<Item = Result<SBP>>;
```

I think building a `stream_messages` variant with the above mentioned `block_on_stream` would result in a pretty sub-optimal async implementation. This is because our sync `iter_messages` returns an iterator whose `next()` method could block (to perform a read). This is problematic because you shouldn't call blocking code from async code (because you'll block the executor, possibly resulting in deadlock). So really what we would have to do is wrap each call to `next()` in something like [spawn_blocking](https://docs.rs/tokio/0.3.5/tokio/task/fn.spawn_blocking.html) from tokio (not sure what the pure futures equivalent is). That will dispatch each call to next to a dedicated pool of threads for blocking calls. But really, if the user is already writing an async app, they shouldn't have to do that. A native `stream_messages` performs non-blocking reads and would not have that problem.

With that in mind, I picked async first because the cost sync users pay is mostly compile time, while the cost to async users in the other direction would mostly be runtime.

> Is this a typical approach in other libraries that have both blocking and `async` interfaces, to have the `async` interface be the default?

Sort of? Take [reqwest](https://github.com/seanmonstar/reqwest) for example. It's implementation is async, but it provides a blocking wrapper. But that's an http client so it's definitely more async in nature than what we have here. There is an old discussion on this topic from when async was being developed here - https://github.com/rustasync/team/issues/32.

That being said, I don't think it'd actually be that hard to have a native version of each. The `Encoder`/`Decoder` traits themselves are sync. It's just the `FramedRead`/`FramedWrite` parts that operate on `AsyncRead`/`AsyncWrite`. We could write `SyncFramedRead`/`SyncFramedWrite` which would operate the same way but on blocking IO, and then probably reuse everything we have setup with the encoding/decoding.

TLDR:
 - Wanted to keep things DRY
 - Both conversions sync -> async and async -> sync are not 0 cost
 - `async -> sync` cost appears to mostly be compile time while `sync -> async` appears to mostly be runtime

notoriaga:
I was able to get something together such that if you want sync, it pulls in no async stuff and vice versa - #896

I had to sort of cannibalize the futures_codec crate in order to make the changes to add a sync `FramedRead`, as well as get the feature gating right, but it seems to work

silverjam:
> I had to sort of cannibalize the futures_codec crate in order to make the changes to add a sync `FramedRead`, as well as get the feature gating right, but it seems to work

Is this something we could upstream to the futures_codec crate?

notoriaga:
> Is this something we could upstream to the futures_codec crate?

Hard to say, the changes seem kind of out of scope for `futures_codec` seeing as it's not really related to futures at all. Basically the changes would involve adding `blocking::FramedRead` and then setting up features such that we could pull in just that and the `Encoder`/`Decoder` traits and ignore all the async stuff.

I could fork the crate add try that. Even if we can't get it upstreamed it's still probably cleaner to have this stuff out of `sbp` itself

silverjam:
> > Is this something we could upstream to the futures_codec crate?
> 
> Hard to say, the changes seem kind of out of scope for `futures_codec` seeing as it's not really related to futures at all. Basically the changes would involve adding `blocking::FramedRead` and then setting up features such that we could pull in just that and the `Encoder`/`Decoder` traits and ignore all the async stuff.
> 
> I could fork the crate add try that. Even if we can't get it upstreamed it's still probably cleaner to have this stuff out of `sbp` itself

We can always send a PR to them and see what they say, even if they're not keen on it, they may have other suggestions.

silverjam:
Added https://swift-nav.atlassian.net/browse/INFRA-337 to cover updating nov2sbp.

generator/sbpg/targets/resources/sbp_messages_mod.rs
notoriaga: Changed it so we just generate the code that enum dispatch would generate. It brings down the compile time quite a bit

silverjam:
@jbangelo Following up: @notoriaga proposed changes to the futures-codec people to support our use case and they didn't seem interested-- to resolve this we created the `dencode` crate (which is a fork of futures-codec) but adds support for our use case

silverjam:
@jbangelo note, the CI failure is expected, we haven't yet updated the thresholds for the benchmarks

jbangelo (approved):
Sorry about the delay. Seems good to me! Thanks for undertaking all of this @notoriaga !!

silverjam (approved):

PR #885, opened 2020-12-01
Add generic system status reporting message [AMZ-168]

gsmcmullin:
Concrete message format suggestion for new status reporting system.
https://swift-nav.atlassian.net/wiki/spaces/~gareth/pages/1121910907/Status+reporting+requirements+Request+for+comments

If this is adopted in principle, more work will be required on the generators to show the value enumerations.
It would also be useful to generate lookup tables to convert these back to strings in the various language bindings.


dgburr (changes requested):
If it is planned to use this message as part of a safety mechanism (i.e. the expectation is that an external entity will monitor `MSG_STATUS_REPORT` messages to detect faults in Starling) then each message should be unique to allow detection of "stuck" faults (where the safety monitor keeps receiving an old "good" message instead of the latest "bad" one).  The `uptime` message nearly accomplishes this, but the problem is that it only has 1 second resolution, which would limit the fault detection time on the safety monitor side to 2 seconds.  

Therefore I would suggest either:
1. Changing the uptime value to millisecond resolution
2. Add a separate `counter` field which ensures that each message is unique

spec/yaml/swiftnav/sbp/system.yaml
```diff
@@ -134,6 +134,83 @@ definitions:
...
+              - 0-15:
+                  desc: Subsystem
+                  values:
+                      - 0: Primary GNSS Antenna
+                      - 1: Measurement Engine
+                      - 2: Corrections Client
+                      - 3: Differential GNSS Engine
+                      - 4: CAN
+                      - 5: Wheel Odometry
+                      - 6: Sensor Fusion Engine
```
woodfell, line 156: Is this meant to be covering all the starling subcomponent as well? If so then perhaps some more values - IMU, outputs

spec/yaml/swiftnav/sbp/system.yaml
```diff
@@ -134,6 +134,83 @@ definitions:
...
+              - 8-16:
+                  desc: SBP major protocol version number
+              - 0-7:
+                  desc: SBP minor protocol version number
+        - sequence:
+            type: u32
+            desc: Increments on each status report sent
+        - uptime:
+            type: u32
+            desc: Number of milliseconds since system start-up
```
woodfell, line 208: Is this mean to be time since starling startup or system? 2^32ms is only ~7weeks so wrapping isn't out of the questions particularly if the value comes from system time and starling was started some time after system start. Is it possible to use u64 here?

gsmcmullin: @dgburr What is your use case for wanting the system uptime in milliseconds?  This was intended to cover the requirement brought up by @switanis to report system uptime, currently reported in seconds in a MSG_LOG output.

To me, it seems interesting to know how many minutes, hours or days the system has been running, I can't imagine caring about sub-second resolution.

dgburr: @gsmcmullin: I was not involved in the definition of the uptime message (it seems to originate from the Amazon project)

PR #887, opened 2020-12-04
Dzollo/sender id notes

denniszollo:
@reimerix


denniszollo:
image

image

spec/yaml/swiftnav/sbp/observation.yaml
```diff
@@ -9,7 +9,11 @@
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
 package: swiftnav.sbp.observation
-description: Satellite observation messages from the device.
+description: Satellite observation messages from the device. 
+             The SBP sender ID of 0 indicates that observations
+             originated from a separate device such as a GNSS
```
dgburr, line 14: Suggest using phrase "remote observations" here since that would be more consistent with the terminology used in Swift Console as well as the Starling YAML config.

dgburr (approved):
LGTM, but suggested a minor clarification.

PR #891, opened 2020-12-11
docker: update docs for most recent tag [INFRA-333]

silverjam:
requires: https://github.com/swift-nav/docker-recipes/pull/130

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ performance on par with the Haskell and Rust variants.
 The quickest method to get going is to just pull a prebuilt copy from DockerHub
 (no guarantees on freshness):
 
-``docker run  -v $PWD:/mnt/workspace -i -t swiftnav/libsbp-build:2020.09.15``
+``docker run --rm -v $PWD:/mnt/workspace -i -t swiftnav/libsbp-build:2020.12.11``
 
 Check this [link](https://hub.docker.com/r/swiftnav/libsbp-build/tags) for newer tags.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,7 +108,7 @@ performance on par with the Haskell and Rust variants.
 The quickest method to get going is to just pull a prebuilt copy from DockerHub
 (no guarantees on freshness):
 
-``docker run  -v $PWD:/mnt/workspace -i -t swiftnav/libsbp-build:2020.09.15``
+``docker run --rm -v $PWD:/mnt/workspace -i -t swiftnav/libsbp-build:2020.12.11``
```
lkloh, line 111: Why not
```
docker run --rm -v $PWD:/mnt/workspace -i -t swiftnav/libsbp-build:latest-master
```
per https://snav.slack.com/archives/C03V0SS40/p1607125513112900?thread_ts=1606263555.109000&cid=C03V0SS40 
That would get the latest version right?

silverjam: @lkloh it's not recommended practice for docker images since that always pulls the latest version which could (in theory) break things by pulling in different versions of software if the docker build is not mostly "idempotent" between builds (and parts of the libsbp docker build are not since we changed over to pulling cmake from kitware's apt repo, which is always the latest cmake).  One of the benefits of docker images is reproducibility, so in general you should avoid operations that can change over time (unless it's something that's specifically designed to maintain backwards compatibility, like system package updates).

lkloh (approved):

PR #896, opened 2020-12-15
Better sync support

notoriaga:
This pulls in what we were using out of `futures_codec`, and adds `blocking::FramedRead`. So we can reuse all the encoding/decoding stuff, but backed by a sync `Read`.

Not really sure what the equality for `blocking::FramedWrite` would look like. For the async version it's centered around the futures [Sink](https://docs.rs/futures/0.3.8/futures/sink/trait.Sink.html) trait, and I'm not sure what the equivalent is for iterators (if it exists). So for now there is no blocking equivalent for the conversion functions

### DIFF
--- a/rust/sbp/src/codec/fuse.rs
+++ b/rust/sbp/src/codec/fuse.rs
@@ -1,0 +1,87 @@
+use std::{
+    io::Error,
+    marker::Unpin,
+    ops::{Deref, DerefMut},
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use futures::io::{AsyncRead, AsyncWrite};
+use pin_project_lite::pin_project;
+
+pin_project! {
+    #[derive(Debug)]
+    pub(crate) struct Fuse<T, U> {
+        #[pin]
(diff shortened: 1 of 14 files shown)


silverjam (approved):
Asked in the other PR, but it would be great if we could upstream this at some point, so we don't need to pull in a custom futures_codec lib.

rust/sbp/src/codec/fuse.rs
```diff
@@ -0,0 +1,87 @@
...
+use pin_project_lite::pin_project;
+
+pin_project! {
+    #[derive(Debug)]
+    pub(crate) struct Fuse<T, U> {
+        #[pin]
+        pub t: T,
+        pub u: U,
+    }
+}
```
jbangelo, line 19: A comment describing this struct would be helpful. It looks like it's just a wrapper around a Reader/Writer object and an Encoder/Decoder object? It might also be helpful to give the members more descriptive names like `io` and `codec`.

jbangelo (approved):

PR #897, opened 2020-12-16
add workaround for MinGW GCC bug [INFRA-329]

silverjam:
Works around a bug introduced in MinGW GCC 4.7 that causes it to ignore the `__attribute__((packed))` annotation on structures.  See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=52991 and https://wintermade.it/blog/posts/__attribute__packed-on-windows-is-ignored-with-mingw.html for reference.


woodfell (approved):

dgburr (approved):
Confirmed working by testing https://github.com/swift-nav/gnss-converters-private/pull/634

PR #899, opened 2020-12-19
c: toggle C++ on to fix configure error [INFRA-339]

silverjam:


silverjam:
@RReichert FYI, thanks for the fix

PR #900, opened 2020-12-21
[INFRA-340] Replace fail_unless with ck_assert_msg

reimerix:
This PR replaces most of the `fail_unless` usages with `ck_assert_msg`, the former has been deprecated for a use with a message in the check library.

Most of the format warnings in the generated C tests are gone. However, a couple of messages like

```
warning: format '%d' expects argument of type 'int', but argument 5 has type 'u64 {aka long unsigned int}
```

remain.

I noticed that some of the auto tests were not re-generated and deleted the respective tests. Hence the large number of changed lines.


silverjam:
A little concerned that 40k lines where deleted, which looks like a fair number of tests were deleted-- I think @jbangelo ran in to a similar issue with the rust tests?

jbangelo:
I did run into similar behavior where rust test cases weren't getting re-generated. I think I ended up deleting the rust test files that were being generated and didn't root cause the issue. Looks like that is what has been done here as well?

reimerix:
> Looks like that is what has been done here as well?

@jbangelo yes, exactly.

silverjam (approved):
@reimerix @jbangelo I looked through the Python code that generates the test files and what's generated (in auto generated tests) seems in line with what we have in git for YAML "test specification" that generate these files.  That is, each test specification that references an SBP module (e.g. `sbp.acquisition` or `sbp.tracking`) causes a file to be generated.  In the case of the `sbp.acquisition` there are 3 test spec YAMLs that reference this, so there are three files that are generated for Rust and C.  Rust in fact had some more files that were no longer being generated (I fixed this by deleting everything in git, then regenerating, and re-adding).

jbangelo (approved):
Changes look good to me. Thanks @silverjam for root causing the test generation issue. I guess we just need to be more careful about removing test code when test cases in the YAML are removed?

silverjam:
> Changes look good to me. Thanks @silverjam for root causing the test generation issue. I guess we just need to be more careful about removing test code when test cases in the YAML are removed?

I think a future improvement could be to make sure that all auto generated files get touched somehow.  By (for example) creating a manifest to track generated files and diff'ing that against what's on disk.

silverjam:
> > Changes look good to me. Thanks @silverjam for root causing the test generation issue. I guess we just need to be more careful about removing test code when test cases in the YAML are removed?
> 
> I think a future improvement could be to make sure that all auto generated files get touched somehow. By (for example) creating a manifest to track generated files and diff'ing that against what's on disk.

Created https://swift-nav.atlassian.net/browse/INFRA-347 to track

PR #904, opened 2021-01-17
Fix the documented offset for fixed length arrays

jbangelo:
The PDF docs weren't properly displaying the offset of fixed length arrays (and fixed length strings) nested in unbounded length arrays. This change fixes that.

Before:
image

After:
image


silverjam (approved):

denniszollo (approved):
Thanks Joseph!

PR #905, opened 2021-01-26
Migrate CI to github actions [INFRA-351]

notoriaga:
For the most part this tries to be a straight port of what travis did, with a few exceptions -

- CI is run on a per-language basis. Tags and pushes to master will trigger everything. You can also manually trigger any of the workflows from the `Actions` tab.
- All release artifacts have a `.zip` extension (previously a mix of `.zip` and `.tar.gz`)

This also includes a few miscellaneous changes to get everything working -

- Bumps `serialport` in the js library
- Fixes `test-javascript` make recipe 
- Benchmark threshold tweaks


.github/actions/build.bash
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
notoriaga, line 1: If you wanna write a js github action, you need to include your node_modules in git. `ncc` seems to be the recommend way of bundling it all together

notoriaga:
Test release - https://github.com/swift-nav/libsbp/releases/tag/v0.0.0-test

.github/actions/build.bash
silverjam: What's the benefit of creating these custom actions as opposed to using existing ones?

silverjam:
@notoriaga I'm only seeing the Rust workflows running, is there something we need to do to enable the other workflows?

notoriaga:
> @notoriaga I'm only seeing the Rust workflows running, is there something we need to do to enable the other workflows?

I was trying out using these as triggers for each language - 

```yaml
on:
  push:
    tags:
      - "*"
  pull_request:
    paths:
      - "$lang/**"
```

So the language specific workflows only run on prs that change that folder. I removed the travis badge from the Cargo.toml which is why it's building rust. This is obviously not how travis is doing it now, but figured it might be nice to save minutes. It runs everything on tags just to be sure. If we wanna run everything on all prs we just need to remove the `paths:` directive

.github/actions/build.bash
notoriaga: It seems to be the best way (that I've found so far) to keep things DRYish. For example, the release publish job is identical for haskell/rust. What I had using marketplace actions was roughly -

```yaml
  publish:
    if: startsWith(github.ref, 'refs/tags/v')
    needs:
      - build
    name: Publish github release
    runs-on: ubuntu-latest
    steps:
       - uses: actions/download-artifact@v2
         with:
           path: ./artifacts

       - name: Create Release
         id: create_release
         uses: actions/create-release@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           tag_name: ${{ github.event.ref }}
           release_name: ${{ github.event.ref }}
      
       - name: Upload artifacts
         uses: csexton/release-asset-action@v2
         with:
           pattern: |
             artifacts/**/*@(.zip|.tar.gz)
           github-token: ${{ secrets.GITHUB_TOKEN }}
           release-url: ${{ steps.create_release.outputs.upload_url }}
```

and I wanted to figure out a way to avoid copy-pasting this into `haskell.yaml` and `rust.yaml`. 

There is this concept of [composite steps](https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action), which in theory let you create new actions from a sequence of steps. The problem is you can't call other actions inside composite actions, you can only use `run` (i.e. script) steps. It's really odd, kind of like a language that lets you define functions, but only if they don't contain other non built-in functions. 

There is actually an issue about it here - https://github.com/actions/runner/issues/646 (what I'm referring to is the lack of `uses` support)

That all being said, not actually that against the copy/paste approach. The js actions are kind of a lot of overhead for not much functionality

silverjam:
> > @notoriaga I'm only seeing the Rust workflows running, is there something we need to do to enable the other workflows?
> 
> I was trying out using these as triggers for each language -
> 
> ```yaml
> on:
>   push:
>     tags:
>       - "*"
>   pull_request:
>     paths:
>       - "$lang/**"
> ```
> 
> So the language specific workflows only run on prs that change that folder. I removed the travis badge from the Cargo.toml which is why it's building rust. This is obviously not how travis is doing it now, but figured it might be nice to save minutes. It runs everything on tags just to be sure. If we wanna run everything on all prs we just need to remove the `paths:` directive

Ok, this makes sense, I think it'd also make sense to have a way to force everything to build, (which would be nice for this PR in particular, so we can see everything running).

.github/actions/build.bash
silverjam: > That all being said, not actually that against the copy/paste approach. The js actions are kind of a lot of overhead for not much functionality

Being DRY would be good, but yeah we need to consider the complexity/understandability trade-off.  Another way we could be DRY-ish is to trigger another workflow that has all of the things it acts on abstracted into variables.  The github release publish is an example:

https://github.com/swift-nav/gnss-converters-private/pull/625/files#diff-71cabc4177e41ea8f15a89eb65398fa8187739f68a2762706d84885cd8b2ddc3R210

If it's possible trigger a workflow that lives in another yaml file, maybe we could achieve some abstraction by having a library of abstract/parameterized workflows.  IOW, we'd be using workflows as our means of abstraction, rather than actions.

silverjam:
@notoriaga untagged myself for now, please re-tag me when there's new things to review

notoriaga:
@silverjam A few updates/questions

- Removed the custom actions
- Added `workflow_dispatch` as an event to all the workflows. This lets you trigger them manually. The option doesn't show until you have the workflows on the default branch for some reason. after you do though you can run the workflow on any branch
- I changed all the artifacts to `.zip` to be consistent with how we've been doing it elsewhere. A little worried this might break something, not sure if we have any automation around these release

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,38 @@
+name: Benchmarks
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
```
silverjam, line 6: default branch on libsbp is still called "master"

silverjam:
> 
> 
> @silverjam A few updates/questions
> 
>     * Removed the custom actions
> 
>     * Added `workflow_dispatch` as an event to all the workflows. This lets you trigger them manually. The option doesn't show until you have the workflows on the default branch for some reason. after you do though you can run the workflow on any branch

👍

>     * I changed all the artifacts to `.zip` to be consistent with how we've been doing it elsewhere. A little worried this might break something, not sure if we have any automation around these release

I don't think this will break anything

silverjam:
@notoriaga is this ready for review again?

notoriaga:
@silverjam Yeah should be ready for a look. Only thing I can't figure out is why the rust versions seem to run slower on gh actions compared to travis

.github/workflows/haskell.yaml
```diff
@@ -0,0 +1,67 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    tags:
+      - "*"
+  pull_request:
+    paths:
+      - "haskell/**"
```
jayvdb, line 11: Is it possible to also add `.github/workflows/haskell.yaml` to this list, so that modifications to the workflow trigger it ?  which also helps check that it is actually working on this PR

notoriaga: Great idea, added it to all them

.github/workflows/javascript.yaml
```diff
@@ -0,0 +1,21 @@
...
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    tags:
+      - "*"
+  pull_request:
+    paths:
+      - "javascript/**"
+      - .github/workflows/rust.yaml
```
jayvdb, line 12: "rust" ?

.github/workflows/javascript.yaml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    tags:
+      - "*"
+  pull_request:
+    paths:
+      - "javascript/**"
```
jayvdb, line 11: also needs package.json and the lock file which are not in this subdirectory

.github/workflows/python.yaml
```diff
@@ -0,0 +1,24 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    tags:
+      - "*"
+  pull_request:
+    paths:
+      - "python/**"
```
jayvdb, line 11: also needs requirements.txt and tox.ini in the root.

and ./scripts/ci_prepare_python.bash (and similar for the other helper scripts for other langs)

jayvdb (changes requested):

.github/workflows/javascript.yaml
```diff
@@ -0,0 +1,21 @@
...
+      - .github/workflows/rust.yaml
+jobs:
+  coverage:
+    name: Test
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Run tests
+        run: make test-javascript
```
jayvdb, line 21: looks like an exit status is being ignored somewhere because https://github.com/swift-nav/libsbp/pull/905/checks?check_run_id=2124840858#step:3:479 doesnt look very successful.

may be related to https://github.com/swift-nav/libsbp/issues/927

.github/workflows/c.yaml
```diff
@@ -0,0 +1,35 @@
...
+          sudo apt-get -qq install lcov
+
+      - name: Run tests
+        run: make test-c
+        env:
+          CMAKEFLAGS: -DCODE_COVERAGE=ON -DCMAKE_BUILD_TYPE=Debug
+
+      - name: Upload coverage to codecov.io
+        run: |
+          bash <(curl -s https://codecov.io/bash) -s c/build || echo "Codecov did not collect coverage reports";
```
jayvdb, line 35: IIRC, the bash codecov usually doesnt set an exit status, so the `|| ..` here is useless.

This upload isnt working; see https://github.com/swift-nav/libsbp/runs/2124966802#step:5:530

It seems codecov hasnt been working for a while in this repo, so perhaps that could be another task to get them all working in github actions after this PR.

.github/workflows/java.yaml
```diff
@@ -0,0 +1,21 @@
...
+    tags:
+      - "*"
+  pull_request:
+    paths:
+      - "java/**"
+      - .github/workflows/java.yaml
+jobs:
+  coverage:
+    name: Test
+    runs-on: ubuntu-18.04
```
jayvdb, line 16: there are three jvms on https://github.com/actions/virtual-environments/blob/ubuntu18/20210309.1/images/linux/Ubuntu1804-README.md .  It would be useful to print the JVM version that is being used for these tests.  (Under Travis, the version was explicitly being set to 1.8.0)

.github/workflows/c.yaml
jayvdb: It seems codecov<->github-integration isnt enabled for this repo.  It is possible to send reports without that integration being enabled.

Makefile
```diff
@@ -314,7 +314,7 @@ endif
 
 test-javascript:
 	$(call announce-begin,"Running JavaScript tests")
-	npm install; \
+	npm install
```
notoriaga, line 317: With the `\` I guess make treats `npm install; npm test` as one command. Without it, `npm install` fails properly

notoriaga: @silverjam should I bump serialport now that we know it fails properly?

silverjam: @notoriaga please, we should refresh the javascript stuff to whatever makes sense for modern JS development

.github/workflows/rust.yaml
```diff
@@ -0,0 +1,107 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    tags:
+      - "*"
+  pull_request:
+    paths:
+      - "rust/**"
```
jayvdb, line 11: Cargo.toml and its lock should be added.

jayvdb (commented):
Likely all workflows should run if test_data/** is modified

.github/workflows/haskell.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - .github/workflows/haskell.yaml
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
```
silverjam, line 16: We should pin this to ubuntu-1804 since the haskell binary is not fully static, otherwise we'll produce an artifact that'll only run on ubuntu 20.04 once `ubuntu-latest` switches:

```
❯ ldd sbp2prettyjson                                                                                                                                        (base)
        linux-vdso.so.1 (0x00007ffd42b6f000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f535776d000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f5357751000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f5357746000)
        libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007f5357741000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f535773b000)
        libgmp.so.10 => /lib/x86_64-linux-gnu/libgmp.so.10 (0x00007f53576b7000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f5357692000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f5357677000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f5357485000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f53578c9000)
 ```

Comment (silverjam) [reply .github/workflows/haskell.yaml:16]: Actually looks like this already happened: https://github.blog/changelog/2020-10-29-github-actions-ubuntu-latest-workflows-will-use-ubuntu-20-04/

File: README.md [hunk 13]
```diff
@@ -1,7 +1,5 @@
 ## Specification and Bindings for Swift Binary Protocol
 
-[![Build status][1]][2]
```
silverjam, line 3: So we can probably add this back later, but it would be for all languages, instead of just one global "CI" badge?

jayvdb (approved):
the important comments are resolved

scripts/ci_build_rust.bash
```diff
@@ -0,0 +1,30 @@
...
+
+set -ex
+
+VERSION="$(git describe --always --tags --dirty)"
+
+if [ "$RUNNER_OS" == "Linux" ] || [ "$RUNNER_OS" == "macOS" ]; then
+    BUILD_TRIPLET="$(cc -dumpmachine)"
+    ARTIFACT_NAME="sbp_tools-${VERSION}-${BUILD_TRIPLET}.zip"
+    EXECUTABLES=("sbp2json" "json2sbp" "json2json")
+    PACKAGE_CMD="zip ../../$ARTIFACT_NAME "${EXECUTABLES[@]}""
```
jayvdb, line 11: fwiw,..
```
❯ shellcheck scripts/ci_build_rust.bash

In scripts/ci_build_rust.bash line 11:
    PACKAGE_CMD="zip ../../$ARTIFACT_NAME "${EXECUTABLES[@]}""
                ^-- SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.
                                           ^---------------^ SC2027: The surrounding quotes actually unquote this. Remove or escape them.


In scripts/ci_build_rust.bash line 16:
    PACKAGE_CMD="7z a -tzip ../../$ARTIFACT_NAME "${EXECUTABLES[@]}""
                ^-- SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.
                                                  ^---------------^ SC2027: The surrounding quotes actually unquote this. Remove or escape them.

For more information:
  https://www.shellcheck.net/wiki/SC2027 -- The surrounding quotes actually u...
  https://www.shellcheck.net/wiki/SC2124 -- Assigning an array to a string! A...
```

silverjam (approved):
@notoriaga let's fix-up the PR message to include some more details and please tag the summary with a Jira and then...

squirrel

notoriaga:
> please tag the summary with a Jira and then...

@silverjam do you mean by adding the ticket to the pr name?

silverjam:
> 
> 
> > please tag the summary with a Jira and then...
> 
> @silverjam do you mean by adding the ticket to the pr name?

Yes

PR #911, opened 2021-02-04
[WIP] enable clang-format

woodfell:
I'm really not sure about this. Probably need to exclude the generated code at a minimum


denniszollo:
I feel like we could probably either adjust the templates or the .clang_format file so that no changes are required after generation.

PR #925, opened 2021-03-10
[INFRA-389]Add get_message_name function to SBPMessage .rs

john-michaelburke:
* Adds a get_message_name function to SBPMessage.

I'm not sure if returning `Unknown` is a better choice than an option/result. This also introduces `use serde_json::{json, Value}` to all messages which seems unfortunate.


generator/sbpg/targets/resources/sbp_messages_template.rs
```diff
@@ -84,6 +86,15 @@ impl (((m.identifier|camel_case))) {
 
 ((*- if m.is_real_message *))
 impl super::SBPMessage for (((m.identifier|camel_case))) {
+    fn get_message_name(&self) -> String {
+        if let Value::Object(obj) = json!(&self){
+            if let Some(key) = obj.keys().next() {
+                return key.to_string()
```
notoriaga, line 92: I think you should be able to do something like 

```rust
fn get_message_name(&self) -> String {
    "((m.identifier))".to_string()
}
```
you could also probably have it return `&'static str` instead

notoriaga: I see the haskell version has this helper https://github.com/swift-nav/libsbp/blob/38d77d28bab20cbca342e1bd63e2b9c67ee7dcef/generator/sbpg/targets/haskell.py#L73

Something like that might be useful, you'd define that in `rust.py` and then `"((m.identifier|helper_fn))"

john-michaelburke: Done. I had to use three parentheses to get it to work.

I'm not sure I understand completely. Are you thinking we could use the helper_fn for generating global constants?

john-michaelburke:
@notoriaga I'm not sure what would have caused this to timeout:
https://travis-ci.com/github/swift-nav/libsbp/jobs/489590670

Are ther

generator/sbpg/targets/resources/sbp_messages_template.rs
notoriaga: Oh I just wasn't sure that `identifier` would be the exact right case or whatever. If it works verbatim even better

notoriaga:
Hmm not sure, I'd probably just restart it and see if it happens again

notoriaga (approved):

jbangelo (approved):
Looks good to me! 🦀

silverjam (approved):
🥳

PR #926, opened 2021-03-10
[HS-131] Add INS status mode

reimerix:
This PR introduces a new mode value for `MSG_INS_STATUS` to indicate an unsafe fast start (e.g. using a seed from periodic storage instead of a seed from graceful shutdown).


woodfell (approved):
LGTM!

PR #928, opened 2021-03-16
Define a message for PPS time [STEP-231] [SEN-567] [SEN-578]

gsmcmullin:
The PPS time message contains the value of the CPU local time in microseconds at the moment a pulse is detected on the PPS input. This is to be used for synchronization of sensor data sampled with a local timestamp (e.g. IMU or wheeltick messages) where GNSS time is unknown to the sender.

A flags field is included. The first 2 bits are used to indicate the time uncertainty. The remaining bits are reserved for possible future expansion.

Internal design doc: https://swift-nav.atlassian.net/l/c/1BvX0WHm


dgburr (approved):
LGTM.  My only suggestion is that perhaps we should consider adding an "uncertainty" field which could potentially help the Fusion Engine to assess the trustworthiness of the timestamp.  What do you think @reimerix?

reimerix:
@dgburr I think that'd be a good idea. I don't think we'd make much use of a real uncertainty parameter, but maybe we could use a couple of the reserved bits for some "figure of merit" indication?

reimerix (approved):
LGTM. Do you want me to generate bindings and docs for this?

gsmcmullin:
@reimerix I'm quite happy for you to take ownership of the message format and add whatever uncertainty field you think is useful.  I'll do the implementation on the STEP once this lands.

dgburr:
A couple of thoughts:
* Perhaps we want to add a "sensor type" field? I can imagine that when integrating onto customer platforms we may have cases where we want to use PPS-based synchronisation for both IMU and WO input from two different sources which have their own concept of "local CPU time".
* Perhaps we want to consider making the message more generic to allow synchronisation to sources other than PPS, e.g. sensors which provide data with PTP timestamps
(cc: @larrychangSN)

richarddeurloo:
@dgburr For you second point, how do you see "synchronisation to sources other than PPS"? For PPS we know implicitly when the pulse happened in GNSS time. So we can derive the offset with GNSS time. For other pulses we do not have that implicit information.

I'm not very familiar with PTP. But isn't a PTP time-stamp a UTC time-stamp? Can we not treat this as an local time-stamp for which the customer platform needs to send a MSG_GNSS_TIME_OFFSET or MSG_PPS_TIME? Or, alternatively, introduce a new time status of the MSG_IMU_RAW such as "Full UTC time".

dgburr:
> @dgburr For your second point, how do you see "synchronisation to sources other than PPS"? For PPS we know implicitly when the pulse happened in GNSS time. So we can derive the offset with GNSS time. For other pulses we do not have that implicit information.
> 
> I'm not very familiar with PTP. But isn't a PTP time-stamp a UTC time-stamp? Can we not treat this as an local time-stamp for which the customer platform needs to send a MSG_GNSS_TIME_OFFSET or MSG_PPS_TIME? Or, alternatively, introduce a new time status of the MSG_IMU_RAW such as "Full UTC time".

When Starling is running on an automotive ECU we cannot assume that the sender of IMU or WO information has any knowledge of "GPS time" or a connection to the GPS PPS signal.  However we can assume that there is some kind of external time synchronisation mechanism (typically PTP, aka IEEE1588) which the sending ECU and the receiving ECU both have access to.  Therefore our timestamping mechanism for sensor data should assume that the timestamps are being provided relative to some kind of common source, which just happens to be the PPS signal in the case of a platform such as the STEP, but could potentially also be some other kind of technology.

reimerix (approved):
The message description could be a little more explicit that the local time used to timestamp the PPS pulse has to be generated by the same clock used for timestamping the IMU data and needs to follow the same rollover rules. Other than that nit, this LGTM!

dgburr (approved):
Super minor nit: suggest capitalising the 'l' in `desc: local time in microseconds` since this will eventually wind up in the specification PDF

PR #930, opened 2021-03-16
[INFRA-404] Remove python-jit

jayvdb:


generator/sbpg/generator.py
```diff
@@ -34,9 +34,6 @@ def get_args():
   parser.add_argument('--python',
                       action="store_true",
                       help='Target language: Python.')
-  parser.add_argument('--python_jit',
```
jayvdb, line 37: Is it worthwhile keeping this and throwing a meaningful error to the user?

python/sbp/msg.py
```diff
@@ -42,28 +42,7 @@
...
-  else:
-    # not found -> compile
-    try:
-      from sbp.jit import parse
-      parse.compile()
-      parse_jit = importlib.import_module('sbp.jit.' + parse_jit_name)
-    except ImportError:
-      return None
-  return parse_jit
+  return None
```
jayvdb, line 45: I've kept `try_import_jit` to avoid unnecessarily breaking any code elsewhere which might try to invoke this.  It could emit a UserWarning, and in time become an exception.

It also wouldnt hurt to do something similar to aide deprecation with `sbp.jit`, creating a very small `sbp/jit.py` that emits an error on import, and re-implements any names which can be safely re-implemented without the jit.

python/tox.ini
```diff
@@ -1,14 +1,13 @@
 [tox]
-envlist = py27-nojit,py35-nojit,py36-nojit,py37-nojit,py38-nojit,py36-jit,py37-jit,py38-jit
+envlist = py{27,35},py{36,37,38}-{aot,noextra}
```
jayvdb, line 2: Is `aot` still an appropriate name?

python/setup.py
```diff
@@ -204,13 +190,6 @@ def write_version_py(filename=VERSION_PY_PATH):
...
-          'jit': [
-              'numpy~=1.18',
-              'pybase64~=0.3',
-              'python-rapidjson~=1.0',
-              'numba==0.47',
-              'llvmlite==0.31',
-          ],
           'aot': [
               'numpy~=1.18',
               'pybase64~=0.3',
```
jayvdb, line 195: pybase64 appears to not be directly used any longer

generator/sbpg/generator.py
silverjam: No need to keep this around, the generator is only used by internal people and are not usually invoked by hand (just via the Makefile)

python/sbp/msg.py
silverjam: I like the UserWarning idea and creating a small sbp.jit with a similar purpose sounds reasonable

python/tox.ini
silverjam: No, maybe we rename this to sbp2json?

python/setup.py
```diff
@@ -204,13 +190,6 @@ def write_version_py(filename=VERSION_PY_PATH):
...
         scripts=['bin/sbp2json'],
         extras_require={
-          'jit': [
-              'numpy~=1.18',
-              'pybase64~=0.3',
-              'python-rapidjson~=1.0',
-              'numba==0.47',
-              'llvmlite==0.31',
-          ],
           'aot': [
```
silverjam, line 193: we need to remove/transform the `aot` extra too since it was essentially just the "pre-compiled" variant of sbp.jit-- should probably instead transform this to `sbp2json`

python/sbp/sbp2json.py
```diff
@@ -203,36 +193,25 @@ def sbp_main(args):
...
-                        m = dispatch_nojit(m)
-                        dump(args, m)
-                        consumed = header_len + m.length + 2
-                    except StreamError:
-                        # Framing failed, find next preamble
-                        consumed = 1
-                    except ValueError as exc:
-                        # CRC error, skip malformed message and find next preamble
-                        consumed = header_len + exc.malformed_msg.length
+            from construct.core import StreamError
```
john-michaelburke, line 196: Maybe break this import out if it is in a While True, I see that it was this way before though. Up to you.

john-michaelburke (approved):

python/sbp/sbp2json.py
jayvdb: I kept it as is so the block is kept intact for `git blame`

python/sbp/jit.py
```diff
@@ -0,0 +1,3 @@
+from warnings import UserWarning, warn
+
+warn("sbp.jit has been removed", UserWarning, stacklevel=1)
```
jayvdb, line 3: I'll add a test for this, or the other warning

silverjam (approved):
We'll need to update benchmark thresholds and get a clean CI run prior to pushing this, but I think we can wait until https://github.com/swift-nav/libsbp/pull/905 lands to do that

PR #932, opened 2021-03-19
python: fix call_git_describe in sbp.version

silverjam:
a previous PR amended this function to avoid the actual `git describe` call, restore the original behavior


silverjam:
@gsmcmullin Is there anything more we need to fix here, or was this just accidentally committed?

PR #933, opened 2021-03-22
[ESD-1897] Add support for Greenhill compiler

woodfell:
Greenhill compiler has the same struct packing attribute as gcc but needs an extra preprocessor test to enable it. Currently causes a build error using Greenhill compiler.


ductranc (approved):

RReichert (approved):
LGTM

PR #939, opened 2021-03-25
[STEP-217] Specify unit of mem_total in MSG_LINUX_SYS_STATE

IsakTjernberg:
### Description
Adds a comment that the unit of total system memory is measured in MiB, on the message MSG_LINUX_SYS_STATE.
The message is used in piksi_apps. Piksi_apps is used on both Piksi and STEP, in both cases total memory is reported as MiB.

I've separated the change to the spec and the generated files in separate commits for easier review, will squash on merge.


silverjam:
@IsakTjernberg python CI failure is fixed on master, so rebasing should fix this PR

silverjam (approved):

dgburr (approved):
LGTM

IsakTjernberg:
> @IsakTjernberg python CI failure is fixed on master, so rebasing should fix this PR

Thanks for the tip, was just about to ask.

PR #941, opened 2021-03-27
use binascii instead of plain python [INFRA-404]

silverjam:
+ strip out use of use of construct when deframing, this moves construct out of the deframing hot path

+ use binascii (which is a C extension module) to implement the CRC calculation: https://github.com/python/cpython/blob/master/Modules/binascii.c

+ Used sbp2json (skipping the `dispatch` method which parses into fields) to test the speed of just deframing. Speed of deframing (no JSON serializatoin) with master:
```
Benchmark #1: python -m sbp2json <test_data/benchmark.sbp
  Time (mean ± σ):     11.990 s ±  0.331 s    [User: 11.979 s, System: 0.011 s]
  Range (min … max):   11.441 s … 12.556 s    10 runs
```

Speed of deframing (no JSON serialization) with this branch:
```
❯ hyperfine 'python -m sbp2json --include 1 <test_data/benchmark.sbp'                                                                                                                                                (libsbp)
Benchmark #1: python -m sbp2json --include 1 <test_data/benchmark.sbp
  Time (mean ± σ):     961.3 ms ±  18.7 ms    [User: 953.2 ms, System: 8.1 ms]
  Range (min … max):   927.6 ms … 979.2 ms    10 runs
```


scripts/ci_benchmark.bash
```diff
@@ -61,7 +61,7 @@ sudo chown -R "$USER:$USER" "$PWD"
 
 rust_bins="$HOME/.docker-cache/cargo/work/release"
 
-mkdir bin
+mkdir -p bin
```
jayvdb, line 64: why `-p`, no parents are given.

python/sbp/msg.py
```diff
@@ -9,36 +9,30 @@
...
 
-import base64
+import binascii
 import copy
 import json
 import struct
-
-import construct
+try:
+  import pybase64 as base64
```
jayvdb, line 17: needs to be re-added to extras

scripts/ci_benchmark.bash
silverjam: Was failing when running the benchmark repeatedly, -p silently continues if the dir always exists

jayvdb (approved):

silverjam:
@dgburr Do you have a moment to try out this change?

silverjam:
> 
> 
> Shouldnt there be a inverse of [f3e62b1#diff-f0624b2d9bbb332561d971b7a21d1d172676d1a2937130ec6180077f6c5a4d50](https://github.com/swift-nav/libsbp/commit/f3e62b1d3b483c6e4d179eb40a92bcac3c077593#diff-f0624b2d9bbb332561d971b7a21d1d172676d1a2937130ec6180077f6c5a4d50) now? Some part of the benchmark rig needs to add the extra?

Indeed, I'll try adding this

python/sbp/sbp2json.py
```diff
@@ -202,16 +103,15 @@ def sbp_main(args):
...
-                    dump(args, m)
+                    m = sbp.msg.SBP.unpack(b)
+                    if not include or m.msg_type in include:
+                        m = sbp.table.dispatch(m)
+                        dump(args, m)
                     consumed = header_len + m.length + 2
-                except StreamError:
-                    # Framing failed, find next preamble
+                except (UnpackError, StreamError):
+                    break
```
dgburr, line 112: shouldn't there be a `consumed = 0` before the break??

dgburr (changes requested):

python/sbp/sbp2json.py
dgburr: Scratch that, I can see that it is breaking out of the loop so `consumed` is not used

silverjam:
Will need to tweak benchmarks once they post

silverjam:
> Will need to tweak benchmarks once they post

Follow up: since the benchmarks are based on multiples it seems the improvements in Python are within the 25% margin we allow for-- but we could probably do a follow up to confirm this and see if it's feasible to add some absolute benchmarks.

PR #944, opened 2021-03-30
package.json: Unpin binary-parser

jayvdb:
https://github.com/keichi/binary-parser/pull/21 was merged upstream

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,7189 @@
 {
   "name": "sbp",
   "version": "3.4.5",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1,8 +1,7189 @@
 {
   "name": "sbp",
   "version": "3.4.5",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
jayvdb, line 4: just noting this is changing the npm lock file version. It should be backwards compatible c.f. https://github.com/swift-nav/libsbp/issues/927

silverjam (approved):

package-lock.json
jayvdb: Matt has reported that this caused problems with npm 6.14.0.

PR #946, opened 2021-04-01
generator: mark intermediate versions as alpha [INFRA-436]

silverjam:
Not marking intermediate git versions misrepresents the state of the library, instead we should mark intermediate git versions as "alpha" in the generated code (and docs) so it's clearer that the current state of the code and documentaiton is not a released version of SBP.

Other updates:
+ Node is bumped to v12, this is required for the recent bump to binary-parser 1.7
+ Added in the hack @lkloh found, which invokes the pdf generation twice so tables are updated correctly

Related: https://github.com/swift-nav/docker-recipes/pull/151


test_data/benchmark_main.py
```diff
@@ -12,7 +12,7 @@
 # How much faster Rust should be than other implementations
 RATIOS_SBP2JSON = {
     "haskell": 2.06,
-    "python": 25.32,
+    "python": 18.8,
```
silverjam, line 15: @jayvdb sneaking this in, this is failing occasionally now, so there does appear to be a pretty good speed-up

jbangelo (approved):
Looks good to me. Would there be benefit to including the git commit hash in the version for intermediate commits?

silverjam:
> 
> 
> Looks good to me. Would there be benefit to including the git commit hash in the version for intermediate commits?

This is what I had at first, the problem is that those hashes/versions typically get thrown away once a PR is merged, they'd only be valid if we used merge commits.

test_data/benchmark_main.py
silverjam: Actually had to revert this again, seems on some runs it's going faster, and going outside the 25% range, on others it's not.

jayvdb: It was 2.74 before jit was removed, so this much higher number still concerns me a bit.

It guess it means the benchmark we have is not covering a similar load that @dgburr is using, as he saw perf return to almost the jit-era perf.  So perf regressions of a load like @dgburr is using can still silently occur and not trigger a CI failure.

silverjam: > 
> 
> It was 2.74 before jit was removed, so this much higher number still concerns me a bit.
> 
> It guess it means the benchmark we have is not covering a similar load that @dgburr is using, as he saw perf return to almost the jit-era perf. So perf regressions of a load like @dgburr is using can still silently occur and not trigger a CI failure.

I think @dgburr is not unpacking/parsing the messages (which goes through `construct`), that and JSON output are a big source of slowness in the Python sbp2json.  If I disable JSON output and unpacking I see similar speed improvements to what @dgburr was seeing.

dgburr: @silverjam: you are correct - I deliberately designed Drongo around the idea of postponing the `dispatch` step (i.e. when the type-specific fields get unpacked) for as long as possible for this exact reason.  So my profiling tests last week were focused mostly on parsing/deframing rather than dispatching/unpacking.  With my latest improvements to the STEP Log Postprocessing Utility I have had to do a lot of message dispatching much earlier than I did previously which has significantly degraded performance again.  It would be really nice to have JIT-accelerated message unpacking, or failing that, a rust-based parser which produces fully unpacked Python object types.

PR #952, opened 2021-04-17
Clarify units used in MSG_LINUX_SYS_STATE message

dgburr:
The documentation for MSG_LINUX_SYS_STATE says that the `pcpu` and `pmem` fields are percentages, however, from looking at test results and the code (https://github.com/swift-nav/piksi_apps/blob/master/app/resource_monitor/query_sys_state.c#L371) it is expressed as a fraction of 256 (i.e. equivalent to the MSG_LINUX_CPU_STATE and MSG_LINUX_MEM_STATE messages).


IsakTjernberg (approved):
Nice!

PR #960, opened 2021-04-23
SSR documentation clarification

scarcanague:
Clarifying  the code biases signal encoding and correcting a few nits


package-lock.json
```diff
@@ -626,7 +626,7 @@
     },
     "node_modules/assert/node_modules/util": {
       "version": "0.10.3",
-      "resolved": "https://registry.npmjs.org/util/-/util-0.10.3.tgz",
+      "resolved": "http://registry.npmjs.org/util/-/util-0.10.3.tgz",
```
richarddeurloo, line 629: Is this correct? I'd expect all addresses to be http**_S_**

package-lock.json
```diff
@@ -7523,7 +7522,7 @@
         },
         "util": {
           "version": "0.10.3",
-          "resolved": "https://registry.npmjs.org/util/-/util-0.10.3.tgz",
+          "resolved": "http://registry.npmjs.org/util/-/util-0.10.3.tgz",
```
richarddeurloo, line 7525: Same as above.

richarddeurloo (commented):
Just wondering about the updated hyperlink.

ljbade (approved):

package-lock.json
scarcanague: Hum that's a good point. This is automatically generated but Ill ping the `libsbp` channel to see if we can update the generator

PR #963, opened 2021-04-26
[ESD-2013] Add more test specs

woodfell:
The autogenerated tests don't cover a huge number of messages. A lot of them are quite old, some are organised as 1 spec file per module, some are 1 spec file per message type.

This PR tidies up the test specs by splitting up the per-module test specs to per-file specs and adding many more tests for more recent messages. The tests data was taken from an SBP recording from a test drive and covers all (most) of the messages produces and consumed by a recent starling-binary.

There is no attempt to cover every single SBP message, only the ones which are most likely to be used at the time of writing.

This PR covers just the test specs, the newly generated tests are moved over to a separate PR to keep this one easier to review - https://github.com/swift-nav/libsbp/pull/964


RReichert (approved):
Cant say I looked through the changes with a fine tooth combine, but given that these are only test codes, it shouldnt have an impact on production code.

silverjam (approved):
🤘

PR #965, opened 2021-04-26
[ESD-2009] Revert formatting changes

woodfell:
Some issue with a name clash appeared after merging https://github.com/swift-nav/libsbp/pull/954
Revert the change and defer until a larger rework is commissioned.

There have been a few commits since #954 was merged, it's probably best to review by comparing against c463fa9e0dc1e45478584e4492a77860c646cb67


HOWTO.md
```diff
@@ -92,7 +92,7 @@ Some thoughts to consider when adding a new message:
 It's highly recommended to use the docker container to run the release process,
 the docker container can be pulled from DockerHub and launched via this command:
 
-``docker run  -v $PWD:/mnt/workspace -i -t swiftnav/libsbp-build:2021.04.22``
```
silverjam, line 95: Should we keep this change?

Makefile
```diff
@@ -297,9 +297,7 @@ test-c:
 	cd $(SWIFTNAV_ROOT)/c; \
 	mkdir -p build/ && cd build/; \
 	cmake $(CMAKEFLAGS) ../; \
-	$(MAKE) clang-format-all; \
```
silverjam, line 300: Maybe leave the `do-all-tests` invocation and just remove `clang-format-all`?

README.md
```diff
@@ -110,7 +110,7 @@ Start [Docker desktop](https://docs.docker.com/docker-for-mac/).
 The quickest method to get going is to just pull a prebuilt copy from DockerHub
 (no guarantees on freshness) by running the following on your command line:
 
-``docker run --rm -v $PWD:/mnt/workspace -i -t swiftnav/libsbp-build:2021.04.22``
+``docker run --rm -v $PWD:/mnt/workspace -i -t swiftnav/libsbp-build:2021.04.02``
```
silverjam, line 113: Keep this change too

HOWTO.md
woodfell: Yeah can do. What do you want to do about `2021.04.22` - the only change in that image was to include `clang-format` but since it's not being used... Just keep using `2021.04.22` anyway since it's the latest?

silverjam: Yes, let's keep the latest, and when clang-format is reintroduced we won't need to worry about this

woodfell: Done

silverjam (approved):

PR #966, opened 2021-04-28
Unit documentation improvements [DEVINFRA-833]

richarddeurloo:


silverjam:
@richarddeurloo Do we still need this?

richarddeurloo:
@silverjam As a whole we don't need it anymore. But I'm wondering if part of the commits (at least a1a63a7) should be merged into master. If you don't mind, I'd like to keep it open for the moment. I'll check if part of it can be merged into master.

silverjam:
@richarddeurloo can we close this or move it into a Jira ticket?

silverjam:
@adrian-kong for next sprint, let's just undo the changes to the `public` field and the `gnss` module, I think we can port everything else over to current master

adrian-kong:
> @adrian-kong for next sprint, let's just undo the changes to the `public` field and the `gnss` module, I think we can port everything else over to current master

It looks like most of the things on master / conflicts already applies changes, don't think we need this unless we just want the desc to be bracketed or something

PR #967, opened 2021-04-28
add gps_time method

notoriaga:
Adds 

```rust
fn gps_time(&self) -> Option<Result<swiftnav_rs::time::GpsTime, crate::GpsTimeError>>;
```

to the `SBPMessage` trait behind a `swiftnav-rs` feature flag. Also reexports GpsTime as `sbp::messages::GpsTime`

- I went with `Option<Result<swiftnav_rs::time::GpsTime, crate::GpsTimeError>>`, but maybe the Result/Option should be swapped?
- Maybe should go with a default impl of `gps_time` that returns `None` instead of actually implementing a bunch of functions that return `None`. would make it easier to search for the impls and validate they are correct (also tests would be nice)
 - This crate already has a GpsTime struct which might be a little confusing - https://github.com/swift-nav/libsbp/blob/f2a1804d0253b288aa186960abdb782a38198fed/rust/sbp/src/messages/gnss.rs#L92


generator/sbpg/targets/rust.py
```diff
@@ -101,11 +122,37 @@ def parse_type(field):
...
+      has_tow = True
+    elif f.identifier == "wn":
+      has_wn = True
+
+  if has_tow and has_wn:
+    return GPS_TIME
+  elif has_tow:
+    return GPS_TIME_ONLY_TOW
+  else:
+    return 'None'
```
notoriaga, line 148: Not sure if this is the best way to determine which messages are time-aware/how that time is stored but it seems to work so far

silverjam:
>     * I went with `Option<Result<swiftnav_rs::time::GpsTime, crate::GpsTimeError>>`, but maybe the Result/Option should be swapped?

Should definitely be swapped, without swapping it makes it kind of hard to use with `?`, right?

>     * Maybe should go with a default impl of `gps_time` that returns `None` instead of actually implementing a bunch of functions that return `None`. would make it easier to search for the impls and validate they are correct (also tests would be nice)

👍

>     * This crate already has a GpsTime struct which might be a little confusing - https://github.com/swift-nav/libsbp/blob/f2a1804d0253b288aa186960abdb782a38198fed/rust/sbp/src/messages/gnss.rs#L92

Maybe `MessageGpsTime` or `MsgGpsTime` could help differentiate?

notoriaga:
> Should definitely be swapped, without swapping it makes it kind of hard to use with `?`, right?

It depends on if your function returns Option or Result. Maybe a custom enum with three variants is best, with two helpers to convert to option or result

jbangelo (commented):
> I went with `Option<Result<swiftnav_rs::time::GpsTime, crate::GpsTimeError>>`, but maybe the Result/Option should be swapped?

I'd vote to keep the `Option` on the outside, it seems more logical to me. Getting the time is the first operation, and the not having a time isn't an error per se. The second operation is converting the message's time representation into a common format which is a fallible operation. I think the `?` operator also works with `Option`, assuming that the function is returning an `Option` as well.

generator/sbpg/targets/rust.py
jbangelo: I think this is a good starting point at the very least. I'd have to peruse all the message definitions to figure out if there's other ways we include a GPS time in messages, but this should cover the vast majority if not all of them.

I think it should be noted that the time values can have different meanings depending on the message type they're in. For instance the time value in `MSG_OBS` is the time that the observations were made, but in `MSG_EPHEMERIS_GPS` (or any ephemeris message) it's a center point in the period of time the data is valid so it is typically several minutes to a couple of hours in the future from when the message was actually generated. We can discuss offline if this is a problem for what you're working on.

john-michaelburke:
LGTM thank you for doing this!!

rust/sbp/Cargo.toml
```diff
@@ -33,6 +33,11 @@ log = "0.4"
 nom = "6.0"
 thiserror = "1.0"
 
+[dependencies.swiftnav-rs]
+git = "https://github.com/swift-nav/swiftnav-rs"
+tag = "v0.4.0"
+optional = true
```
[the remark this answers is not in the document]
notoriaga: Hmm not sure if that's a problem. I think as long as this crate doesn't enable that feature in swiftnav-rs we should be alright? 

Either way, I think `SBPMessage` should have a method like this, and this is just the first way that popped into my head on how to do it. Definitely would be open to other ideas. There are three apps that I can think of that read messages and at some point want to check if the current message is time aware - console, choppa, and icbins. So ideally I think this functionality should be provided from one crate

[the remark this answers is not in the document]
silverjam: We should move the libsbp->swiftnav-rs structs feature into libsbp then?

silverjam:
> 
> 
> > I went with `Option<Result<swiftnav_rs::time::GpsTime, crate::GpsTimeError>>`, but maybe the Result/Option should be swapped?
> 
> I'd vote to keep the `Option` on the outside, it seems more logical to me. Getting the time is the first operation, and the not having a time isn't an error per se. The second operation is converting the message's time representation into a common format which is a fallible operation. I think the `?` operator also works with `Option`, assuming that the function is returning an `Option` as well.

👍

generator/sbpg/targets/rust.py
silverjam: For this use case the "time of event" or "time of occurrence" is the meaning of `.gps_time()` -- that is, we want to define a linear time ordering of the messages for use in things like the "sbp chopper" and the interleaver.

I talked with @notoriaga about this briefly and I agree that "t"/"wn"/"tow" covers most of the use case correctly, but we should do an audit of everything that gets grabbed by this and then maybe introduce a "time_ordering" property that we can expose to the yaml spec for this purpose.

generator/sbpg/targets/rust.py
```diff
@@ -23,6 +23,33 @@
 MESSAGES_TEMPLATE_NAME = "sbp_messages_template.rs"
 MESSAGES_MOD_TEMPLATE_NAME = "sbp_messages_mod.rs"
 
+GPS_TIME = """\
```
silverjam, line 26: Mentioned offline, but if the type uses GPSTime, then we potentially have the "nanosecond residual" available... but digging in to this deeper I don't think we need to worry about this (maybe?).  The bare `GPSTime` object is only used in `MSG_TRACKING_STATE_DETAILED_DEP_A`, `MSG_SPECAN`, and (the only one that might matter) is `MSG_OBS`.

rust/sbp/Cargo.toml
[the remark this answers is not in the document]
jbangelo: I'm cool with moving it, but it would be best to make sure these integration features all live in one spot. Whether that's in libsbp or swiftnav-rs doesn't really matter much to me.

generator/sbpg/targets/rust.py
jbangelo: It's not really a technical problem since the two structs with the same name are defined in different crates, so they're in different name spaces. It's mostly a potential cause for confusion among developers, right? I'd like to think that's less likely to happen since the libsbp `GPSTime` struct typically won't be instantiated directly. I imagine it would be accessed as a member of a larger SBP message most of the time. Maybe some clarification in the docs would help?

rust/sbp/Cargo.toml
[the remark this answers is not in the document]
silverjam: 👍

generator/sbpg/targets/rust.py
silverjam: > It's not really a technical problem since the two structs with the same name are defined in different crates, so they're in different name spaces. It's mostly a potential cause for confusion among developers, right? I'd like to think that's less likely to happen since the libsbp `GPSTime` struct typically won't be instantiated directly. I imagine it would be accessed as a member of a larger SBP message most of the time. Maybe some clarification in the docs would help?

Yeah, that's probably sufficient, I was also thinking a wrapper type might be appropriate to disambiguate `MessageTime(GpsTime)` or something similar.

rust/sbp/src/lib.rs
```diff
@@ -169,6 +169,175 @@ pub enum FramerError {
     NoSenderId,
 }
 
+#[cfg(feature = "swiftnav-rs")]
+mod swiftnav_rs_conversions {
```
notoriaga, line 173: pr to remove them in swiftnav-rs here - https://github.com/swift-nav/swiftnav-rs/pull/50

generator/sbpg/targets/rust.py
```diff
@@ -101,11 +128,37 @@ def parse_type(field):
     # This is an inner class, call default constructor
     return "%s::parse(_buf)" % field.type_id
 
+def gps_time(msg, all_messages):
+  def time_aware_header(type_id):
+    for m in all_messages:
+      if m.identifier == type_id:
+        return any([f.identifier == "t" for f in m.fields])
```
silverjam, line 135: Talking with @john-michaelburke we worked out that MSG_OBS and MSG_OSR are a good example where this heuristic doesn't work (for ordering purposes) -- the .t field in the ObservationHeader type is the time of the observation/correction, which could be several seconds before the correct GPS time.  In the console (and the SBP chopper) these messages should be considered part of the GNSS epoch that they are closest too instead of the time that's indicated on the messages.

generator/sbpg/targets/resources/sbp_messages_template.rs
```diff
@@ -15,6 +15,9 @@
 
 //! (((description | replace("\n", "\n//! "))))
 
+#[allow(unused_imports)]
```
jayvdb, line 18: we could remove the need for this, and other lint exclusion rules with
```diff
diff --git a/generator/sbpg/targets/resources/sbp_messages_template.rs b/generator/sbpg/targets/resources/sbp_messages_template.rs
index e2f85da1..ffd7729c 100644
--- a/generator/sbpg/targets/resources/sbp_messages_template.rs
+++ b/generator/sbpg/targets/resources/sbp_messages_template.rs
@@ -15,16 +15,10 @@

 //! (((description | replace("\n", "\n//! "))))

-#[allow(unused_imports)]
-use std::convert::TryFrom;
-
-#[allow(unused_imports)]
-use byteorder::{LittleEndian,ReadBytesExt};
-
 #[allow(unused_imports)]
 use crate::SbpString;
 #[allow(unused_imports)]
-use crate::serialize::SbpSerialize;
+use byteorder::ReadBytesExt;

 ((*- for i in includes *))
 use super::(((i)))::*;
```

clippy doesnt seem to mind; it does however have another 150 warnings :/

generator/sbpg/targets/rust.py
notoriaga: Also looking at the choppa more there seems to be a few more edge cases

- https://github.com/swift-nav/thachoppa/blob/af018c8922749ca7db9a1c1306e0d339babd1bab/src/lib.rs#L135
- https://github.com/swift-nav/thachoppa/blob/af018c8922749ca7db9a1c1306e0d339babd1bab/src/lib.rs#L87

Both of those messages we just treat as normal but it appears they have some special handling required.

notoriaga: > these messages should be considered part of the GNSS epoch that they are closest too instead of the time that's indicated on the messages.

Could this be another category of messages? Like we have time-aware, meaning that message has a time useful for ordering. Then there are "contextually time aware" messages that on their own aren't useful but with additional information can be. Then everything else that's not time aware.

silverjam: The context is definitely very important for what we're building here, I think we need to describe a couple different things:
- any time from a message that maybe be useful
- messages that don't have a time, or that have a time that's not useful in any context we describe with the type system
- time from a message that's useful for ordering or slicing a stream of messages *originating from a device*

So, maybe this could be an enum of some sort:
```rust
enum MessageTime {
  Unknown,
  Base(GpsTime),
  Rover(GpsTime),
}
```

Then we could implement different `PartialOrd` traits depending on use case:
- one that only compares `MessageTime::Rover` for ordering-- that is, only `PartialOrd(Rover, Rover)` can produce an ordering, `PartialOrd(Rover, X)` or `PartialOrd(X, Rover)` should will produce `Ordering::Equal` -- the use case here would be the current use case of ordering and slicing messages from a rover
- one that does the same thing, but only compares `PartialOrd(Base, Base)` -- use case would be ordering/slicing messages from a base station or correction network
- one that compares either `PartialOrd(Base, Rover)`,  `PartialOrd(Rover, Base)`, `PartialOrd(Rover, Rover)`, or  `PartialOrd(Base, Base)` -- not sure what the use case would be since it compares things that are technically "incomparable", but could be included for completeness

silverjam: > 
> 
> Also looking at the choppa more there seems to be a few more edge cases
> 
>     * https://github.com/swift-nav/thachoppa/blob/af018c8922749ca7db9a1c1306e0d339babd1bab/src/lib.rs#L135
> 
>     * https://github.com/swift-nav/thachoppa/blob/af018c8922749ca7db9a1c1306e0d339babd1bab/src/lib.rs#L87
> 
> 
> Both of those messages we just treat as normal but it appears they have some special handling required.

So, under the idea described above we should "denylist" MSG_OBS, MSG_OSR (and MSG_SSR?) from the current ordering and classify it as `MessageTime::Base` and for the Imu message referenced in the thechoppa, we'd probably need a custom trait implementation pulled from theachoppa.

silverjam: Further expanding on the SBP chopper use case, we probably need to have a means to specify a "reference time" when generating a GpsTime for comparison, the sbp chopper this is done when updating the "running clock": https://github.com/swift-nav/thachoppa/blob/master/src/lib.rs#L166.  Many messages only include a TOW, so this would allow the generation of a GpsTime to fill in the `wn` field from a reference time.

notoriaga: > we probably need to have a means to specify a "reference time" when generating a GpsTime for comparison

What if we changed the method to be -

```rust
fn gps_time(&self, reference: impl Into<Option<GpsTime>>) -> Option<Result<GpsTime>>
```

Or maybe we could have two different methods

silverjam: > > we probably need to have a means to specify a "reference time" when generating a GpsTime for comparison
> 
> What if we changed the method to be -
> 
> ```rust
> fn gps_time(&self, reference: impl Into<Option<GpsTime>>) -> Option<Result<GpsTime>>
> ```
> 
> Or maybe we could have two different methods

This sounds fine

notoriaga: Also, I wonder if an iterator adapter could be a good solution here. Something to transform 

`Iterator<Item = SBP>` -> `Iterator<Item = Wrapper>`

```rust
enum Wrapper {
    Message(SBP),
    TimeChanged(GpsTime),
}
```

Whenever a message would change the current time, a `Wrapper::Time` is returned before that message. We could have one adapter for `MessageTime::Base` and one for `MessageTime::Rover`.  With an extension trait like 

```rust
trait SBPIterExt: Iterator<Item = SBP> {
    fn rover_time(self) -> RoverTimeIter;
    fn base_time(self) -> BaseTimeIter;
}
```
people can do `iter_messages(rdr).rover_time().etc()`. also could keep it simple and just expose the adapter it's like

```rust
let messages = iter_messages(rdr);
for _ in RoverTimeIter::new(messages) {}
```

Either way - the adapter seems useful for holding state like the running clock/reference time.

rust/sbp/src/messages/imu.rs
```diff
@@ -176,6 +179,19 @@ impl super::SBPMessage for MsgImuRaw {
     fn write_frame(&self, frame: &mut Vec<u8>) -> std::result::Result<(), crate::FramerError> {
         crate::write_frame(self, frame)
     }
+
+    #[cfg(feature = "swiftnav-rs")]
+    fn gps_time(
+        &self,
+    ) -> Option<std::result::Result<crate::time::MessageTime, crate::time::GpsTimeError>> {
+        let tow_s = (self.tow as f64) / 1000.0;
```
notoriaga, line 187: Is is correct to assume that the tow field always is in milliseconds?

rust/sbp/src/sbp_tools.rs
```diff
@@ -0,0 +1,259 @@
...
+        }
+    }
+}
+
+#[cfg(feature = "swiftnav-rs")]
+impl<I> Iterator for RoverTimeIter<I>
+where
+    I: Iterator<Item = SBP>,
+{
+    type Item = (SBP, Option<GpsTime>);
```
notoriaga, line 141: should probably be `(SBP, Option<Result<GpsTime>>)`

rust/sbp/src/sbp_tools.rs
```diff
@@ -0,0 +1,259 @@
...
+        })
+    }
+
+    fn handle_errors<F>(self, on_err: F) -> HandleErrorsIter<Self, F>
+    where
+        Self: Iterator<Item = Result<SBP>> + Sized,
+        F: FnMut(&crate::Error) -> bool,
+    {
+        HandleErrorsIter::new(self, on_err)
+    }
```
notoriaga, line 33: wanted to try and capture the common kinds of error handling we've been doing when going from 

```
Iterator<Item = Result<SBP>> -> Iterator<Item = SBP>
``` 

The bool in `F: FnMut(&crate::Error) -> bool` controls whether to continue after an error. If it's false it stops and you can call `.check_error()` on the iterator see what the error was. Pretty much https://docs.rs/itertools/0.10.0/itertools/fn.process_results.html but not a freestanding function

I think being able to do both is important but not sure this is the best way of doing it.

rust/sbp/src/sbp_tools.rs
silverjam: agree

rust/sbp/src/messages/imu.rs
silverjam: @jbangelo @denniszollo are there some messages that report tow in seconds, not ms?

@notoriaga there is a "units" field in the yaml spec that we could potentially inspect here.

rust/sbp/src/sbp_tools.rs
silverjam: If we're following an existing pattern then I think this is reasonable.  Maybe @john-michaelburke would have an opinion since he's worked on stuff like this in ICBINS.

john-michaelburke: I think this sounds great with the one request of an additional hybrid of "ignore + log errors". Maybe log them as debug instead of warning though or even specifying loglevel?

notoriaga: Great idea, I added a level param to `log_errors`. Also replaced the bool with a poor mans https://doc.rust-lang.org/std/ops/enum.ControlFlow.html (it's not stable) to clear up the behavior a bit.

rust/sbp/src/messages/imu.rs
notoriaga: I added an assert that the units are ms. Only one wasn't `IMU_MSG_RAW`. For now just skipping that msg, it seems like it might not actually be representing 'rover' time - but rather 'Time since reference epoch in milliseconds' which seems like not what we want

john-michaelburke (approved):
This looks great!

silverjam (approved):

generator/sbpg/targets/resources/sbp-cargo.toml
```diff
@@ -33,6 +33,11 @@ log = "0.4"
 nom = "6.0"
 thiserror = "1.0"
 
+[dependencies.swiftnav-rs]
+git = "https://github.com/swift-nav/swiftnav-rs"
+branch = "steve/sbp"
```
jayvdb, line 38: are we waiting to get https://github.com/swift-nav/swiftnav-rs/pull/50 merged first?

generator/sbpg/targets/rust.py
```diff
@@ -101,17 +134,69 @@ def parse_type(field):
...
+    if msg.identifier in SKIP_GPS_TIME_MSGS:
+        return ""
+
+    body = gen_body()
+    if body is None:
+        return ""
+
+    ret = gen_ret()
+
+    return f"""\
```
jayvdb, line 180: line continuations are an abomination that cause problems for tools parsing the python.  Doing `var = """...""".strip()` usually means it is unnecessary.

also the code is prettier if `textwrap.dedent` is used, and usually .

jayvdb: I think you missed this one, and did all the others.

notoriaga: good catch, thanks

silverjam (changes requested):
Adding this to note that we decided to add something custom to handle IMU_RAW and WHEELTICKS messages from thachoppa: https://github.com/swift-nav/thachoppa/blob/master/src/lib.rs#L148

silverjam (approved):

PR #969, opened 2021-05-06
WIP: Python: Detect and Fix pyflakes errors

jayvdb:


silverjam (commented):
looks great so far

python/sbp/acquisition.py
```diff
@@ -20,7 +20,7 @@
 
 from sbp.msg import SBP, SENDER_ID
 from sbp.utils import fmt_repr, exclude_fields, walk_json_dict, containerize
-from sbp.gnss import *
+from sbp.gnss import GnssSignal, GnssSignalDep, GPSTime, CarrierPhase, GPSTime, GPSTimeSec, GPSTimeDep, SvId  # noqa
```
john-michaelburke, line 23: I find this pylint rule fascinating, you could use 90% of a file and it still prefers you write out all imports as long as one is not imported.

python/test_requirements.txt
```diff
@@ -1,6 +1,7 @@
...
-pytest-cov~=2.5.1
-cov-core~=1.15.0
-coverage~=4.4.1
+pytest
+pytest-cov
+pytest-flakes
+coverage
 tox
 pyyaml
+numpy
```
john-michaelburke, line 7: What is the numpy req used for?

jayvdb: In order for pytest to load `python/sbp/client/util/fftmonitor.py`, which currently has 0% test coverage

python/sbp/acquisition.py
john-michaelburke: Hmm yeah that could definitely be problematic. Thanks for sharing.

jayvdb: fwiw, I was the one who added this rule to pyflakes: https://github.com/PyCQA/pyflakes/pull/46 .  And the errors it generates drives the fixer [`autoflake --expand-star-imports`](https://pypi.org/project/autoflake/)

python/tests/sbp/generator.py
```diff
@@ -0,0 +1,138 @@
...
+# Contact: https://support.swiftnav.com
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+"""
+Generate missing test cases automatically.
```
jayvdb, line 11: I am aware that there is `generator/sbpg/generator.py` . This could be merged there , but that would involve sbpg depending on `sbp` which seems undesirable (possibly an optional dep?), and moving some of the test runner logic into sbpg or sbp.

I am wondering if auto generation of test cases is already possible with sbpg or some other tool; it didnt seem so.

There is still a bug in the currently logic.  The generated test for `MsgNetworkBandwidthUsage` fails `assert_package`.

After that, the goal is to not only look for missing tests, but also to measure the coverage of `assert_package` for each existing test package, and look for incoming msgs which provide new coverage.  This will provide test cases for the inner msg classes that are currently not being exercised because the existing tests are often for simpler instances of the outer "real" message types.

generator/sbpg/targets/python.py
```diff
@@ -114,12 +123,16 @@ def render_source(output_dir, package_spec, jenv=JENV):
   module_path = ".".join(package_spec.identifier.split(".")[1:-1])
   includes = [".".join(i.split(".")[:-1]) for i in package_spec.includes]
   includes = [i for i in includes if i != "types"]
+  include_map = {
```
jayvdb, line 126: this can be re-used in the JS generator which also has a similar map

generator/sbpg/targets/python.py
```diff
@@ -97,6 +97,15 @@ def classnameify(s):
   """
   return ''.join(w if w in ACRONYMS else w.title() for w in s.split('_'))
 
+
+def has_real_message(l):
```
jayvdb, line 101: this can be re-used in the .jj2 wherever they have
```jinja2
((* for m in msgs *))
  ((*- if m.static *))
    ((*- if not m.is_real_message *))
      ...
```

generator/sbpg/targets/resources/sbp_construct_template.py.j2
```diff
@@ -14,15 +14,25 @@
...
 from sbp.utils import fmt_repr, exclude_fields, walk_json_dict, containerize
+((*- else -*))
+import construct
+
+from sbp.utils import fmt_repr
+((*- endif -*))
 
 ((*- for i in include *))
+  ((*- if i in include_map *))
+from (((module_path))).(((i))) import ((* for name in include_map[i] -*))((( name )))((*- if not loop.last *)), ((* endif -*))((*- endfor *))  # noqa
```
jayvdb, line 32: ideally remove the `noqa` and post-process the generated python through autoflake to remove the unused imports.

generator/sbpg/test_structs.py
```diff
@@ -50,13 +56,29 @@ def filepath(self):
...
+      'tests': [test.to_json() for test in self.tests],
+    }
+
+  def write(self, filename, exists_ok=False):
+    file_exists = os.path.exists(filename)
+    assert exists_ok or not file_exists
+    data = self.to_json()
+    try:
+      with open(filename, "w") as f:
+        yaml.dump(data, f)
```
jayvdb, line 73: use yaml formatter settings, or ruamel.yaml, for prettier yaml, or post-process with a yaml reformatter.

generator/sbpg/test_structs.py
```diff
@@ -50,13 +56,29 @@ def filepath(self):
...
+    except Exception:
+      os.remove(filename)
```
jayvdb, line 75: this should only be done if the file did not already exist.  possibly even remove try by doing yaml.dumps outside of the `with`, so the only exceptions possible inside the `with` are fs related problems, which should ripple up.

if keeping the `try`, also re-raise the exception

python/tests/sbp/test_messages.py
```diff
@@ -11,39 +11,44 @@
...
-    test_count += process_files(root, files)
-  for dir in dirs:
-    test_count += process_path(os.path.join(root, dir))
-  return test_count
+    for path in process_files(root, files):
+      yield path
 
-def test_messages():
-  test_count = process_path(ROOTPATH)
+CASES = list(process_path(ROOTPATH))
```
jayvdb, line 37: post-process to remove path prefix which is common to all cases, shortening the test names shown by parameterisation.

python/tests/sbp/generator.py
silverjam: This looks great, if it becomes a lot of work to resolve the final bug let's split this off into a separate PR.

spec/tests/yaml/swiftnav/sbp/observation/test_MsgSvAzEl.yaml
```diff
@@ -0,0 +1,227 @@
+description: Unit tests for swiftnav.sbp.observation MsgSvAzEl
```
silverjam, line 1: We'll need to run the other generators to generate test code for the other languages

python/tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,35},py{36,37,38,39}-{sbp2json,noextra}
+envlist = py{36,37,38,39}-{sbp2json,noextra}
```
silverjam, line 2: What work do we need to do re-enable tests for Python 2.7 and Python 3.5?  It would be nice to keep those for the time being.

john-michaelburke (approved):
LGTM

python/tox.ini
jayvdb: I need to move the new tool out of `tests`, or fiddle with `sbpg` so that it is py27 compatible.  c.f. https://github.com/swift-nav/libsbp/pull/969#discussion_r628288755 about where to put this new tool.

Makefile
```diff
@@ -102,7 +102,7 @@ verify-prereq-python: verify-prereq-generator
 	@command -v pip 1>/dev/null 2>/dev/null    || { echo >&2 -e "I require \`pip\` but it's not installed. Aborting.\n\nHave you installed pip? See the Python readme at \`python/README.rst\` for setup instructions.\n"; exit 1; }
 	@command -v tox 1>/dev/null 2>/dev/null    || { echo >&2 -e "I require \`tox\` but it's not installed. Aborting.\n\nHave you installed tox? See the Python readme at \`python/README.rst\` for setup instructions.\n"; exit 1; }
 	@command -v pandoc 1>/dev/null 2>/dev/null || { echo >&2 -e "I require \`pandoc\` but it's not installed. Aborting.\n\nHave you installed pandoc? See the Python readme at \`python/README.rst\` for setup instructions.\n"; exit 1; }
-	@command -v llvm-config 1>/dev/null 2>/dev/null || { echo >&2 -e "I require \`llvm-config\` but it's not installed. Aborting.\n\nHave you installed LLVM? See the Python readme at \`python/README.rst\` for setup instructions.\n"; exit 1; }
+	@command -v llvm-otool 1>/dev/null 2>/dev/null || { echo >&2 -e "I require LLVM, but \`llvm-otool\` is not installed. Aborting.\n\nHave you installed LLVM? See the Python readme at \`python/README.rst\` for setup instructions.\n"; exit 1; }
```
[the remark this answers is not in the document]
silverjam: It's not needed anymore, it was a dependency of the sbp.jit module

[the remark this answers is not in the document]
silverjam: @jayvdb Let's remove this check?

c/test/auto_check_sbp_observation_55.c
```diff
@@ -10,7 +10,7 @@
  * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
  */
 
-// This file was auto-generated from spec/tests/yaml/swiftnav/sbp/observation/test_msgEphemerisDepB.yaml by generate.py. Do not modify by hand!
+// This file was auto-generated from spec/tests/yaml/swiftnav/sbp/observation/test_MsgSvAzEl.yaml by generate.py. Do not modify by hand!
```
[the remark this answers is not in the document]
silverjam: Would fixing this be a matter of just adding a stable sort somewhere?  Perhaps we can include some part of the `test_MsgSvAzEl` filename in the name of the generated test so we aren't randomly re-writing tests?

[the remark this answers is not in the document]
jayvdb: Created https://swift-nav.atlassian.net/browse/INFRA-465 .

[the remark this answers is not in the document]
silverjam: 👍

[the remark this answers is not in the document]
woodfell: :+1: for basing the output file name on the input file name. The number at the moment really doesn't help when something fails

python/tox.ini
silverjam: I'm find to fork this to a new PR/Jira if that helps move things forward

silverjam (approved):
many suggestions/future improvements inline with the PR, feel free to move those to Jiras for future work

silverjam:
> 
> 
> many suggestions/future improvements inline with the PR, feel free to move those to Jiras for future work

Also, should fix-up the PR title and we should remove this check before merging: https://github.com/swift-nav/libsbp/pull/969#discussion_r629783188

python/tox.ini
jayvdb: Moved to https://github.com/swift-nav/libsbp/pull/979 , and moved it into `generator/` to avoid the py27 problem.

generator/sbpg/test_structs.py
jayvdb: Spawned https://swift-nav.atlassian.net/browse/INFRA-468

PR #971, opened 2021-05-08
Add tcp_2sigma_example

dgburr:
Add an example demonstrating how to parse SBP_MSG_LLH_COV messages and calculate the ellipse corresponding to the 2 sigma/95% confidence interval


dgburr:
I have tidied up the `example` directory: from what I can tell, there was originally only one example in there, but two more examples (`cpp_example` and `tcp_example`) were added later.  I have now moved each example to its own directory and added a top-level CMakefile which will build all 4 of them if BUILD_EXAMPLES is defined.

silverjam (approved):

c/examples/tcp_2sigma_example/tcp_2sigma_example.cc
```diff
@@ -0,0 +1,267 @@
...
+ * by the libopencv-dev package on Debian-like Linux distributions.
+ *
+ * Compile using:
+ * g++ tcp_2sigma_example.cc -DSHOW_PLOT -I ${LIBSBP}/c/include/ -L ${LIBSBP}/c/build/src/ -lsbp -lopencv_core -lopencv_imgproc -lopencv_highgui
+ *
+ * Originally adapted from https://github.com/swift-nav/libsbp/blob/master/c/example/tcp_example/tcp_example.c
+ */
+
+#ifndef SHOW_PLOT
+#define SHOW_PLOT 0
```
silverjam, line 22: @dgburr is it worth installing opencv in CI so we can test compilation of the plotting code?

dgburr: Sorry, but I didn't see this comment before I merged (didn't hit refresh).  Please LMK if you think this is important, in which case I will open a new PR.

silverjam: @dgburr my concern would be around bitrot, if we can compile it easily it'll be easier to keep the code healthy in the long term

PR #973, opened 2021-05-11
Remove haskell from python CI

jayvdb:
test_data/sanity.sh does not use any cli args.


scripts/ci_prepare_python.bash
```diff
@@ -10,30 +10,6 @@ sudo add-apt-repository -y ppa:deadsnakes/ppa
...
-sudo chown -R "$USER:$USER" "$HOME/.docker-cache/stack"
-sudo chown -R "$USER:$USER" "$PWD"
-
-haskell_bins="$(find "$HOME/.docker-cache/stack/work/install" -name bin)"
-
-cp "$haskell_bins/sbp2json" .
-cp "$haskell_bins/sbp2prettyjson" .
-cp "$haskell_bins/json2sbp" .
-cp "$haskell_bins/json2json" .
-cp "$haskell_bins/sbp2nmea" .
```
silverjam, line 39: Aren't these binaries used for the benchmarks?

jayvdb: Hmm... `ci_benchmark.bash` does refer to three of them. We'll see if it fails.  We can have a separate script with these for benchmark only, as they are not needed for the Python test job.

silverjam: Looks like the benchmarks passed and from looking at the CI output it looks like the benchmark phase built the binaries.  I'll wait for @notoriaga to chime in since he mostly recently looked through that code.

notoriaga: Yeah I think it build's them separately. I think initially I tried to share artifacts across jobs but it didn't end up working. I think this should be alright

PR #974, opened 2021-05-11
python: Run coverage and parameterize test cases

jayvdb:
Single commit 4425ac928 on top of https://github.com/swift-nav/libsbp/pull/972

Extracted from https://github.com/swift-nav/libsbp/pull/969


silverjam (approved):
Not sure how to review this, but going to approve based on the fact that there's more lines added than removed 😅

PR #976, opened 2021-05-11
sbpjson: Sync generated code

jayvdb:


jayvdb:
If there are no problems with these changes, I can add CI to enforce this are updated, and instructions for doing it.  `quicktype` could be added to a packages.json, so it is easy for contributors to install it.  Or we could have this target only required to be updated before tags, if there is some way to trigger workflows only prior to a tag.  Perhaps a manual workflow that can be run to verify a branch is ready for tagging.

silverjam (approved):

jayvdb:
Unless there is a better way, I would try making the `USER` based on an arg, like https://stackoverflow.com/questions/57776452/is-it-possible-to-map-a-user-inside-the-docker-container-to-an-outside-user .  Maybe I am missing something simple.

silverjam:
> Unless there is a better way, I would try making the `USER` based on an arg, like https://stackoverflow.com/questions/57776452/is-it-possible-to-map-a-user-inside-the-docker-container-to-an-outside-user .  Maybe I am missing something simple.

We'd need to have a base image (that would probably just be  the current image) and a small docker file that gets built locally based on this image in order to map to a local user.  On Ubuntu systems you can usually just map UID 1000 into the docker container since a typical local install will only have one user... it gets more complicated if you need to build a custom image with a custom UID.  We've done this for piksi_buildroot in the past and it was a bit complicated, maybe there's a better way now, but I don't know if one.

silverjam (approved):

PR #977, opened 2021-05-11
Add tests

jayvdb:
Extracted from https://github.com/swift-nav/libsbp/pull/969


jayvdb:
The bulk of this has been approved as part of https://github.com/swift-nav/libsbp/pull/969 ; there are a few minor non-controversial additions and changes, such as fixing it for py27, and two one-liner CI improvements.

PR #979, opened 2021-05-11
WIP: python: Add automatic test spec generator

jayvdb:
Pulled out of https://github.com/swift-nav/libsbp/pull/969 ; lots of review comments yet to address.


silverjam (commented):
Adding a review so this will get cleared off the devinfra queue but we'd like to port this to master at some point, but maybe change it over to work JSON input from a file rather than opening a socket and listening for messages

PR #980, opened 2021-05-12
[ESD-2024] Auto generate cpp tests

woodfell:
Reuse the existing test schema to generate C++ tests to better exercise C++ SBP components

### DIFF
--- a/.github/workflows/generator.yaml
+++ b/.github/workflows/generator.yaml
@@ -24,7 +24,7 @@ jobs:
       - name: Run generator
         run: |
           pip3 install tox
-          rm c/test/auto_check_*.c rust/sbp/tests/auto_check_*.rs
+          rm c/test/auto_check_*.c c/test/cpp/auto_check_*.cc rust/sbp/tests/auto_check_*.rs
           make gen-c gen-rust
 
       - name: Check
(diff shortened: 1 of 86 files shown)


jayvdb:
We also need to modify https://github.com/swift-nav/libsbp/blob/master/.github/workflows/generator.yaml#L27 to delete the generated c++ test files, so that deleting test specs require that the test code is also deleted.  I have a bunch of other CI work coming up, so I can do this as part of those, if there are no other changes needed for this PR.

woodfell:
> We also need to modify https://github.com/swift-nav/libsbp/blob/master/.github/workflows/generator.yaml#L27 to delete the generated c++ test files, so that deleting test specs require that the test code is also deleted. I have a bunch of other CI work coming up, so I can do this as part of those, if there are no other changes needed for this PR.

Might as well do it here for completeness sake

jayvdb (approved):
It is a bit difficult differentiating C vs C++ test invocation in the run log. The test suite has the same name.  `Running post build test test-libsbp` vs `Running post build test test-libsbp-cpp` is a clue, but not a very big one.  Not sure how we would remedy it, or how important it is.

silverjam (approved):

jbangelo (approved):

PR #985, opened 2021-05-17
[INFRA-288] Fix Java parsing and generate tests

jayvdb:


generator/sbpg/targets/resources/MessageTest.java.j2
```diff
@@ -0,0 +1,83 @@
...
+        org.junit.Assert.assertEquals(msg.(((prefix))), (((value|str_escape))) );
+((*- elif value is array_type *))
+((*- for ff in value *))((( compare_value( (((prefix))) + '[' + (((loop.index0|to_str))) + ']', (((ff))) ) )))((*- endfor *))
+((*- elif value is dict_type *))
+((*- for k in (((value|sorted))) *))((( compare_value( (((prefix))) + '.' + (((k))), (((value[k]))) ) )))((*- endfor *))
+((*- elif value is float_type *))((=
+    Note: the ("%.17e"|format(value)) filter is intended to preserve float
+    literal precision accross all value ranges. =))
+        org.junit.Assert.assertEquals(msg.(((prefix))), ((("%.17e"|format(value)))), DELTA);
+((*- elif value in [2159176030, 3411995557, 7909447587, 8409447265, 8907446923, 9406446591, 9906446269] *))
```
jayvdb, line 39: These seven are somehow different, possibly a bug in the parsing.

generator/sbpg/targets/test_rust.py
```diff
@@ -20,19 +20,13 @@
 
 TEST_TEMPLATE_NAME = "sbp_tests_template.rs"
 
-def b64_decode(field):
```
silverjam, line 23: Moved to common loc. or not used?

generator/sbpg/targets/resources/sbp_java.java.j2
```diff
@@ -52,7 +52,7 @@ public class (((m.identifier | classnameify))) extends SBPMessage {
     public (((m.identifier | classnameify))) () { super(TYPE); }
     public (((m.identifier | classnameify))) (SBPMessage msg) throws SBPBinaryException {
         super(msg);
-        assert msg.type != TYPE;
+        assert msg.type == TYPE;
```
silverjam, line 55: 🤔 how did this ever work? Do asserts not fire in certain cases?

generator/sbpg/targets/resources/MessageTest.java.j2
silverjam: These all look larger than what will fit in 32-bit unsigned int-- 🤷‍♀️

silverjam (commented):
Looks promising so far, int parsing bug is a bit concerning, but we can possibly ship this and do a follow-up since the bug already exists

generator/sbpg/targets/resources/sbp_java.java.j2
jayvdb: no ;-)

I assume the existing usage/examples doesnt go through this constructor, as it uses SBPHandler.

generator/sbpg/targets/test_rust.py
jayvdb: Moved to .common

generator/sbpg/targets/resources/MessageTest.java.j2
jayvdb: The first two were erroneously listed here; and the rest are all greater than unsigned int.
However the problem was endianness. It is only a few attributes where this crops up, and it silently failed, so not surprising nobody noticed this.

silverjam (approved):

PR #986, opened 2021-05-18
rust: support encoding borrowed messages 

notoriaga:
needs https://github.com/swift-nav/dencode/pull/2

Makes it so the encoders work with borrowed messages. Adds convenience functions to get a framed writer. Should we reexport dencode so people don't have to bring it in?

### DIFF
--- a/generator/sbpg/targets/resources/sbp-cargo.toml
+++ b/generator/sbpg/targets/resources/sbp-cargo.toml
@@ -56,8 +56,7 @@ version = "0.3"
 optional = true
 
 [dependencies.dencode]
-git = "https://github.com/swift-nav/dencode"
-branch = "steve/generic"
+version = "0.3.0"
 default-features = false
 
 [dev-dependencies]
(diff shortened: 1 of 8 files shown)


notoriaga:
@john-michaelburke with these changes you can do something like 

```rust
fn test() {
    use dencode::IterSinkExt;

    let input_file = std::fs::File::open("messages").unwrap();
    let log_file = std::fs::File::create("log").unwrap();
    let json_log_file = std::fs::File::create("json_logs").unwrap();

    let mut bin_encoder = SbpEncoder::framed(log_file);
    let mut json_encoder = JsonEncoder::framed(json_log_file, serde_json::ser::CompactFormatter);

    for msg in iter_messages(input_file) {
        let msg = msg.unwrap();

        bin_encoder.send(&msg).unwrap();
        json_encoder.send(&msg).unwrap();
    }
}
```

silverjam:
> ... Should we reexport dencode so people don't have to bring it in?

👍

silverjam (approved):
contingent on re-exporting dencode

john-michaelburke (approved):
Thank you for this!!

PR #987, opened 2021-05-19
docker: add clang for rust build [INFRA-477]

silverjam:
Dependent on https://github.com/swift-nav/swiftnav-rs/pull/51 but will allow us to run `cargo test --verbose --all-features --all-targets` inside the docker image.

In order to build all the Rust targets we need clang to build the swiftnav-rs  dependency.  This is apparently a transitive dependency of bindgen.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -29,4 +29,4 @@ jobs:
       - name: Run make all inside image
         # javascript fails due to permissions
         # rust fails due to missing stdbool.h
-        run: docker run --rm -v $PWD:/mnt/workspace -t libsbp-build:latest sudo --preserve-env=PATH,RUSTUP_HOME,CARGO_HOME,NVM_DIR,NODE_PATH make c python gen-javascript java docs haskell protobuf gen-rust jsonschema quicktype
+        run: docker run --rm -v $PWD:/mnt/workspace -t libsbp-build:latest sudo --preserve-env=PATH,RUSTUP_HOME,CARGO_HOME,NVM_DIR,NODE_PATH make c python gen-javascript java docs haskell protobuf gen-rust jsonschema quicktype rust
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -29,4 +29,4 @@ jobs:
       - name: Run make all inside image
         # javascript fails due to permissions
         # rust fails due to missing stdbool.h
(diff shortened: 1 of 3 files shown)


silverjam:
@jayvdb What should I update so that this gets pulled in to the new CI?

rust/sbp/Cargo.toml
```diff
@@ -60,4 +61,4 @@ version = "0.3.0"
 default-features = false
 
 [dev-dependencies]
-serialport = "2.1.0"
\ No newline at end of file
+serialport = "2.1.0"
```
jayvdb, line 64: eol at eof

Dockerfile
```diff
@@ -48,7 +48,8 @@ RUN \
       uuid-dev \
       libgmp-dev \
       zlib1g-dev \
-      clang-format-6.0
+      clang-format-6.0 \
+      clang
```
jayvdb, line 52: maybe move this up near llvm?

jayvdb (approved):

jayvdb:
Please add rust to the list of targets at https://github.com/swift-nav/libsbp/blob/master/.github/workflows/docker.yaml#L32

.github/workflows/docker.yaml
```diff
@@ -29,4 +29,4 @@ jobs:
       - name: Run make all inside image
         # javascript fails due to permissions
         # rust fails due to missing stdbool.h
-        run: docker run --rm -v $PWD:/mnt/workspace -t libsbp-build:latest sudo --preserve-env=PATH,RUSTUP_HOME,CARGO_HOME,NVM_DIR,NODE_PATH make c python gen-javascript java docs haskell protobuf gen-rust jsonschema quicktype
+        run: docker run --rm -v $PWD:/mnt/workspace -t libsbp-build:latest sudo --preserve-env=PATH,RUSTUP_HOME,CARGO_HOME,NVM_DIR,NODE_PATH make c python gen-javascript java docs haskell protobuf gen-rust jsonschema quicktype rust
```
jayvdb, line 32: rust can replace `gen-rust`, and that puts it in the same order as the `all` target.

But not worth another rebuild in CI.  I'll do the javascript one soon, and then this line changes to `make all`

silverjam: Ok, I want to see the current CI build pass, then I'll push this fix-up

PR #990, opened 2021-05-21
[WIP] [ESD-2025] Unpacked C structs

woodfell:
See https://swift-nav.atlassian.net/wiki/spaces/~61688916/pages/1722942301/Design+Doc+-+libsbp for a background to the problem.

# Overview

Introduced a new generator to produce unpacked C structs - message definitions which are free from the packed attribute, correctly handle variable length arrays. Generate pack and unpack functions for every kind of SBP message. Handle encoding issues such as endianess and string representation in pack/unpack functions.

The unpacked C structs are free from all compiler extensions like zero length arrays. The pedantic flag is now turned on for libsbp, -Wconversion is turned back on and all issues relating to that fixed up.

The old "packed" header files are still available and have not been modified. For the purposes of development they have been moved from `c/include/libsbp` to `c/include/libsbp/packed` although they will be moved back before merging so as to cause as few changes downstream as possible.

# Unpacked structs

The unpacked structs are generated from the same schema and all existing conventions are in use. The structure of the header files matches the packed headers. The unpacked headers can be found under `c/include/libsbp/unpacked/`

Unpacked structs have the same members fields and order as the packed structs but do not have the packed attribute applied to them. The name of the unpacked struct is the same as the packed struct but with `sbp_` prefixed to it - `msg_obs_t` becomes `sbp_msg_obs_t`.

## Variable length arrays

The packed structs use zero length arrays for variable length messages. This is a GCC compiler extensions, the proper C syntax of flexible arrays is not valid C++.

The unpacked structs instead declare variable length arrays to be the maximum possible size and add in a new field to indicate how many items the array contains. Consider MSG_OBS as an example.

The packed version of this struct is:
```
typedef struct __attribute__((packed)) {
  obs_header_t header;
  packed_obs_content_t obs[0];
} msg_obs_t;
```

In order to send a message of this type one must write code such as 
```
u8 payload[SBP_MAX_PAYLOAD_LEN];
msg_obs_t *msg = (msg_obs_t*)payload;
msg->header.t = ...;
msg->obs[0] = ...
msg->obs[1] = ...
sbp_send_message(SBP_MSG_OBS, sizeof(*msg) + n_obs * sizeof(msg->obs[0]), payload);
```

In other words the send of a variable length message must track the encoded size of the message in order to send it. Similarly the receiver of a variable length message must consider the encoded size when accessing a variable length array:
```
void process_obs(msg_obs_t *msg, u8 len) {
u8 n_obs = (len - sizeof(*msg) / sizeof(msg->obs[0]);
for (u8 i = 0; i < n_obs; i++) process(msg->obs[i]);
```

The unpacked structs simplify this process for both the sender and receiver. The unpacked struct is defined as

```
typedef struct {
sbp_obs_header_t header;
sbp_packed_obs_content_t obs[14];
u8 n_obs;
} sbp_msg_obs_t;
```

The value of 14 is calculated by the new generator as the maximum number of items that can fit in to an encoded SBP frame. The extra field `n_obs` is automatically generated and indicates the number of items in the variable length array. The process of using a variable length array is simplified for both sender and receiver.

```
void send_obs() {
sbp_msg_obs_t msg;
msg->header = ...;
msg->n_obs = 2;
msg->obs[0] = ...
msg->obs[1] = ...
sbp_pack_and_send_message(SBP_MSG_OBS, &msg);
```

Similarly for the receiver

```
void process_obs(const sbp_msg_obs_t *msg) {
  for (u8 i = 0; i < msg->n_obs; i++) process(msg->obs[i]);
}
```

## Strings

Variable length strings now have their encoding handled by libsbp. There are a series of helper struct for different encoding formats. They are:
- Unterminated - strings which do not have a NULL terminator on the wire, eg `sbp_msg_log_t::text`
- NULL terminated - strings which do have a NULL terminator on the wire, eg `sbp_msg_fileio_write_req_t::filename`
- Multipart - strings which consist of multiple NULL delimited sections, eg `sbp_msg_settings_read_req_t::setting`
- Double NULL terminated - similar to Multipart but with a double NULL indicating end of sequence, eg `sbp_msg_linux_process_fd_summary::most_opened`
- Binary - strings which are actually binary data, more similar to an array than anything else, eg `sbp_msg_fwd::fwd_payload`

The string encodings have a series of helper functions to handle interacting with the objects. They are defined under `c/include/libsbp/unpacked/string`.

The new generator will process any variable length strings and generate wrapper functions for interacting with the fields in message objects. For example to set the text of a log message one would write
```
sbp_msg_log_t msg;
sbp_msg_log_t_text_set(&msg->text, "A helpful log message");
```

libsbp will track the length and encoding of the string and ensure that on the wire it is represented properly.

The string functions try to copy with encoding issues they may encounter due to not having these helpers before such as an incoming packet which contains a NULL terminated string which for whatever reason doesn't have a NULL terminator.

The string classes ensure that strings are always properly represented in C format

## Differences

Previously empty structs were not generated (eg, msg_settings_read_by_index_done). 

A new behaviour is that the unpacked generate will create a definition for these messages as well. Since empty structs are illegal in C and badly defined in C++ the generate will create a single dummy field in these struct which is ignored by all other parts of libsbp when processing these new structs. This allows some useful work later on such as expanding C++ message traits to include the full set of SBP messages.

# New APIs

There is a new callback type to receive unpacked structs. It behaves in the same manner as the other 2 callback types except it received a parameter of type `const sbp_msg_t *` and does not have any length parameters.

`sbp_msg_t` is created by the new generator and is the union of all real SBP messages:

```
typedef union {
sbp_msg_obs_t obs;
sbp_msg_log_t log;
sbp_msg_pos_llh_cov_t pos_llh_cov;
...
} sbp_msg_t;
```

There is a new send function which can take an instance of `sbp_msg_t`.
```
s8 sbp_pack_and_send_message(u16 sender_id, u16 msg_type, const sbp_msg_t *msg);
```

There is also `sbp_process_unpacked` which behaves similar to `sbp_process_payload`

A user of libsbp can manually pack and unpack message if required with the general purpose function

```
s8 sbp_pack_msg(u8 *buf, u8 len, u16 msg_type, const sbp_msg_t *msg);
s8 sbp_unpack_msg(const u8 *buf, u8 len, u16 msg_type, sbp_msg_t *msg);
```

The size of unpacked structs do not corrospond to the encoded length. Should a user actually need to know this information they can call `u8 sbp_packed_size(u16 msg_type, const sbp_msg_t *msg)`

The generator creates specialisations for each real message type to avoid having to use the union (most useful for C++)

```
s8 sbp_pack_sbp_msg_obs_t(u8 *buf, u8 len, const sbp_msg_obs_t *msg);
s8 sbp_unpack_sbp_msg_obs_t(const u8 *buf, u8 len, sbp_msg_obs_t *msg);
u8 sbp_packed_size_sbp_msg_obs_t(const sbp_msg_obs_t *msg);
```

To replace the use of `memcmp` with unpacked structs the generator creates from `cmp` functions:
```
int sbp_msg_cmp(u16 msg_type, const sbp_msg_t *a, const sbp_msg_t *b);
int sbp_cmp_sbp_msg_obs_t(const sbp_msg_obs_t *a, const sbp_msg_obs_t *b);
```

There are a full set of C++ operators based on these functions.

# API changes

The existing C API is unchanged.

The C++ wrappers in `c/include/libsbp/cpp` have had to change slightly.

The packed headers contain syntax which is illegal in C++ so moving fowards we must rely on the unpacked structs solely. 

`sbp::MessageHandler` is updated to work only with unpacked messages. The CallbackInterface helper is updated to receive an `sbp_msg_t` from libsbp and pull out the correct field to pass to the overridden message handles.

For users of this class the only real change is that the handlers need to be updated to remove the `uint8_t message_length` parameter which is no longer provided, and change the type of the incoming message from `msg_xxx_t` to `sbp_msg_xxx_t`. Any handler which deals with variable length messages will need to be changed to use the generate `n_xxx` field instead of the message length.

Message traits is expanded to include the full set of SBP message types thanks to now generating "empty" structs.

# Testing

The auto generated test suites are updated to deal with unpacked structs in addition to packed. The packed checks remain the same, the unpacked checks follow the same procedure of creating an unpacked struct, sending it, validating the encoded frame, unpacking it, checking field values.

Some test schemas need updating to cope with the new fields generated for use with variable length arrays.

# Schema changes

A few schema items needed to be updated although wire format is unaffected.

## MSG_GROUP_META

The field `n_group_msgs` has some meta information added to it which links it to the variable length array `group_msgs`. This stops the generator from trying to automatically generate a field of the same name. Users of this message do not need to change, they can simply stop tracking the encoded size of the message since this field now behaves like one would expect.

## Strings

Any message which contains a variable length string has some meta information attached to it which describes the string encoding. This is only used by the new unpacked generator and is invisible anywhere else

## SSR GridElement

The types GridElement and GridElementNoStd both contain a variable length array but are not real messages. They are currently used as an embedded type in a couple of messages.

The problem with this is that it's impossible to calculate the maximum number of items stored in the variable length array in a way that is valid for all potential uses of the embedded type. This is already a problem for MSG_SSR_GRIDDED_CORRECTION and MSG_SSR_GRIDDED_CORRECTION_DEP_A which can store different numbers of items in their GridElement::stec_residuals fields

There are 3 possible solutions to this:
- Generate embedded types as nested structs. This was trialled and caused a huge number of changes downstream
- Special case for these types, generate them as nested structs but nothing else. Extra complication in the generator
- Change the spec to remove GridElement and GridElementNoStd, move their fields in to the real messages

Option 3 is implemented here. The wire encoding for these messages is not changed but it is technically a "real" schema change.

# Demonstration

A prototype up to the level of starling-binary has been demonstrated for these changes, although it is a little out of date now. It was functional, the changes were not too intrusive and it allowed the `pedantic` flag to be turned on for all portable targets.

# Work required

This is still a little rough around the edges. Documentation is missing for the unpacked structs and the formatting is terrible. The string helpers in particular can be simplified greatly

# Review notes

I'd like some feedback about the general approach here without concentrating too much on the nitty gritty. The timeline on this is pretty open but this is the one thing that's really blocking strict language compliance in starling so I would like to make some progress on it. I don't intend to merge this PR as it is.

Particularly looking for comments on anything that might cause downstream problems I haven't thought of. Are there any users of libsbp that might be affected unintentionally? 

What is the future of the pack structs? Are they worth keeping around?

How much of the new API is useful? Is there anything missing?


jbangelo:
> The old "packed" header files are still available and have not been modified.

> What is the future of the pack structs? Are they worth keeping around?

Leaving the packed structs around for now is probably the right approach. Longer term I think we should plan on making a breaking API change and move towards removing them entirely. Maybe do it in a couple steps, the first to deprecate them and move them to a `packed` subdirectory for a time and the second step would be to completely removing them?

dgburr (approved):
LGTM: this version compiles with gcc 8.3 (Debian 10.9), whereas current master does not (due to `array subscript 0 is above array bounds` warnings), which is great.  One minor nit (which I realise is not caused by this PR): would it be possible to change all of the `This file was auto-generated [...] by generate.py` lines to `This file was auto-generated [...] by generator.py`?  Here is an example patch to do this:
[fix-comment.diff.txt](https://github.com/swift-nav/libsbp/files/6565015/fix-comment.diff.txt)

silverjam:
From @jbangelo 
> > The old "packed" header files are still available and have not been modified.
> 
> > What is the future of the pack structs? Are they worth keeping around?
> 
> Leaving the packed structs around for now is probably the right approach. Longer term I think we should plan on making a breaking API change and move towards removing them entirely. Maybe do it in a couple steps, the first to deprecate them and move them to a `packed` subdirectory for a time and the second step would be to completely removing them?

From @woodfell
> The old "packed" header files are still available and have not been modified. For the purposes of development they have been moved from `c/include/libsbp` to `c/include/libsbp/packed` although they will be moved back before merging so as to cause as few changes downstream as possible.

I think this API should be the default in the future, and we should plan a for a migration strategy

I'll propose the following layout and migration strategy:
- `c/include/libsbp/new` - new "unpacked" API
- `c/include/libsbp/legacy` - old "packed" API 
- `c/include/libsbp` - also the old "packed" API

As you also mentioned, this will have minimal downstream impact and will allow us to move code that needs the new API to the `#include <libsbp/new/...>` API path.  We should migrate all internal code that needs the legacy API to the `#include <libsbp/legacy/...>` include path.  Then eventually migrate all internal code to use the new API.

Then I would propose the following rough timeline:
- v4 release: "old" API maintained at default location and at `/legacy`
- v5 release: "old" API removed from  default location, "new" API moved to default location, old applications must switch to use "/legacy" include to continue using the old API
- v6 release: "old" API removed

Each of these releases could be on a two month cadence.

silverjam:
We should name things here with the intent that this will eventually replace the existing API, e.g.
- `sbp_pack_and_send_message` -> `sbp_send_message` (name can be shortened since packing/encoding is implied)

In general I think should use "encode" and "decode" nomenclature since it's a more common terminology for message formats (and it's more consistent with other languages in this repository), pack/unpack are somewhat C specific.  For example, we should rename these:

- `sbp_pack_msg` -> `sbp_encode_message`
- `sbp_unpack_msg` -> `sbp_decode_message`

Then we can drop `_packed` or `_unpacked` suffixes, since then intent is to eventually replace the existing API:

- `sbp_process_unpacked` -> `sbp_process_message`

It would also be nice if we could be consistent about where `msg` vs `message` is used, maybe leave the abbreviation `msg` for types, but spell out the word for function invocations?

silverjam:
> Option 3 is implemented here. The wire encoding for these messages is not changed but it is technically a "real" schema change.

As long as the wire format isn't changed, I think we can get away with this-- I would worry about this becoming a problem in the future though.  Can we add a check to the generator to disallow variable length arrays in nested/embedded types?

c/include/libsbp/unpacked/string/multipart.h
```diff
@@ -0,0 +1,68 @@
...
+#include <stdint.h>
+
+#include <libsbp/common.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#define SBP_MULTIPART_STRING_MAX_SECTIONS 20
```
silverjam, line 15: Where did the max of 20 come from?

c/include/libsbp/unpacked/string/multipart.h
```diff
@@ -0,0 +1,68 @@
...
+                                                va_list ap);
+
+  bool sbp_multipart_string_append(sbp_multipart_string_t *s,
+                                   const sbp_multipart_string_params_t *params,
+                                   const char *new_str);
+  bool sbp_multipart_string_append_vprintf(sbp_multipart_string_t *s,
+                                           const sbp_multipart_string_params_t *params,
+                                           const char *fmt,
+                                           va_list ap);
+  const char *sbp_multipart_string_get_section(const sbp_multipart_string_t *s,
```
silverjam, line 60: So, seeing two future tasks here:
- rework https://github.com/swift-nav/libsettings to utilize this new API and help with validation
- create a Piksi Multi build that utilizes this new code, also update piksi_tools (via libsettings Python code)

c/include/libsbp/unpacked/string/null_terminated.h
```diff
@@ -0,0 +1,60 @@
...
+  uint8_t sbp_null_terminated_string_packed_len(const sbp_null_terminated_string_t *s,
+                                                const sbp_null_terminated_string_params_t *params);
+  uint8_t sbp_null_terminated_string_space_remaining(const sbp_null_terminated_string_t *s,
+                                                     const sbp_null_terminated_string_params_t *params);
+  uint8_t sbp_null_terminated_string_strlen(const sbp_null_terminated_string_t *s, const sbp_null_terminated_string_params_t *params);
+
+  bool sbp_null_terminated_string_set(sbp_null_terminated_string_t *s,
+                                      const sbp_null_terminated_string_params_t *params,
+                                      const char *new_str);
+  bool sbp_null_terminated_string_vprintf(sbp_null_terminated_string_t *s,
```
silverjam, line 42: Does this (and others) need to be annotated with a printf format attribute?  E.g. `__attribute__((format(printf, ...)))`?

c/src/include/libsbp/internal/unpacked/common.h
```diff
@@ -0,0 +1,298 @@
...
+  return pack_u16(ctx, (const u16 *)v);
+}
+
+static inline size_t unpack_s16(sbp_unpack_ctx_t *ctx, s16 *v) {
+  return unpack_u16(ctx, (u16 *)v);
+}
+
+static inline size_t sbp_packed_size_s32(const s32 *v) {
+  (void)v;
+  return 4;
```
silverjam, line 150: Would a compiler assertion here be useful?  To validate that sizes are what we expect them to be?

c/src/include/libsbp/internal/unpacked/common.h
```diff
@@ -0,0 +1,298 @@
...
+
+static inline size_t sbp_packed_size_double(const double *v) {
+  (void)v;
+  return 8;
+}
+
+static inline size_t pack_double(sbp_pack_ctx_t *ctx,
+                                     const double *v) {
+  if (!SBP_CAN_PACK(ctx, double)) { return false; }
+  memcpy(&ctx->buf[ctx->offset], v, 8);
```
silverjam, line 215: Double's are represented the same on big/little endian?

silverjam (commented):
Promising start, some additional general feedback on the PR:
- generated test suite should cover both C APIs, new and legacy
- we'll need unit tests for new string types and new pack/unpack functions
- new string types seem like a lot of similar code, would be good to try and consolidate

c/include/libsbp/unpacked/string/multipart.h
woodfell: Chosen at random. This can be removed at the expense of CPU time. I went through several different versions of the string functions and they were all pretty terrible, but they proved a point. I think they all need to be rewritten and, like you said elsewhere, consolidate.

c/include/libsbp/unpacked/string/multipart.h
woodfell: Interesting. Does the python version of libsettings wrap the C library somehow?

c/src/include/libsbp/internal/unpacked/common.h
woodfell: Good point. I think the size of these basic types can be added to the list of safe assumptions and a static_assert added

c/src/include/libsbp/internal/unpacked/common.h
woodfell: :+1: yeah, no, that needs to be done as well

c/include/libsbp/unpacked/string/multipart.h
silverjam: Yes, there's a Cython wrapper of libsettings that's used in piksi_tools (the Python version of the console) in the next couple weeks we'll have a Rust wrapper for libsettings for the new console

c/include/libsbp/unpacked/string/multipart.h
silverjam: Ok, 20 seems like a reasonable upper limit given the max usage that I know of is 3 for the settings code

woodfell:
> Then I would propose the following rough timeline:
> 
>     * v4 release: "old" API maintained at default location and at `/legacy`
> 
>     * v5 release: "old" API removed from  default location, "new" API moved to default location, old applications must switch to use "/legacy" include to continue using the old API
> 
>     * v6 release: "old" API removed
> 
> 
> Each of these releases could be on a two month cadence.

I'm in favour of changing `sbp_pack_and_send_message` to `sbp_send_message` and doing something else with the current function. Where would that fit in to this timeline? Part of the v4 release? Or deferred to later?

c/include/libsbp/unpacked/string/multipart.h
woodfell: There are a few messages which can potentially have more. settings read resp (I think) has an optional 4th section which is only present for enum types. Some of the more piksi specific messages like read_dir_resp or process_fd_summary are technically unbounded, although there is practical limit based on the length of filenames and such.

silverjam:
> > I'm in favour of changing `sbp_pack_and_send_message` to `sbp_send_message` and doing something else with the current function. Where would that fit in to this timeline? Part of the v4 release? Or deferred to later?
> 
> If we can, I think it makes sense to use whatever the final naming is going to be, so we can minimize the impact of each of these major version revs.

That means there's a few different usage scenarios, for the releases:

+ I'm already in legacy code, I don't need to do anything for the v4 release, but I'm aware that there's a new API that will become the default in X months (where X is 4-6 months, see below)

+ I'm writing new code, I can use the new API, for the v5 release I just need to update my include paths, for v6 I don't need to do anything

+ I ported during the v4 release, I'm in the same boat as the "new code" people

+ I didn't port by the v5 release, I need to update my include paths, I must port for the v6 release

+ I didn't port for the v6 release, I don't need to update my include paths, but I now need to port to the new API

As for the value of "X" (the release cadence) I was thinking we should align with an existing customer facing release vehicle like Starling and do this over the next 3 releases.

woodfell:
Perhaps we could look at applying the `deprecated` attribute to all the old APIs as well. It would be quite annoying at first but would quickly show up what needs changing. We'd probably have to change the set of compiler options to include `-Wno-error=deprecated-declaration`

woodfell:
Also to ease the various stages it might make sense to reorganise the headers in the initial v4 release. `sbp.h` is pretty full of stuff, breaking the old API out to something like `legacy/api.h` and just `#include`ing that file from `sbp.h` in v4 could make later releases much less work

woodfell:
@silverjam I've written up a description of how I can see us moving away from the old API towards the new one based on my understanding of your comments - https://swift-nav.atlassian.net/wiki/spaces/~61688916/pages/1832157449/C+API+migration+plan. Does this align with what you were thinking?

PR #991, opened 2021-05-31
[ESD-2032] Correct spec for fwd_payload member of MSG_FWD

woodfell:
fwd_payload is currently defined as a string. In reality the contents of this field is binary data with an undefined structure. 

This isn't causing problems at the moment, most languages handle strings in a way which is suitable. For the improvements to the C bindings which are required for portability and language compliance this field encoding would need special handling. It can be simplified by altering it so that it's an array of u8 rather than a string.


silverjam:
Should we deprecate and create a new message?

woodfell:
> Should we deprecate and create a new message?

Not sure. If we deprecate it and the old definition hangs around then we'd still need special handling for this string-but-not-a-string field such as the implementation of `sbp_binary_string_t` [here](https://github.com/swift-nav/libsbp/blob/3f13480a1dd313a4810b9e2ff4d587d4e6ca4e4d/c/src/unpacked/string/binary.c). And if we have to have this special case anyway there isn't really any point in deprecating it.

This change doesn't actually change the wire encoding, only how it's represented in the language bindings. There will be some changes to types all languages, not just C, but this message isn't exactly widely used. If you think this is too much of a change to make to the spec then it's probably better to just go with the special case sbp_binary_string_t in C

silverjam (approved):
Approved since there's no change in the wire encoding

jayvdb (approved):

PR #993, opened 2021-06-03
rust: wip callback based handler

notoriaga:
For the file io stuff I'm working on it'd be nice to have a way to register callbacks for specific messages. This kind of works?

### DIFF
--- a/generator/sbpg/targets/resources/sbp_messages_mod.rs
+++ b/generator/sbpg/targets/resources/sbp_messages_mod.rs
@@ -27,6 +27,7 @@ use crate::serialize::SbpSerialize;
 pub trait SBPMessage: SbpSerialize {
     fn get_message_name(&self) -> &'static str;
     fn get_message_type(&self) -> u16;
+    fn message_type() -> u16 where Self: Sized;
     fn get_sender_id(&self) -> Option<u16>;
     fn set_sender_id(&mut self, new_id: u16);
     fn to_frame(&self) -> std::result::Result<Vec<u8>, crate::FramerError>;
@@ -86,6 +87,10 @@ impl crate::SBPMessage for SBP {
                 msg.get_message_type()
             },
         }
+    }
+
+    fn message_type() -> u16 {
+        0
     }
 
     fn get_sender_id(&self) -> Option<u16> {
--- a/generator/sbpg/targets/resources/sbp_messages_mod.rs
+++ b/generator/sbpg/targets/resources/sbp_messages_mod.rs
@@ -27,7 +27,6 @@ use crate::serialize::SbpSerialize;
 pub trait SBPMessage: SbpSerialize {
     fn get_message_name(&self) -> &'static str;
     fn get_message_type(&self) -> u16;
-    fn message_type() -> u16 where Self: Sized;
     fn get_sender_id(&self) -> Option<u16>;
     fn set_sender_id(&mut self, new_id: u16);
     fn to_frame(&self) -> std::result::Result<Vec<u8>, crate::FramerError>;
@@ -36,6 +35,10 @@ pub trait SBPMessage: SbpSerialize {
     fn gps_time(&self) -> Option<std::result::Result<crate::time::MessageTime, crate::time::GpsTimeError>> {
         None
     }
+}
+
+pub trait MessageType: Sized {
+    fn message_type() -> u16;
 }
 
 #[cfg_attr(feature = "sbp_serde", derive(serde::Serialize), serde(untagged))]
@@ -87,10 +90,6 @@ impl crate::SBPMessage for SBP {
                 msg.get_message_type()
             },
         }
-    }
-
-    fn message_type() -> u16 {
-        0
     }
 
     fn get_sender_id(&self) -> Option<u16> {
--- a/generator/sbpg/targets/resources/sbp_messages_mod.rs
+++ b/generator/sbpg/targets/resources/sbp_messages_mod.rs
@@ -201,3 +201,14 @@ impl From<Unknown> for SBP {
         SBP::Unknown(msg)
     }
 }
+
+#[derive(Debug)]
+pub struct TryFromSBPError;
+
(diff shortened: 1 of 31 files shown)


generator/sbpg/targets/resources/sbp_messages_mod.rs
```diff
@@ -27,6 +27,7 @@ use crate::serialize::SbpSerialize;
 pub trait SBPMessage: SbpSerialize {
     fn get_message_name(&self) -> &'static str;
     fn get_message_type(&self) -> u16;
+    fn message_type() -> u16 where Self: Sized;
```
notoriaga, line 30: An associated method is needed here so that you can get the message type of a message without having to actually construct one. Not sure the best way to do this. Generally `get_message_type` could be an associated function rather than a method except for in the case of `SBP` where the message type changes depending on the message.

Might be worth having a separate trait like 
```rust
trait OtherSBPMessageTrait {
    fn message_type() -> u16;
}
``` 
that is only implemented by the actual messages

rust/sbp/src/sbp_tools.rs
```diff
@@ -1,13 +1,52 @@
...
 #[cfg(feature = "swiftnav-rs")]
-use crate::{
-    messages::SBPMessage,
-    time::{GpsTimeError, MessageTime, RoverTime},
-};
+use crate::time::{GpsTimeError, MessageTime, RoverTime};
+
+use crate::messages::{SBPMessage, SBP};
+
+pub struct Dispatcher {
```
notoriaga, line 11: I'm not sure if this actually belongs in libsbp, just want to show what I'd like the end result to be

rust/sbp/src/sbp_tools.rs
```diff
@@ -269,4 +311,29 @@ mod tests {
         assert_eq!(messages.count(), 0);
         assert_eq!(err_count, 1);
     }
+
+    fn test_dispatcher() {
```
notoriaga, line 315: what it looks like to use it

rust/sbp/src/sbp_tools.rs
```diff
@@ -1,13 +1,52 @@
...
+}
+
+impl Dispatcher {
+    pub fn new() -> Self {
+        Self {
+            callbacks: HashMap::new(),
+        }
+    }
+
+    pub fn on<F, M>(&mut self, mut func: F)
```
notoriaga, line 22: there should also be a method `on_once` where the callback gets removed after it gets triggered once

generator/sbpg/targets/resources/sbp_messages_mod.rs
notoriaga: And re the `Self: Sized` bound, I think we could probably just add that as 

```
trait SBPMessage: SbpSerialize + Sized {
   ...
}

because all the messages have a known size

File: rust/sbp/src/sbp_tools.rs [hunk 5]
```diff
@@ -7,7 +9,47 @@ use crate::{
...
+    where
+        F: FnMut(M) + 'static,
```
jbangelo, line 27: I'm sure it'll make things much more complicated, but it would be awesome if we could drop the static lifetime here.

rust/sbp/src/sbp_tools.rs
```diff
@@ -7,7 +9,47 @@ use crate::{
     time::{GpsTimeError, MessageTime, RoverTime},
 };
 
-use crate::messages::SBP;
+use crate::messages::{MessageType, SBP};
+
+pub struct Dispatcher {
+    callbacks: HashMap<u16, Vec<Box<dyn FnMut(SBP)>>>,
```
jbangelo, line 15: Probably an over optimization (and would need some benchmarking to prove/disprove), but I think a `Vec<Box<dyn FnMut(SBP)>>` might be faster, you'd also need to make the the closure swallow the error on a failed conversion. Maybe doing all those conversions is too expensive?

rust/sbp/src/sbp_tools.rs
notoriaga: Hmm not 100% sure how to do it/if it's possible. Afaik boxed closures will always need a static lifetime. It might be possible to not box the closures, which would be great. They'll still need to be behind a pointer of some kind because they are unsized. So something like having the callbacks be `&'a dyn FnMut(SBP)`.  So changing

```rust
self.callbacks
    .entry(M::message_type())
    .or_default()
    .push(Box::new(move |msg: SBP| {
        let msg: M = msg.try_into().unwrap();
        func(msg)
    }));
```

to 

```rust 
self.callbacks
    .entry(M::message_type())
    .or_default()
    .push(&move |msg: SBP| {
        let msg: M = msg.try_into().unwrap();
        func(msg)
    });
```

Would let us drop `'static` in favor of generic lifetime `'a`. But, because we actually are storing a new closure rather than the one that was passed by the users you'll get a `temporary value dropped while borrowed` error. Still might be possible just not immediately obvious how to do it

jbangelo: I think the boxed closure can have a shorter lifetime than `'static`, `Dispatcher` just needs to be generic over that lifetime. That'll be a limitation of any callback system like this, any references that the function objects capture will have to outlive the `Dispatch` object. I'm not sure how you could get around that.

Here's a very basic example of how I got the lifetimes to work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f1e74838c33e509a33d637d4a213cc41

rust/sbp/src/sbp_tools.rs
jbangelo: I'm not sure if the libsbp repo is the right place, but I definitely could see this being a separate crate (`sbp-dispatcher` maybe?).

rust/sbp/src/sbp_tools.rs
notoriaga: Ahh I see, worked like a charm!

silverjam:
@notoriaga Can close this PR now right?

PR #996, opened 2021-06-07
generator (c): Sort the spec files

samvrlewis:
Makes the ordering of the includes and case statements in sbp_msg.h
deterministic.


woodfell (approved):

RReichert (approved):
that will work :+1:

PR #997, opened 2021-06-07
additional generated work

RReichert:
This PR piggy backs on https://github.com/swift-nav/libsbp/pull/996. Seems like the `Generated artifacts` stage was failing because the Java and Rust generated files weren't up to date. Just introducing them into the commit didn't kick off the stage because the files themselves weren't in the list of file changes that would kick off the stage. I've suggested removing those files from the list, not sure if this was done to save $$$.

### DIFF
--- a/c/include/libsbp/new/sbp_msg.h
+++ b/c/include/libsbp/new/sbp_msg.h
@@ -52,42 +52,40 @@ extern "C" {
 struct sbp_state;
 
 typedef union {
-  sbp_msg_tracking_state_detailed_dep_a_t tracking_state_detailed_dep_a;
-  sbp_msg_tracking_state_detailed_dep_t tracking_state_detailed_dep;
-  sbp_msg_tracking_state_t tracking_state;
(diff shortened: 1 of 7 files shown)


woodfell (approved):

c/include/libsbp/new/sbp_msg.h
```diff
@@ -52,42 +52,40 @@ extern "C" {
 struct sbp_state;
 
 typedef union {
-  sbp_msg_tracking_state_detailed_dep_a_t tracking_state_detailed_dep_a;
```
samvrlewis, line 55: Weird, I don't think there's actually any diff between these two branches on this file. Maybe a Github bug?

samvrlewis (approved):

jayvdb (approved):
> ... $$$

Not really.  It started to track the tests which were out of sync.  It has rapidly grown to check most of the generated files, and it is a quick job so doesnt hurt running it all the time now.

PR #998, opened 2021-06-07
[ESD-2047] Generated tests for new API

woodfell:
All existing generated tests are renamed/moved to contain `legacy_`.

Added a new generator to replace `test_c.py` which generates tests for the new API based on the test specs. There are tests for both the C and C++ APIs, they are essentially duplicates of all the existing generated tests.

This diff is awful because github doesn't like it when fileA moves to fileB and then a new version of fileA is made. I suggest reviewing just the new templates and one or two of the generated test suites, then run `make gen-c` to see that everything else is just generated


c/include/libsbp/cpp/state.h
```diff
@@ -85,6 +86,11 @@ class State {
     // NOLINTNEXTLINE
     return sbp_process_payload(&state_, sender_id, msg_type, msg_length, const_cast<u8 *>(payload));
   }
+
+  template<typename T>
+  s8 send_message(u16 sender_id, const T &msg) {
```
woodfell, line 91: This somehow got missed from the initial import of the prototype generator. It's used heavily in the generate C++ tests for the new API

c/scripts/clang-format.sh
```diff
@@ -7,8 +7,8 @@ set -e -x
 # in the Google style of hand-written code.
 GENERATED_HEADERS=$(grep -rl --include="*.h" "Automatically generated" include/libsbp/*)
 CLANG_FORMAT=clang-format-6.0
-$CLANG_FORMAT -i $GENERATED_HEADERS test/*.c test/*.h test/cpp/*.cc
+$CLANG_FORMAT -i $GENERATED_HEADERS test/*.c test/*.h test/legacy/cpp/*.cc test/legacy/*.c test/auto* test/cpp/auto*
 
 # clang-format-6 adds whitespace to blank lines inside comments
-sed -i.bak 's/^  *$//' $GENERATED_HEADERS
+sed -i 's/^  *$//' $GENERATED_HEADERS
```
jayvdb, line 13: this is needed for compatibility with older sed, esp mac.

RReichert: I've fallin into this trap so many times

.gitignore
```diff
@@ -4,7 +4,6 @@
 *.*~
 limbo/
 .idea/
-*.bak
```
RReichert, line 7: :tada:

samvrlewis (approved):
Reviewed this by looking at the legacy test vs. the new test for a new generated test files, the new test structure looks good and the checks that have been removed make sense to me.

The only issue is, as mentioned, that this doesn't build on GCC 9.3.0  for me at the moment.

RReichert (approved):
I haven't had a chance to review this, given that there are a number of merge conflicts, I'd be more in favour of getting everything into the portability branch ASAP and deal with having a big review at the end of the project.

PR #1008, opened 2021-06-08
Fix big endian for new API [ESD-2057]

samvrlewis:
Fixes all of the remaining big endian issues for the new API (including some pre-existing issues that were in sbp.c) 

To go with this, and so that we have confidence that we truly do support big endian I also added a new CI test that runs the generated tests against the new API on a MIPS (big endian) qemu instance. The legacy API doesn't pass this (as expected) so to get the tests to pass, I divided up the legacy and new API tests into two separate test programs and added the ability to only test the new API to the makefile.

Happy to hear any suggestions on how to make any of this cleaner.


c/src/sbp.c
```diff
@@ -438,6 +438,16 @@ static u16 sbp_u8_array_to_u16(u8 *array_start)
   return (u16) array_start[0] + ((u16) array_start[1] << 8);
 }
 
+/** Helper to convert a u16 in the platform's representation
+ *  to the a 2 byte array in network byte order without
```
woodfell, line 442:
```suggestion
 *  to the a 2 byte array in little endian byte order without
```

silverjam (commented):
+1 for @woodfell's requests but otherwise this looks great

woodfell (approved):

PR #1009, opened 2021-06-08
rust: impl TryFrom<SBP> for all messages

notoriaga:
Still working on the event handler thing, but it definitely still needs some work. I think for now it should live in the console as it's a pretty good proving ground for the concept. this is just the minimum changes that are needed to implement it elsewhere. Both changes also seem generally useful

### DIFF
--- a/generator/sbpg/targets/resources/sbp_messages_mod.rs
+++ b/generator/sbpg/targets/resources/sbp_messages_mod.rs
@@ -37,6 +37,21 @@ pub trait SBPMessage: SbpSerialize {
     }
 }
 
+pub trait RealMessage: SBPMessage + std::convert::TryFrom<SBP, Error = TryFromSBPError> + Clone + Sized {
+    const MESSAGE_TYPE: u16;
+}
+
+#[derive(Debug, Clone)]
+pub struct TryFromSBPError;
+
+impl std::fmt::Display for TryFromSBPError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "invalid message type for conversion")
+    }
+}
+
+impl std::error::Error for TryFromSBPError {}
+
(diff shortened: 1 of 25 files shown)


generator/sbpg/targets/resources/sbp_messages_mod.rs
```diff
@@ -37,6 +37,21 @@ pub trait SBPMessage: SbpSerialize {
     }
 }
 
+pub trait RealMessage: SBPMessage + std::convert::TryFrom<SBP, Error = TryFromSBPError> + Clone + Sized {
+    const MESSAGE_TYPE: u16;
```
notoriaga, line 41: Could also add `const MESSAGE_NAME: &static str` here as well I suppose

generator/sbpg/targets/resources/sbp_messages_mod.rs
```diff
@@ -37,6 +37,21 @@ pub trait SBPMessage: SbpSerialize {
     }
 }
 
+pub trait RealMessage: SBPMessage + std::convert::TryFrom<SBP, Error = TryFromSBPError> + Clone + Sized {
```
notoriaga, line 40: Not sure if `RealMessage` is the best name, but it's what it's called in the code generator

silverjam: `ConcreteMessage`? `KnownMessage`? `ReifiedMessage`?

generator/sbpg/targets/resources/sbp_messages_mod.rs
silverjam: 👍

silverjam (commented):
lgtm, I think adding `MESSAGE_NAME` makes sense and I like `ConcreteMessage` (but I'll defer to whatever name you prefer)

silverjam (approved):

jbangelo (approved):

PR #1017, opened 2021-06-10
Add enum for message type

woodfell:
This PR includes both the most recent commits from the integration branch and merges in the string encoder changes so the diff is a little weird.

Generates a new enum `sbp_msg_type_t` from the SBP specs which is then used in all V4 APIs. The old `#define` still exists with the original names. The enum values are named similarly in SnakeCase (ie `SBP_MSG_OBS` -> `SbpMsgObs`)

There is a fairly large shuffle of things in sbp.c so that `sbp_msg_type_t` is used for all internal functions but the legacy API still uses `u16` everywhere.


jayvdb (approved):

silverjam (approved):

RReichert (approved):
:tada:

c/src/sbp.c
```diff
@@ -840,12 +651,130 @@ s8 sbp_payload_send(sbp_state_t *s, u16 msg_type, u16 sender_id, u8 len, u8 *pay
...
+s8 sbp_message_send(sbp_state_t *s, sbp_msg_type_t msg_type, u16 sender_id, const sbp_msg_t *msg, sbp_write_fn_t write) {
   uint8_t payload[SBP_MAX_PAYLOAD_LEN];
   uint8_t payload_len;
   s8 ret = sbp_message_encode(payload, sizeof(payload), &payload_len, msg_type, msg);
   if (ret != SBP_OK) { return ret; }
-  return sbp_payload_send(s, msg_type, sender_id, payload_len, payload, write);
+  return send_payload(s, msg_type, sender_id, payload_len, payload, write);
+}
+
+s8 sbp_process_message(sbp_state_t *s, u16 sender_id, sbp_msg_type_t msg_type,
```
samvrlewis, line 686: As far as I could tell, we don't have any tests for this, should we?

samvrlewis: Also, slightly confused, maybe I'm completely missing something but how do you process a raw frame with the new v4 API? As far as I can tell this is what `process_frame` does but nothing in the new API calls it?

samvrlewis (approved):
Just a few things that probably don't even apply for this change but have been caught in the diff. Apologies if they've been handled in another PR, finding it a little difficult to keep up with it all.

c/src/sbp.c
```diff
@@ -566,23 +536,25 @@ s8 sbp_process(sbp_state_t *s, s32 (*read)(u8 *buff, u32 n, void *context))
...
       return ret;
     }
     if (s->n_read >= SBP_CRC_LEN) {
       s->state = SBP_WAITING;
       s->crc = sbp_u8_array_to_u16(&(s->frame_buff[SBP_FRAME_OFFSET_CRC(s->msg_len)]));
 
-      sbp_u16_to_u8_array(s->msg_type, (u8*)&buf);
+      u16 raw_msg_type = (u16)s->msg_type;
+      sbp_u16_to_u8_array(raw_msg_type, (u8*)&buf);
+      s->msg_type = (sbp_msg_type_t)raw_msg_type;
```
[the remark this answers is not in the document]
woodfell: `s->msg_type` has changed to be type `sbp_msg_type_t` rather than `u16`, and the size of an enum is implementation defined. It needs to be copied to a temp variable which is definitely `u16` before trying to encode it.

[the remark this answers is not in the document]
samvrlewis: Sorry, was actually asking about line 549 - do we need to copy it back?

c/src/sbp.c
woodfell: You're right, there aren't any tests for this specifically yet.

In the v4 API the idea is to never interact with encoded frames at all. `process_frame` is what's called from `sbp_process` (ie, reading data from the input stream) and also from the legacy API `sbp_frame_process`. The unfortunate duplication of the loop for processing callbacks is because `process_frame` needs to accept an encoded frame which the v4 API will only accept a decoded message as its input. They kind of work opposite from each other as far as encoding/decoding is concerned, but they process callbacks in the same way.

samvrlewis: Got it - thanks! I was missing that both users for the legacy and v4 APIs were able to use `sbp_process` but that makes sense.

c/src/sbp.c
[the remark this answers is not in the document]
woodfell: Oh I see, yeah I think you're right that doesn't need to be there

PR #1043, opened 2021-07-02
change SBP table to use decimal instead of hex [DOC-25]

silverjam:


silverjam:
Example PDF: 
[sbp.pdf](https://github.com/swift-nav/libsbp/files/6756658/sbp.pdf)

switanis:
Perfect!

PR #1045, opened 2021-07-05
Enable VS2019 tests

RReichert:
Just a quick PR to enable VS 2019 unit tests, didn't get around to introducing it last time.

### DIFF
--- a/.github/workflows/c.yaml
+++ b/.github/workflows/c.yaml
@@ -181,8 +181,11 @@ jobs:
           fetch-depth: 0
 
       - name: Configure
-        run: cmake -G "Visual Studio 16 2019" -S c -B build -Dlibsbp_ENABLE_TESTS=false -Dlibsbp_ENABLE_EXAMPLES=false
+        run: cmake -G "Visual Studio 16 2019" -S c -B build -D gtest_force_shared_crt=true
 
       - name: Build
         run: cmake --build build --target ALL_BUILD
+
+      - name: Test
+        run: cmake --build build --target do-all-tests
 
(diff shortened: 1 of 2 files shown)


.github/workflows/c.yaml
```diff
@@ -181,8 +181,11 @@ jobs:
           fetch-depth: 0
 
       - name: Configure
-        run: cmake -G "Visual Studio 16 2019" -S c -B build -Dlibsbp_ENABLE_TESTS=false -Dlibsbp_ENABLE_EXAMPLES=false
+        run: cmake -G "Visual Studio 16 2019" -S c -B build -D gtest_force_shared_crt=true
```
RReichert, line 184: This is required for MSVS, see [here](https://github.com/google/googletest/tree/master/googletest#visual-studio-dynamic-vs-static-runtimes) for details

c/test/CMakeLists.txt
```diff
@@ -25,29 +25,31 @@ if(APPLE)
...
-    check_main_legacy.c
-    check_edc.c
-    check_sbp.c
-    check_bitfield_macros.c
-    ${generated_legacy_c_sources}
-  INCLUDE
-    ${TEST_INCLUDES}
-  LINK
-    ${TEST_LIBS}
+if (NOT CMAKE_C_COMPILER_ID STREQUAL "MSVC")
```
RReichert, line 28: MSVC doesn't support any zero array length and a number of other stuff, hence the removal of this and the legacy C++ library

woodfell (approved):
If we can even build with MSVC we must be doing something right 😅

c/test/CMakeLists.txt
woodfell: Is there a compiler flag for that anywhere? In `common.h` we have some #define for MSVC which suggests there has been at lest historical use of the legacy API

RReichert: Good point, looks like there was evidence to support MSVC - https://github.com/swift-nav/libsbp/pull/568 - but there doesn't seem to be any build system enforcing it. Rerunning the build it was evident that the problem was in `msg_fileio_write_req_t` which had 2 zero array length member variables, which is a problem and something that still surprises me that GCC and Clang are okay with.

PR #1046, opened 2021-07-13
[ESD-2122] Add functions that set and truncate strings

samvrlewis:
Open to suggestions on improving this - in particular I feel like the naming could be improved here to make it more clear what these new functions actually do.

Will hold off committing the new generated code until this has been reviewed, so that it's a little easier to review. This is why the `Generated artifacts / Generated artifacts (pull_request) ` job is currently failing.


c/src/include/libsbp/internal/v4/string/null_terminated.h
```diff
@@ -152,6 +152,21 @@ size_t sbp_null_terminated_string_strlen(const sbp_string_t *s,
...
+ * If the new string contents are larger than can be stored, then as much as
+ * possible will be written from new_str to s
+ *
+ * @param s string
+ * @param maxlen Maximum encoded length
+ * @param new_str New string contents
+ * @param new_str_len Length of the new string, not including NULL terminator (if present)
+ * @return Number of characters written from new_str to s
+ */
+size_t sbp_null_terminated_string_set_truncating(sbp_string_t *s, size_t maxlen,
```
woodfell, line 167: I wonder if a flag to indicate behaviour would be better - `sbp_null_terminated_string_set(sbp_string_t *s, size_t maxlen, bool truncate, const char *new_str, size_t new_str_len)` - @silverjam @jayvdb any thoughts?

c/src/v4/string/unterminated.c
```diff
@@ -72,6 +72,18 @@ bool sbp_unterminated_string_set(sbp_string_t *s, size_t maxlen,
...
+  size_t truncated_len = maxlen > new_str_len ? new_str_len : maxlen;
+  if (!sbp_string_copy_n_to_buf(s->data, &copied, maxlen + 1,
+                              new_str, truncated_len)) {
+    return 0;
+  }
+  s->encoded_len = copied - 1;
+  return s->encoded_len;
+}
+
 bool sbp_unterminated_string_vprintf(sbp_string_t *s, size_t maxlen,
```
woodfell, line 87: Can we get variants for the `printf` family as well? I think that would be very helpful when making log messages in particular

c/src/include/libsbp/internal/v4/string/null_terminated.h
silverjam: Yeah, I think a flag would be fine and avoids the trouble of naming

jayvdb: What return type would this merged function have?  If size_t, would that mean current usage of sbp_null_terminated_string_set becomes more complicated to check for success?

If size_t, it would make sense to explicitly reject lengths of 0, so that a `>1` can be used for success.

samvrlewis: A flag sounds good, will do!

To @jayvdb 's point - do you think it'd be too unwieldy to keep it as a bool and output the number of chars written through a parameter? `bool sbp_null_terminated_string_set(sbp_string_t *s, size_t maxlen, bool truncate, const char *new_str, size_t new_str_len, size_t *written)`

silverjam: I mentioned ssize_t before, can we explore that possibility? Then we don't need an output parameter.

samvrlewis: It turns out `ssize_t` is a POSIX extension, so I've stuck with `size_t` for now. It does have the limitation that you can't distinguish between an error and a successful write of `0` length, but the only way around that (I think) would be to use an output parameter.

silverjam: Ok, what about some other sentinel value to indicate failure?  Like SIZE_MAX?  Or we can return a struct with PODs: `sbp_string_result_t sbp_null_terminated_string_set(sbp_string_t *s, size_t maxlen, bool truncate, const char *new_str, size_t new_str_len)` where `sbp_string_result_t` would be `typedef struct { bool success; size_t written; } sbp_string_result_t;`.

silverjam: It'd also be great if we could eliminate some of the verbosity of these functions with a function table, e.g.:

```
typedef struct {
  sbp_result_t (*set)t(sbp_string_t *s, size_t maxlen, bool truncate, const char *new_str, size_t new_str_len);
  size_t strlen(const sbp_string_t *s, size_t maxlen);
  size_t vprintf(sbp_string_t *s, size_t maxlen, bool should_trunc, const char *fmt, va_list ap);
  ... etc ...
}
```

One of these tables would be created with a constructor such as `sbp_string_ops(s, NullTerminated)->set(...)`.

woodfell: My vote goes for returning `bool` and having nwritten as an output parameter. It's fairly common to return the number of bytes copied/written/read in an output parameter and allows simple conditionals testing for success

```
if (sbp_null_terminated_string_set(args)) {}
```

woodfell (approved):
LGTM!

PR #1049, opened 2021-07-16
don't stop converters when they encounter an error

notoriaga:


rust/sbp2json/src/lib.rs
```diff
@@ -67,7 +67,9 @@ where
         sink.send_all(source)?;
     } else {
         while let Some(msg) = source.next() {
-            sink.send(msg?)?;
+            if let Ok(msg) = msg {
```
silverjam, line 70: We should probably write the error message to stderr (or log with `warn!` so the message can be enabled/disabled appropriately) instead of silently dropping.

silverjam (approved):

PR #1050, opened 2021-07-19
Add orthometric height to MSG_POS_LLH_ACC

jbangelo:
This PR adds an additional field to the `MSG_POS_LLH_ACC` which describes the orthometric height at the rover location. This is also colloquially known as the height above mean sea level, and is the height included in the NMEA GGA sentence. The height above the WGS84 ellipsoid is kept in the message. I didn't include the geoid separation value, which is in the GGA sentence since the user would be able to get the geoid separation by subtracting the `height` value from the `orthometric_height` value. The remaining bits in the `flags` field has also been used to signal what geoid model has been used.


switanis:
Is there any possible use case that ortho height is not present/valid while other LLH are valid? Eg. gap in the geoid model?
I wonder if geoid model flags 0 should mean ortho height invalid / not present (for whatever reason). And non-zero 1 to 7 indicate a model.
Other that this LGTM.

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -660,6 +660,10 @@ definitions:
             type: double
             units: m
             desc: Height above WGS84 ellipsoid
+        - orthometric_height:
```
woodfell, line 663: Should we be deprecating the original definition and creating a new version of this message with the extra field?

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -700,7 +704,10 @@ definitions:
             desc: Status flags
             fields:
               - 6-7:
-                  desc: Reserved
+                  desc: Geoid model
```
woodfell, line 707: Are there more models to add in here in the future? Otherwise bit 7 might not be needed

spec/yaml/swiftnav/sbp/navigation.yaml
woodfell: Answered on slack

silverjam (approved):

spec/yaml/swiftnav/sbp/navigation.yaml
reimerix: Yup, the geoid will be updates as more or better survey data becomes available, EGM2020 seems to be the upcoming release: https://ui.adsabs.harvard.edu/abs/2015AGUFM.G34A..03B/abstract

reimerix (approved):
LGTM!

spec/yaml/swiftnav/sbp/navigation.yaml
richarddeurloo: The upcoming EGM2020 may take up one of the two remaining slots, leaving only one more. Just wondering if this is enough. We could try to rename and reuse part of the `confidence` field, if we want more options.

richarddeurloo (approved):
Just wondering about the number of possible geoid model options. If we don't consider that a problem, than this looks good to me.

spec/yaml/swiftnav/sbp/navigation.yaml
switanis: Yeah, better use 3 bits for the model.

jbangelo: Ok. I'll move this to the `confidence` field, and include a "invalid geoid" option as well. I don't think we're likely to add a model that has invalid regions, but it's safer to leave that possibility open.

spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -679,12 +683,18 @@ definitions:
         - h_ellipse:
             type: EstimatedHorizontalErrorEllipse
             desc: The estimated horizontal error ellipse at the user-configured confidence level.
-        - confidence:
+        - confidence_and_geoid:
             type: u8
             desc: Configured confidence level for the estimated position error
```
richarddeurloo, line 688: We'll need to update the `desc` field as well and regen all the files again ... 😢

richarddeurloo (changes requested):

spec/yaml/swiftnav/sbp/navigation.yaml
jbangelo: Good catch!

jbangelo:
@silverjam It seems like the SonarCloud check is complaining about repeated lines. Is there anything I can do to get that check to go green? Or should we be ignoring it for the moment?

spec/yaml/swiftnav/sbp/navigation.yaml
jbangelo: Updated. How does it look now?

silverjam:
> 
> 
> @silverjam It seems like the SonarCloud check is complaining about repeated lines. Is there anything I can do to get that check to go green? Or should we be ignoring it for the moment?

Fine to ignore for now, I've adjusted the SonarCloud settings so this won't trigger again, if you push a new commit, the SonarCloud check should pass

richarddeurloo (approved):
LGTM

PR #1053, opened 2021-07-23
ESD-2131 swift targets

RReichert:
## Changes

Similar to https://github.com/swift-nav/libswiftnav-private/pull/281.

**NOTE**: the example `add_executable`s have been left alone

## Testing

No difference


c/src/CMakeLists.txt
```diff
@@ -11,15 +11,17 @@ set(libsbp_SRCS
...
+swift_add_library(sbp
+  C_STANDARD 99
+  SOURCES ${libsbp_SRCS}
+  ADD_COMPILE_OPTIONS
+    -pedantic
+)
 add_library(swiftnav::sbp ALIAS sbp)
 
 target_include_directories(sbp PUBLIC ${PROJECT_SOURCE_DIR}/include PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/include)
 set_target_properties(sbp PROPERTIES
```
woodfell, line 23: I know this has been hanging around for ages but do you happen to know if there is any reason to keep it?

woodfell (approved):

c/src/CMakeLists.txt
RReichert: no idea why its around, noticed similar commands in `starling` exists. will add as part of tech debt

PR #1054, opened 2021-07-23
[ESD-2139] Add method to process unpacked messages from the state class

samvrlewis:


c/src/sbp.c
```diff
@@ -682,7 +682,7 @@ s8 sbp_message_send(sbp_state_t *s, sbp_msg_type_t msg_type, u16 sender_id, cons
   return send_payload(s, msg_type, sender_id, payload_len, payload, write);
 }
 
-s8 sbp_process_message(sbp_state_t *s, u16 sender_id, sbp_msg_type_t msg_type,
+s8 sbp_message_process(sbp_state_t *s, u16 sender_id, sbp_msg_type_t msg_type,
```
samvrlewis, line 685: I seem to remember there being a reason that this name was different to the one in api.h, but I can't remember why and, as far as I can tell, there doesn't seem to be anything linking the two any more. Should this be fine to rename?

woodfell: Yeah I'm not sure why this was backwards, `sbp_message_process` is correct

woodfell (approved):
LGTM!

silverjam (approved):

PR #1058, opened 2021-07-29
Fix spelling and copyright year [DEVINFRA-494]

jayvdb:


spec/yaml/swiftnav/sbp/file_io.yaml
```diff
@@ -32,7 +32,7 @@ definitions:
       The file read message reads a certain length (up to 255 bytes)
       from a given offset into a file, and returns the data in a
       MSG_FILEIO_READ_RESP message where the message length field
-      indicates how many bytes were succesfully read.The sequence
+      indicates how many bytes were successfully read.The sequence
```
woodfell, line 35:
```suggestion
      indicates how many bytes were successfully read.The sequence
```

spec/yaml/swiftnav/sbp/settings.yaml
```diff
@@ -121,7 +121,7 @@ definitions:
     id: 0x00A5
     short_desc: Read device configuration settings (host <= device)
     desc: >
-      The setting message wich which the device responds after a
+      The setting message which which the device responds after a
```
woodfell, line 124:
```suggestion
      The setting message with which the device responds after a
```

jayvdb:
There are still a few more old copyright lines (findable with `git grep 'Copyright.*Swift' | grep --color=always -v 2021`), but these should be enough to surface any issues.

silverjam (approved):

jbangelo (approved):

PR #1059, opened 2021-07-31
Dockerfile: Use Ubuntu focal

jayvdb:
https://github.com/swift-nav/libsbp/pull/1058 depends on having a word dictionary, and the bionic dictionaries are missing words used in the specs that are in present in the GitHub ubuntu-latest workers used by https://github.com/swift-nav/libsbp/blob/641eda5/.github/workflows/generator.yaml#L20  .
The more recent the default wordlists used in the docker, the lower the chance that spec updates will require adding to the custom wordlist.

focal is the last ubuntu which includes https://packages.ubuntu.com/focal/clang-format-6.0 ; that could be sourced from somewhere else, but there are other issues to be solved moving to a release after focal.  Not especially relevant at present, is this is also the last release which supports https://packages.ubuntu.com/focal/clang-6.0 .  The portability project should mean that gcc-8 and beyond are usable with the current warning settings, so more flexibility should be coming soon.

Using groovy or hirsute instead of focal(LTS) will mean a less stable base, and requires an alternative for the deadsnakes ppa which only supports LTS.
c.f. https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa & https://github.com/deadsnakes/issues/issues/169

* Previous image size: 8.63GB per https://github.com/swift-nav/libsbp/runs/3089280215?check_suite_focus=true
* Switch to focal: 9.42GB per https://github.com/swift-nav/libsbp/pull/1059/checks?check_run_id=3208048893
* Remove rust-docs, `/var/lib/apt/lists/` and use apt-get `--no-install-recommends` (for secondary fetches) https://github.com/swift-nav/libsbp/pull/1059/commits/88818e06a5fbc3cb4025d58f4fa0b493d0ec8501 : 8.44GB
* Use apt-get `--no-install-recommends` (primary): 7.71Gb https://github.com/swift-nav/libsbp/runs/3209384503?check_suite_focus=true
* Remove deadsnakes `*-dev`: 7.61GB - https://github.com/swift-nav/libsbp/pull/1059/checks?check_run_id=3209737607#step:5:15147 - tox summary
* Merge apt-get: 7.56GB https://github.com/swift-nav/libsbp/runs/3213388487?check_suite_focus=true

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,10 +76,10 @@ RUN \
       libpython3.8-stdlib \
       python-is-python3 \
       python3-pip \
-      python3.5-dev \
-      python3.6-dev \
-      python3.7-dev \
-      python3.9-dev python3.9-dist \
+      python3.5 \
+      python3.6 \
+      python3.7\
+      python3.9 \
   && pip3 install tox sphinx tox-run-command \
   && curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain stable --profile minimal --no-modify-path \
   && rustup component add rustfmt \
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,9 @@ ENV PATH=/usr/lib/ccache:/cargo/bin:/rust/bin:${PATH}
 
 RUN \
      apt-get update \
-  && apt-get install -y \
+  && apt-get install -y --no-install-recommends \
+      gnupg \
+      netbase \
       git \
       sudo \
       software-properties-common \
@@ -39,6 +41,7 @@ RUN \
       curl \
       libudev-dev \
       uuid-dev \
+      libffi-dev \
       libgmp-dev \
       zlib1g-dev \
       zip unzip \
@@ -47,14 +50,14 @@ RUN \
       pandoc \
       llvm \
       clang \
-      texlive-science \
       texlive-fonts-extra \
+      texlive-latex-extra \
+      texlive-science \
       check \
       ccache \
       pkg-config \
       doxygen \
       graphviz \
-      texlive-latex-base \
       imagemagick \
       clang-format-6.0 \
   && rm -rf /var/lib/apt/lists/* \
@@ -71,7 +74,7 @@ RUN \
   && gradle --version \
   && add-apt-repository ppa:deadsnakes/ppa \
   && apt-get update \
-  && apt-get install -y \
+  && apt-get install -y --no-install-recommends \
       libpython2.7-stdlib \
       libpython3.8-stdlib \
       python-is-python3 \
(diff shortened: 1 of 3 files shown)


silverjam (approved):

silverjam:
As long as this doesn't change the base set of tools and languages we support I suppose this is ok.  Originally the usage of bionic in docker (or at least my intent) was supposed to imply a check on the base minimum Linux version we support.  Bumping to focal means we don't have that sort of check anymore.  It would be great if we could think of a way to mitigate this.

silverjam:
> As long as this doesn't change the base set of tools and languages we support I suppose this is ok, but a bit concerned that originally the usage of bionic in docker (or at least my intent) was supposed to imply a check on the base Linux version we support.  Bumping to focal means we don't have that sort of check anymore.  It would be great if we could think of a way to mitigate this.

One idea would be to just run the check in CI-- if focal is needed to run correctly locally-- let's just make a dockerfile for that.

Another idea would be to make the dockerfile support bionic or focal as a base.

jayvdb:
> Another idea would be to make the dockerfile support bionic or focal as a base.

Done.  The bionic image is 6.82GB, vs 7.56GB for focal.  Should I add CI to test bionic as well?  

In addition to the spelling database, having everyone use the focal LTS image will likely result in a better generated `sbp.pdf`.

Dockerfile
```diff
@@ -66,29 +124,9 @@ ENV PATH=${SDKMAN_DIR}/candidates/gradle/current/bin:${PATH}
...
-
-ARG KITWARE_KEY_URL=https://apt.kitware.com/keys/kitware-archive-latest.asc
-
-RUN \
-     wget -O - ${KITWARE_KEY_URL} 2>/dev/null | gpg --dearmor - | sudo tee /etc/apt/trusted.gpg.d/kitware.gpg >/dev/null \
-  && add-apt-repository 'deb https://apt.kitware.com/ubuntu/ bionic main' \ 
-  && apt-get update \
-  && apt-get install -y \
-    cmake
+  && curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain stable --profile minimal --no-modify-path \
```
triarius, line 128: Do we do anything to pin the version (as opposed to the channel) of the rust toolchain we use? It should be alright to compile older projects with newer compilers because of the edition guarantees, so I don't think this is critical.

triarius (approved):
LGTM!

Dockerfile
jayvdb: This project seems to be ok with using stable-vs-nightly and edition as the only rustc versioning

PR #1064, opened 2021-08-12
ESD-2138 C11 -> C99

RReichert:


c/src/CMakeLists.txt
```diff
@@ -12,7 +12,6 @@ set(libsbp_SRCS
   )
 
 swift_add_library(sbp
-  C_STANDARD 99
```
RReichert, line 15: no longer needed as the standard will be to use C99 by default from now on.

woodfell (approved):

PR #1066, opened 2021-08-30
ESD-2195 buffer overflow on SBP string decoding

RReichert:
## Changes
Fixes a buffer overflow issue that takes place in SBP string decoding

## Related PR
* https://github.com/swift-nav/gnss-converters-private/pull/944

### DIFF
--- a/c/src/v4/string/sbp_string.c
+++ b/c/src/v4/string/sbp_string.c
@@ -92,6 +92,9 @@ bool sbp_string_decode(sbp_string_t *s, size_t maxlen,
   if (available > maxlen) {
     return false;
   }
+  if (params->inject_missing_terminator && available == 0) {
+    return false;
+  }
   memcpy(s->data, ctx->buf + ctx->offset, available);
   s->encoded_len = available;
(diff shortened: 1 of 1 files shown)


c/src/v4/string/sbp_string.c
```diff
@@ -92,6 +92,9 @@ bool sbp_string_decode(sbp_string_t *s, size_t maxlen,
   if (available > maxlen) {
     return false;
   }
+  if (params->inject_missing_terminator && available == 0) {
```
RReichert, line 95: If a user attempts to decode a multipart string (null terminating string will also present with the same issue) with zero available bytes left to read from the ctx, than there will be a buffer overflow taking place in the statement `s->data[s->encoded_len - 1]` (since `encoded_len` is of type `size_t` it will wrap around and give the highest value, hence overflow).

To fix this I went with early error reporitng approach rather than skipping the `if (params->inject_missing_terminator)` block and letting `if (!params->valid(s, maxlen))` block report the error.

woodfell (approved):
Nice find

PR #1070, opened 2021-09-09
Rust: callback based message handler

notoriaga:


rust/sbp/src/link.rs
```diff
@@ -0,0 +1,369 @@
...
+}
+
+impl<T> Event for T
+where
+    T: ConcreteMessage,
+{
+    const MESSAGE_TYPES: &'static [u16] = &[T::MESSAGE_TYPE];
+
+    fn from_sbp(msg: SBP) -> Self {
+        msg.try_into().unwrap()
```
silverjam, line 228: Should this be fallible?  At least should use `.expect` unless we never expect `.unwrap` to fail?

rust/sbp/src/link.rs
```diff
@@ -0,0 +1,369 @@
...
+
+    pub fn link(&self) -> Link<'link, S> {
+        self.link.clone()
+    }
+
+    pub fn stateless_link(&self) -> Link<'link, ()> {
+        self.stateless_link.clone()
+    }
+
+    pub fn send_state<M>(&self, state: &S, msg: M) -> bool
```
silverjam, line 35:
```suggestion
    pub fn send_with_state<M>(&self, state: &S, msg: M) -> bool
```

Maybe?

rust/sbp/src/link.rs
```diff
@@ -0,0 +1,369 @@
...
+                    _ => unreachable!("wrong event keys"),
+                }
+            }
+        }
+
+        let source = LinkSource::new();
+        let link = source.link();
+        let count = RefCell::new(0);
+
+        link.register(|count: &RefCell<usize>, _: SBP| {
```
silverjam, line 333: Should this test receiving an ObsMsg struct?

rust/sbp/src/link.rs
notoriaga: It shouldn't ever fail, because `from_sbp` is only called after checking that the message was in `MESSAGE_TYPES`. So if you create a custom event and set the wrong messages types you could hit this error. So an expect is probably a good idea to make that more clear

silverjam: Agree, expect could point out that it should only happen if there's a programming error

rust/sbp/src/link.rs
notoriaga: Yup good catch

rust/sbp/src/link.rs
notoriaga: :+1:

rust/sbp/src/link.rs
notoriaga: Well so actually this shouldn't ever error, this blanket impl makes all `ConcreteMessage`s usable as events triggered by that messages type. So baring a logic error in the Link (i.e. sending an event to the wrong handler or something) this should never happen. Replaced the unwrap with a panic that will include the offending message just in case

silverjam: 👍

silverjam (approved):

PR #1071, opened 2021-09-17
rust: release cleanup

notoriaga:
Apologies in advance for the massive diff. Most of it is renaming/moving things around. The bulk of new changes are in `wire_format.rs` and `sbp_string.rs`.

Breaking API changes:
- Removes `get_` prefix from getters - https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter
- Renames `SBP` -> `Sbp` and `SBPMessage` -> `SbpMessage`. Also treats acronyms like GPS/NED as one word (i.e. Gps/Ned) - https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case. The casing of JSON messages was left as is
- Renames `sbp_serde` feature to just `serde` - https://rust-lang.github.io/api-guidelines/naming.html?highlight=fea#feature-names-are-free-of-placeholder-words-c-feature
- Replaces the `Error` enum with two, `DeserializeError`/`SerializeError`. The finer grained types prevent you from having to deal with errors that can never happen
- Marks Sbp enum as `#[non_exhaustive]` (not that anyone is realistically matching on all variants but you never know)
- No longer reexports stuff from `dencode`/`bytes`. Reading messages is done via `iter_messages` like before. Writing messages now can be done with `sbp::to_writer` or `sbp::to_vec`. `SbpEncoder` is available for writing messages with buffered output. Previously you needed to use `dencode::FramedWrite` with the encoder, but now each encoder contains the `FramedWrite`
- `SBPTools` -> `SbpIterExt` because that's what it is/that's normally what I see things like this called

Additions:
- Adds preliminary support for the new string encodings.
- Adds support for fixed size arrays (similar to what @jbangelo had here https://github.com/swift-nav/libsbp/pull/903). Haven't seen much of a perf change, but maybe it might help with the fragmentation issues @silverjam mentioned. Also feels more semantically correct to say that something like a velocity vector is an `[f64; 3]` not a `Vec<f64>`
- Doc comments/some doc tests

Internal improvements:
- Fixes generated docs to escape braces/wrap links in `<link>`
- Rejiggers the auto generated tests to produce a single test binary rather than one for each
- Removes all but one of the `#[allow(..)]` statements in the generated code by tweaking the code gen
- Removes `nom`/`thiserror`/`byteorder`. Without any features enabled we dropped from 46 deps to 10, debug builds drop from ~22 -> ~9 seconds and release builds from ~43 -> ~11 seconds. It turns out `bytes` has all we need for parsing (mainly an easy way to pull LE bytes out of a buffer). 
- Removes `SbpSerialize` trait and replaces it with `WireFormat` (not sure this is the best name). It's very similar to SbpSerialize but also includes functions for reading the types. Unlike `SbpSerialize` this trait is not public. Well it's technically public but unnameable by downstream crates (https://rust-lang.github.io/api-guidelines/future-proofing.html). This was done because it's really an implementation detail. Plus, because `WireFormat` is a super trait of `SbpMessage`, it makes it impossible for downstream crates to implement `SbpMessage` which is probably a good idea.

Future things:
- Expand on the SbpString stuff. Should add validation when parsing/writing that the string satisfies the format. Could also add functions to iterate over the parts of a multipart string for example.
- When the docs refer to other messages like we could replace the names with links to the actual types.

All in all a lot of changes. Happy to undo any of them that people don't like/think are unnecessary


rust/sbp/src/wire_format.rs
```diff
@@ -0,0 +1,210 @@
+//! Functionality for parsing/writing SBP message payloads with in memory buffers
```
notoriaga, line 1: Some actual new-ish code. A combination of `SbpSerialize` and all the free standing `parse_<type>()` functions

rust/sbp/src/wire_format.rs
```diff
@@ -0,0 +1,210 @@
...
+{
+    const MIN_ENCODED_LEN: usize = T::MIN_ENCODED_LEN * LEN;
+
+    fn write(&self, buf: &mut BytesMut) {
+        for item in self {
+            item.write(buf);
+        }
+    }
+
+    fn parse_unchecked(buf: &mut BytesMut) -> Self {
```
notoriaga, line 74: Probably a premature optimization. But, this is pretty much the example right out of the docs for `MaybeUninit` so I don't feel too bad about this touch of unsafe

rust/sbp/src/sbp_string.rs
```diff
@@ -0,0 +1,229 @@
+use std::fmt;
```
notoriaga, line 1: The other chunk of actually new stuff

generator/sbpg/targets/resources/sbp_messages_mod.rs
```diff
@@ -22,26 +23,49 @@ use self::(((p.identifier|mod_name)))::(((m.identifier|camel_case)));
 ((*- endfor *))
 use self::unknown::Unknown;
 
-use crate::serialize::SbpSerialize;
+mod lib {
```
jbangelo, line 26: Should this be called a prelude? Or is that term just for common imports from an external crate?

generator/sbpg/targets/resources/sbp_messages_mod.rs
```diff
@@ -22,26 +23,49 @@ use self::(((p.identifier|mod_name)))::(((m.identifier|camel_case)));
...
-    fn gps_time(&self) -> Option<std::result::Result<crate::time::MessageTime, crate::time::GpsTimeError>> {
+    fn gps_time(&self) -> Option<Result<crate::time::MessageTime, crate::time::GpsTimeError>> {
         None
     }
 }
 
-pub trait ConcreteMessage: SbpMessage + std::convert::TryFrom<Sbp, Error = TryFromSbpError> + Clone + Sized {
+/// Implemented by messages who's message name and type are known at compile time.
+/// This is everything that implements [SbpMessage] except for [Sbp].
+pub trait ConcreteMessage: SbpMessage + TryFrom<Sbp, Error = TryFromSbpError> {
```
jbangelo, line 61: If `ConcreteMessage` is only for use in the sbp crate, should this be marked as `pub(crate)`?

silverjam:
>     * Adds support for fixed size arrays (similar to what @jbangelo had here [[WIP] Take advantage of const generics in the Rust parser #903](https://github.com/swift-nav/libsbp/pull/903)). Haven't seen much of a perf change, but maybe it might help with the fragmentation issues @silverjam mentioned. Also feels more semantically correct to say that something like a velocity vector is an `[f64; 3]` not a `Vec<f64>`

Semantics look a lot better, but I think we'd need further changes for this to have a real impact on (potential) heap fragmentation issues. "Further changes" would probably mean the ability to allocate the whole SBP message struct on the stack with variable sized arrays backed by an "upper maximum" static array (similar to the new v4 libsbp API for C).

I tried to make a change to ensure that we're allocating consistently size buffers here: https://github.com/swift-nav/libsbp/commit/64cafc3d685269875821e08164e14fdd5b4eeb25 to make things easier for heap reclamation -- but it did not have any impact on the specific console issue (which probably means it's just a leak, not a fragmentation issue).

silverjam (approved):
New code looks good

jbangelo (approved):
Changes look good to me.

generator/sbpg/targets/resources/sbp_messages_mod.rs
notoriaga: Hmm it can't be marked private because it a appears in `Event` which needs to be public. Could maybe add `#[doc(hidden)]`. Depends on if we think having access to message names/ids at compile time is useful. Right now this is the only way to get that

generator/sbpg/targets/resources/sbp_messages_mod.rs
notoriaga: Hmm yeah I just took the idea from serde_json where they call it lib - https://docs.serde.rs/src/serde_json/lib.rs.html#366
Although they are doing it for a slightly different reason (still internal though)

generator/sbpg/targets/resources/sbp_tests_template.rs
```diff
@@ -51,15 +44,15 @@ fn test_(((s.suite_name)))()
...
+            sbp::messages::Sbp::(((t.msg.name|lower_acronyms)))(msg) => {
+                assert_eq!( msg.message_type(), (((t.msg_type))), "Incorrect message type, expected (((t.msg_type))), is {}", msg.message_type());
+                let sender_id = msg.sender_id().unwrap();
                 assert_eq!(sender_id, (((t.sbp.sender))), "incorrect sender id, expected (((t.sbp.sender))), is {}", sender_id);
                 ((*- for f in t.fieldskeys *))(((compare_value( (((f))), (((t.fields[f]))) ))))((*- endfor *))
             },
             _ => panic!("Invalid message type! Expected a (((t.msg.name)))"),
         };
-        let frame = sbp_msg.to_frame().unwrap();
+        let frame = sbp::to_vec(&sbp_msg).unwrap();
```
john-michaelburke, line 55: nit. What is the idiomatic rule of thumb for using to_asdf vs as_asdf? Is it related to the cost of the action such as a string to slice using "as" vs a vector to string use "to"

generator/sbpg/targets/rust.py
```diff
@@ -100,7 +116,27 @@ def commentify(environment: Environment,
...
+  for url in urls:
+    if url.startswith("(") and url.endswith(")"):
+      continue
+    if url.endswith(")"):
+      s = s.replace(url, "<" + url[:-1] + ">)")
+    else:
+      s = s.replace(url, "<" + url + ">")
+  return s
+
+def escape_braces(s):
```
john-michaelburke, line 135: Maybe for a future addition, it would be neat to see if we could somehow get python docstrings embedded into the rust generated docs seeing as how these are related to the crate generation.

john-michaelburke (approved):

generator/sbpg/targets/resources/sbp_tests_template.rs
notoriaga: Yeah pretty much - `as_` should be cheap and not consume the original. `to_` can be more expensive and still does not consume the original. Then `into_` can be either I guess but it should consume the original value. So in this case I went with `to_` because it allocates

generator/sbpg/targets/rust.py
notoriaga: Yeah kind of like how the top of each message file tells you which yaml file it came from

PR #1074, opened 2021-09-24
[ESD-2180] clang-tidy improvements

woodfell:
Import new clang-tidy module.

Remove `.clang-tidy`, it's now automatically generated by `ClangTidy.cmake`

Any additional changes needed by new common clang-tidy config


samvrlewis (approved):
Do our clang tidy rules not like 'str' and 'truncate' being in variable names?

woodfell:
> Do our clang tidy rules not like 'str' and 'truncate' being in variable names?

No, it was complaining about different names in the function declaration and definition. I went with the policy of using `buf` for the "raw" function saving `str` for the non-raw functions, and `trunc` rather than `truncate` since it was more common

silverjam (approved):

RReichert (approved):

isaactorz (approved):
LGTM

PR #1078, opened 2021-10-01
rust: add len method to SbpMessage

notoriaga:
While updating the console realized that it was using the `sbp_size`method (now called `encoded_len`). This adds a `len` method to `SbpMessage` (`WireFormat` isn't usable by downstream crates). This function is actually a little different, it adds on the 8 bytes for the header + crc, which sbp_size did not. I think this is probably what we actually want in the console, in order to show how much data the console has read


john-michaelburke (approved):
Finally an explanation for the discrepancy in data rate between new and old console 😄

silverjam (approved):

silverjam (changes requested):
Actually I think our nomenclature could use some consideration here.  I think usually "encoded len" is used to indicate how big the byte buffer would be if the message was encoded into it's "on wire" representation.

silverjam:
> Actually I think our nomenclature could use some consideration here. I think usually "encoded len" is used to indicate how big the byte buffer would be if the message was encoded into it's "on wire" representation.

See:
- https://developer.apple.com/documentation/metal/mtlargumentencoder/2915784-encodedlength
- https://guava.dev/releases/18.0/api/docs/com/google/common/base/Utf8.html

So maybe we should switch around `len` and `encoded_len` or rename `len` to `wire_len` or `packed_len`

notoriaga:
Great idea, pulled the switcheroo

silverjam (approved):

PR #1079, opened 2021-10-01
Add SbpMsgInvalid to sbp_msg_types_t

lkloh:
We've discussed having `SbpMsgInvalid` enum type so that we can use it as a placeholder for variables of the`sbp_msg_types_t` type instead of using `static_cast<sbp_msg_type_t>(0)` as the invalid type.

### DIFF
--- a/c/include/libsbp/sbp_msg_type.h
+++ b/c/include/libsbp/sbp_msg_type.h
@@ -47,6 +47,7 @@ extern "C" {
 #endif
 
 typedef enum {
+  SbpMsgInvalid = -1,
   SbpMsgAcqResultDepA = SBP_MSG_ACQ_RESULT_DEP_A,
   SbpMsgAcqResultDepB = SBP_MSG_ACQ_RESULT_DEP_B,
   SbpMsgAcqResultDepC = SBP_MSG_ACQ_RESULT_DEP_C,
--- a/c/include/libsbp/v4/sbp_msg.h
+++ b/c/include/libsbp/v4/sbp_msg.h
@@ -1862,7 +1862,7 @@ static inline int sbp_message_cmp(sbp_msg_type_t msg_type, const sbp_msg_t *a,
                                   const sbp_msg_t *b) {
   switch (msg_type) {
     case SbpMsgInvalid:
-      return -1;
+      return 0;
     case SbpMsgAcqResultDepA:
       return sbp_msg_acq_result_dep_a_cmp(&a->acq_result_dep_a,
                                           &b->acq_result_dep_a);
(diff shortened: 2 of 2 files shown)


c/include/libsbp/sbp_msg_type.h
```diff
@@ -47,6 +47,7 @@ extern "C" {
 #endif
 
 typedef enum {
+  SbpMsgInvalid = -1,
```
lkloh, line 50: I define the invalid enum to be "-1", following this pattern elsewhere in our codebase:

https://github.com/swift-nav/libswiftnav/blob/c580b9d041a7b6727ea7efdfa3fc9e6443e902f0/include/swiftnav/signal.h#L177

c/include/libsbp/v4/sbp_msg.h
```diff
@@ -277,6 +277,8 @@ static inline s8 sbp_message_encode(uint8_t *buf, uint8_t len,
                                     uint8_t *n_written, sbp_msg_type_t msg_type,
                                     const sbp_msg_t *msg) {
   switch (msg_type) {
+    case SbpMsgInvalid:
```
lkloh, line 280: This is in `sbp_message_encode` which returns `-1` in the `default` condition

c/include/libsbp/v4/sbp_msg.h
```diff
@@ -845,6 +847,8 @@ static inline s8 sbp_message_decode(const uint8_t *buf, uint8_t len,
                                     uint8_t *n_read, sbp_msg_type_t msg_type,
                                     sbp_msg_t *msg) {
   switch (msg_type) {
+    case SbpMsgInvalid:
```
lkloh, line 850: This is in sbp_message_decode which returns -1 in the default condition

c/include/libsbp/v4/sbp_msg.h
```diff
@@ -1397,6 +1401,8 @@ static inline s8 sbp_message_decode(const uint8_t *buf, uint8_t len,
 static inline size_t sbp_message_encoded_len(sbp_msg_type_t msg_type,
                                              const sbp_msg_t *msg) {
   switch (msg_type) {
+    case SbpMsgInvalid:
```
lkloh, line 1404: `sbp_message_encoded_len` returns 0 in the `default` condition

c/include/libsbp/sbp_msg_type.h
woodfell: All of these files are autogenerated and will be overwritten then next time the schema is modified and needs to be regenerated. To avoid this you'll need to add these to the templates in `generator/sbpg/target/resources/c/` and run `make c` in the repo root

c/include/libsbp/v4/sbp_msg.h
```diff
@@ -1855,6 +1861,8 @@ static inline size_t sbp_message_encoded_len(sbp_msg_type_t msg_type,
 static inline int sbp_message_cmp(sbp_msg_type_t msg_type, const sbp_msg_t *a,
                                   const sbp_msg_t *b) {
   switch (msg_type) {
+    case SbpMsgInvalid:
+      return 0;
```
lkloh, line 1865: This is in `sbp_message_cmp`; return 0 to match the behavior of its `default` case

PR #1081, opened 2021-10-12
fix crc calculation bug [DEVINFRA-568]

silverjam:
CRC calculation was only correct for the first packet in a buffer because of a missing offset


silverjam:
CC @switanis Silly CRC bug was causing the slowdown in the Console 3.1 release

PR #1082, opened 2021-10-15
Mark all packages as stable

woodfell:
Several packages in the yaml spec are marked as unstable. This means that all the messages defined in that package are output in a "draft message" section at the end of the pdf doc.

Mark all packages as stable since anything that is out there in the wild can't be changed. The only effect of this is that the pdf doc needs updating but there are no code changes anywhere.


spec/yaml/swiftnav/sbp/file_io.yaml
```diff
@@ -19,7 +19,7 @@ description: >
 
   Note that some of these messages share the same message type ID for both the
   host request and the device response.
-stable: False
+stable: True
```
silverjam, line 22: I don't think we should do this for fileio, the feature was never officially released

spec/yaml/swiftnav/sbp/flash.yaml
```diff
@@ -15,7 +15,7 @@ description: >
   Swift Navigation devices: the STM32 flash and the M25Pxx FPGA
   configuration flash from Piksi 2.3.1.  This module does not apply
   to Piksi Multi.
-stable: False
+stable: True
```
silverjam, line 18: Same here, this is for Piksi v2, so it has little value to current customers so I don't see any benefit in marking it as stable, we should probably add a "deprecated" section instead of promoting this to stable.

spec/yaml/swiftnav/sbp/linux.yaml
```diff
@@ -12,7 +12,7 @@ package: swiftnav.sbp.linux
 description: >
   Linux state monitoring.
 
-stable: False
+stable: True
```
silverjam, line 15: This module is also not part of an officially released feature

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -10,7 +10,7 @@
 
 package: swiftnav.sbp.ssr
 description: Precise State Space Representation (SSR) corrections format
-stable: False
+stable: True
```
silverjam, line 13: @jbangelo Should we mark this as stable?

spec/yaml/swiftnav/sbp/orientation.yaml
```diff
@@ -10,7 +10,7 @@
 
 package: swiftnav.sbp.orientation
 description: Orientation Messages
-stable: False
+stable: True
```
jbangelo, line 13: @reimerix Are these output messages stable in your opinion? I think `MSG_BASELINE_HEADING` is the only one from the GNSS engine and it hasn't needed updating in ages.

spec/yaml/swiftnav/sbp/ssr.yaml
jbangelo: I think so. There might be changes in the future, but we don't have anything planned in the foreseeable future and these messages have been doing well so far.

spec/yaml/swiftnav/sbp/vehicle.yaml
```diff
@@ -10,7 +10,7 @@
 
 package: swiftnav.sbp.vehicle
 description: Messages from a vehicle.
-stable: False
+stable: True
```
jbangelo, line 13: @reimerix , another set of sensor fusion specific messages here.

spec/yaml/swiftnav/sbp/ndb.yaml
```diff
@@ -11,7 +11,7 @@
 package: swiftnav.sbp.ndb
 description: >
   Messages for logging NDB events.
-stable: False
+stable: True
```
jbangelo, line 14: If we're being conservative I would vote to leave these and not stable. I'm not sure what their use is, or even if Piksi Multi produces them.

spec/yaml/swiftnav/sbp/orientation.yaml
jbangelo: I'm not sure if the Duro Inertial outputs these. The descriptions imply that it shouldn't, but I could have sworn it did. Either way I think it would be fine to deprecate these in the future if there's a need to output something equivalent but we don't think these definitions are acceptable.

woodfell:
Ok so I've tried to do everything requested. file_io and linux are now marked as unstable, and file_io, linux, ndb, flash are all marked as non-public. I've also made a small change to the pdf generator so that it doesn't output a Draft Message Definition section when there aren't any entries in it.

New page of contents looks like this
Screenshot_20211022_065716
:

silverjam (commented):
Adding a review to clear this off the DevInfra queue, we have product team request to fix-up SBP docs so we'll defer to that request before we resolve this one

PR #1083, opened 2021-10-20
[ESD-2242] Fix integer promotion issues in generated macros

woodfell:
The generated bitfield macros contains lots of integer promotion when operating on smaller data types. These can cause compiler warnings in consumers of libsbp when compiled with `-Wconversion`

### DIFF
--- a/generator/sbpg/targets/c.py
+++ b/generator/sbpg/targets/c.py
@@ -321,9 +321,9 @@ def create_bitfield_macros(field, msg):
             ret_list.append("#define {}_SHIFT ({}u)".format(base_string, start_bit))
             ret_list.append(
                 """#define {}_GET(flags) \\
-                             (((flags) >> {}_SHIFT) \\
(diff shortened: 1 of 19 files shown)


generator/sbpg/targets/c.py
```diff
@@ -321,18 +321,18 @@ def create_bitfield_macros(field, msg):
             ret_list.append("#define {}_SHIFT ({}u)".format(base_string, start_bit))
             ret_list.append(
                 """#define {}_GET(flags) \\
-                             (((flags) >> {}_SHIFT) \\
-                             & {}_MASK)""".format(
-                    base_string, base_string, base_string
+                             (({})(((flags) >> {}_SHIFT) \\
```
woodfell, line 324: This is the only interesting change, everything else is generated

silverjam (approved):

samvrlewis (approved):

PR #1084, opened 2021-10-21
rust: support Result<Sbp> in with_rover_time

notoriaga:
Previously you had to convert the iterator returned by `iter_messages` from `Iterator<Result<Sbp>>` to `Iterator<Sbp>` before calling `with_rover_time`. Now it works on both kinds of iterators

### DIFF
--- a/rust/sbp/src/sbp_iter_ext.rs
+++ b/rust/sbp/src/sbp_iter_ext.rs
@@ -233,7 +233,7 @@ mod tests {
 
         let gps_time = with_time.next().map(|(_, t)| t).flatten().unwrap().unwrap();
         assert_eq!(gps_time.wn(), 1787);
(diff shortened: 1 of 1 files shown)


rust/sbp/src/sbp_iter_ext.rs
```diff
@@ -123,66 +115,93 @@ where
...
+                },
+                Some(Err(e)) => Some((msg, Some(Err(e)))),
+                None => Some((msg, None)),
+            }
+        }
+    }
 
-    fn next(&mut self) -> Option<Self::Item> {
-        let msg = self.messages.next()?;
+    pub trait HasTime {
```
notoriaga, line 189: Kind of a clunky solution but I don't know a better way to do this without specialization

PR #1085, opened 2021-11-02
[ESD-2250] Encode/decode functions in cpp message traits

woodfell:
While it was originally thought that there wouldn't be a need to deal with encoded messages using the V4 API the ephemeris swiftlet work introduced a need with how it dumps the see after startup. 

Add wrappers for the message type specific encode/decode/length functions to c++ message traits


RReichert (approved):
LGTM

samvrlewis (approved):

antrikshsrivastava (approved):

silverjam (approved):
Done with CI fixes, let's wait for CI to pass before merging (it should pass, but would like to double check)

PR #1091, opened 2021-12-11
fix bit masks for new COG message

silverjam:


lloydmaza (approved):
Oof... the simplest bit logic remains my mortal enemy. Thanks for fixing this one.

PR #1098, opened 2022-02-22
Make integer to enum conversion explicit [SAR-261]

fpezzinosn:
The signature of the function `sbp_register_callback_generic` defined [here](https://github.com/swift-nav/libsbp/blob/8ec25d0e9dc8b22403a1313e5c4d8ca81fd63b89/c/src/sbp.c#L180-L183) expects an `sbp_msg_type_t` enum as a second parameter. The same happens with the third parameter of `process_frame` defined [here](https://github.com/swift-nav/libsbp/blob/8ec25d0e9dc8b22403a1313e5c4d8ca81fd63b89/c/src/sbp.c#L373-L376).

Without this explicit conversion, the static analyzer thinks there are two different functions with the same name.

### DIFF
--- a/c/include/libsbp/v4/sbp_msg.h
+++ b/c/include/libsbp/v4/sbp_msg.h
@@ -831,8 +831,6 @@ static inline s8 sbp_message_encode(uint8_t *buf, uint8_t len,
       return sbp_msg_vel_ned_encode(buf, len, n_written, &msg->vel_ned);
     case SbpMsgWheeltick:
       return sbp_msg_wheeltick_encode(buf, len, n_written, &msg->wheeltick);
-    case SbpMsgAll:
-      break;
     default:
       break;
   }
@@ -1395,8 +1393,6 @@ static inline s8 sbp_message_decode(const uint8_t *buf, uint8_t len,
       return sbp_msg_vel_ned_decode(buf, len, n_read, &msg->vel_ned);
     case SbpMsgWheeltick:
       return sbp_msg_wheeltick_decode(buf, len, n_read, &msg->wheeltick);
-    case SbpMsgAll:
-      break;
     default:
       break;
   }
@@ -1857,8 +1853,6 @@ static inline size_t sbp_message_encoded_len(sbp_msg_type_t msg_type,
       return sbp_msg_vel_ned_encoded_len(&msg->vel_ned);
     case SbpMsgWheeltick:
       return sbp_msg_wheeltick_encoded_len(&msg->wheeltick);
-    case SbpMsgAll:
-      break;
     default:
       break;
   }
@@ -2387,8 +2381,6 @@ static inline int sbp_message_cmp(sbp_msg_type_t msg_type, const sbp_msg_t *a,
       return sbp_msg_vel_ned_cmp(&a->vel_ned, &b->vel_ned);
     case SbpMsgWheeltick:
       return sbp_msg_wheeltick_cmp(&a->wheeltick, &b->wheeltick);
-    case SbpMsgAll:
-      break;
     default:
       break;
   }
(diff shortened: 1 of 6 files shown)


c/src/sbp.c
```diff
@@ -738,7 +738,7 @@ s8 sbp_payload_callback_register(sbp_state_t* s, u16 msg_type, sbp_msg_callback_
 {
   sbp_callback_t callback;
   callback.msg = cb;
-  return sbp_register_callback_generic(s, msg_type, callback,
+  return sbp_register_callback_generic(s, (sbp_msg_type_t) msg_type, callback,
```
fpezzinosn, line 741: This change and the ones in lines 751 and 769 only affect the legacy API

c/src/sbp.c
```diff
@@ -670,7 +670,7 @@ s8 sbp_all_message_callback_register(sbp_state_t *s,
                                  sbp_msg_callbacks_node_t *node) {
   sbp_callback_t callback;
   callback.decoded = cb;
-  return sbp_register_callback_generic(s, SBP_MSG_ALL, callback,
+  return sbp_register_callback_generic(s, (sbp_msg_type_t) SBP_MSG_ALL, callback,
```
fpezzinosn, line 673: The constant `SBP_MSG_ALL` defined [here](https://github.com/swift-nav/libsbp/blob/8ec25d0e9dc8b22403a1313e5c4d8ca81fd63b89/c/include/libsbp/sbp.h#L80-L81) is not part of the enum `sbp_msg_type_t` defined [here](https://github.com/swift-nav/libsbp/blob/8ec25d0e9dc8b22403a1313e5c4d8ca81fd63b89/c/include/libsbp/sbp_msg_type.h#L49-L255). Is this behavior correct?

silverjam (approved):

c/src/sbp.c
silverjam: This was intentional, as far as I know, SBP_MSG_ALL is a special/sentinel value for the API, but maybe it would be special cased to be included with the `sbp_msg_type_t` enum-- @woodfell thoughts?

fpezzinosn: And isn't the cast turning it into the first value of the enum `SbpMsgAcqResultDepA`? (assuming the internal representation of the enum matches the first value to 0)

silverjam: SbpMsgAcqResultDepA value is not zero as far as I can tell: https://github.com/swift-nav/libsbp/blob/989d77da0e50a739255e1cfe97418f6e0910d5bc/c/include/libsbp/acquisition_macros.h#L42

fpezzinosn: Oh, you are right!

woodfell: Yeah I think the better option would be to extend the generate so that `SBP_MSG_ALL` (and `SbpMsgAll`) are output rather than casting to an enum value which isn't actually valid

woodfell (approved):

silverjam (changes requested):
Need to update the templates, otherwise these changes will be lost when we add new messages: https://github.com/swift-nav/libsbp/blob/master/generator/sbpg/targets/resources/c/sbp_msg_type_template.h#L33

c/include/libsbp/v4/sbp_msg.h
```diff
@@ -831,6 +831,8 @@ static inline s8 sbp_message_encode(uint8_t *buf, uint8_t len,
       return sbp_msg_vel_ned_encode(buf, len, n_written, &msg->vel_ned);
     case SbpMsgWheeltick:
       return sbp_msg_wheeltick_encode(buf, len, n_written, &msg->wheeltick);
+    case SbpMsgAll:
```
silverjam, line 834: See https://github.com/swift-nav/libsbp/blob/master/generator/sbpg/targets/resources/c/v4/sbp_msg_template.h#L63 for where these functions are generated

silverjam (approved):

PR #1102, opened 2022-03-21
Add Message API Change Validation [BUILD-99]

isaactorz:
# Description

Adds mechanisms to automatically catch API breaking changes to the `libsbp` message spec.

To that end the following files have been added:
- `scripts/spec_validator.py` - A python script that will compare two sets of package definitions, and raise an error if a breaking change is detected.
- `.github/workflows/validation.yaml` - A workflow that compares the current branches package definitions against master
- `spec/validation/*` - A set of test files for the python script
- `scripts/validation_test_harness.bash` - Runs the python script against the test files
- `.github/workflows/test_validation.yaml` - A workflow that runs the bash test script.

The script currently detects the following changes as breaking (explicitly or implicitly):
- The type of a field is changed (test case 1)
- The ID of a message is changed (test case 2)
- The order of fields are changed (test case 3)
- The order or values of bitfields are changed (test case 8)

The following changes are ok:
- The name of a field changes (test case 4)
- New fields are appended (test case 5)
- New messages are added (test case 6)
- New packages are added (test case 7)

# API compatibility

No breaking changes, the message specs are not modified in this commit.

# JIRA Reference

https://swift-nav.atlassian.net/browse/BUILD-99

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -1,0 +1,38 @@
+name: Message Validation
+
+on:
+  pull_request:
+    paths:
+      - "spec/**"
+  push:
+    branches:
+      - 'starling-v*-release'
+      - 'v*-release'
+    tags:
+      - 'v*'
+    paths:
+      - "spec/**"
+
+jobs:
+  validation:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout Current Spec
+        uses: actions/checkout@v2
+
+      - name: Checkout Previous Spec
+        uses: actions/checkout@v2
+        with:
+          ref: master
+          path: previous
+
+      - name: Prepare Yaml Comparison
+        run: |
+          mkdir previous-spec
+          mkdir current-spec
+          cp previous/spec/yaml/swiftnav/sbp/*.yaml previous-spec/
+          cp spec/yaml/swiftnav/sbp/*.yaml current-spec/
+
+      - name: Validate Spec
+        run: |
+          bash scripts/validate.bash previous-spec current-spec
--- a/scripts/spec_validator.py
+++ b/scripts/spec_validator.py
@@ -1,0 +1,180 @@
+import argparse
+import glob
+import logging
+import os
+import sys
+import yaml
+
+
+packages = {}
+
+
+def unpack(dict):
+    return next(iter(dict.items()))
+
+
+def flatten(definitions):
+    # The yaml gets parsed into a list of dictionary objects with 1 entry each.
+    # The entry is a mapping of the definition name to another dictionary object.
+    # This flattens the list of dictionaries into a dictionary of dictionaries.
+    # This is a much more convenient data structure to perform comparisons.
+    d = {}
+    for item in definitions:
+        key, value = unpack(item)
+        d[key] = value
+    return d
+
+
+def is_message(name):
+    return "MSG" in name
+
+
+def parse_packages(root):
+    paths = glob.glob(f"{root}/*.yaml")
+    d = {}
+    for path in paths:
+        with open(path, "r") as file:
+            spec = yaml.safe_load(file)
+            package = unpack(spec)[1]
+            packages[package] = os.path.basename(path)
+            definitions = spec["definitions"]
+            d[package] = flatten(definitions)
+    return d
+
+
+def type_to_digit(type):
+    return int("".join(filter(str.isdigit, type)))
+
+
+def get_upper(bitfield):
+    return int(bitfield.split("-")[1])
+
+
+def check_bitfield_range(type, bitfields):
+    limit = type_to_digit(type)
+    for item in bitfields:
+        bitfield = unpack(item)[0]
+        val = bitfield if isinstance(bitfield, int) else get_upper(bitfield)
+        if not val < limit:
+            raise RuntimeError(
+                f"Invalid Bitfield Range!: Type {type}, Range: {bitfield}"
+            )
+
+
+def check_incoming_packages(packages):
+    logging.info("Checking Bitfield Ranges")
+    for key, value in packages.items():
+        logging.info(f"Processing Package: {key}")
+        for name, definition in value.items():
+            logging.info(f"Processing Definition: {name}")
+            if "fields" in definition:
+                fields = definition["fields"]
+                for item in fields:
+                    name, field = unpack(item)
+                    if "fields" in field:
+                        logging.info(f"Processing Field: {name}")
+                        type = field["type"]
+                        bitfields = field["fields"]
+                        check_bitfield_range(type, bitfields)
+            else:  # skip empty messages
+                continue
+
+
+def check_package_exists(previous, current):
+    if previous not in current:
+        raise RuntimeError(
+            f"Package: {previous} has been renamed or removed"
+        )
+
+
+def validate_bitfields(previous_bitfield, current_bitfield):
+    previous = flatten(previous_bitfield)
+    current = flatten(current_bitfield)
+    for key, value in previous.items():
+        if "desc" in value:  # not all messages have a desc
+            if value["desc"] == "Reserved":
+                continue
+        if key not in current:
+            raise RuntimeError(
+                "Breaking Message Mutation Detected!\n"
+                f"Bitfield {key}, has been modified"
+            )
+
+
+def validate_fields(filename, previous_fields, current_fields):
+    for previous, current in zip(previous_fields, current_fields):
+        previous_key, previous_value = unpack(previous)
+        current_key, current_value = unpack(current)
+        if previous_key != current_key:
+            logging.warning(
+                f" {filename} Name of Field Changed: {previous_key} to {current_key}"
+            )
+        if "fields" in previous_value:
+            logging.info(f"Processing Bitfield: {previous_key}")
+            previous_bitfield = previous_value["fields"]
+            current_bitfield = current_value["fields"]
+            validate_bitfields(previous_bitfield, current_bitfield)
+        previous_type = previous_value["type"]
+        current_type = current_value["type"]
+        if current_type != previous_type:
+            raise RuntimeError(
+                "Breaking Message Mutation Detected!\n"
+                f"Field (previous, current): {previous_key}, {current_key}\n"
+                f"Type (previous, current): {previous_type}, {current_type}"
+            )
+
+
+def validate_id(key, previous, current):
+    if is_message(key):
+        prev_id = previous["id"]
+        curr_id = current["id"]
+        if prev_id != curr_id:
+            raise RuntimeError(
+                "Message ID Change Detected!\n"
+                f"Previous ID: {prev_id}, Current ID: {curr_id}"
+            )
+
+
+def validate_definitions(filename, previous_definitions, current_definitions):
+    for key in previous_definitions.keys():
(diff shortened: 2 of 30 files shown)


isaactorz:
@swift-nav/devinfra Any other breaking changes we want the script to account for? 

If I get some examples then I can add them to the "test suite" and implement in the script.

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Checkout Previous Spec
+        uses: actions/checkout@v2
+        with:
+          ref: master
+          path: previous
+
+      - name: Prepare Yaml Comparison
+        run: |
+          mkdir previous-spec && \
```
notoriaga, line 33: I don't think you need to chain these with `&& \` a la docker, each line is already a separate command

notoriaga (commented):
Maybe changes to a "flags" field? Like this - https://github.com/swift-nav/libsbp/pull/1093

Not the best example because the outer `type` field changed which this should catch. But I think modifying any of the sub fields, or changing the order of them would also be a breaking change.

Would also be nice to perform some simple validation that the bitfields make sense. For example this pr - https://github.com/swift-nav/libsbp/pull/1092. If you look at the diff you can see I changed the system.yaml spec because the lsb/msb range was incorrect. I think we could do some validation that the ranges make sense/fit in the outer type. But this doesn't really fall under "change validation" and is more just a general check on the spec so maybe that can be for another time

notoriaga:
> * The name of a field changes (test case 4)

I know this isn't a breaking change, but some stuff does rely on the field names, so I think changing the field names should be done carefully. Maybe we could catch changes to names and not outright fail but emit a warning that you could surface as an "annotation"  - https://github.community/t/annotations-how-to-create-them/18387

silverjam (commented):
just some general comments so far

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,39 @@
...
+  pull_request:
+    paths:
+      - "spec/**"
+  push:
+    branches:
+      - 'starling-v*-release'
+      - 'v*-release'
+    tags:
+      - 'v*'
+      - 'starling-v*'
```
silverjam, line 13: Starling tags don't exist on this branch, so we don't need this here.

scripts/spec_validator.py
```diff
@@ -0,0 +1,123 @@
...
+    # This is a much more convenient data structure to perform comparisons.
+    d = {}
+    for item in definitions:
+        key, value = unpack(item)
+        d[key] = value
+    return d
+
+
+def is_message(name):
+    return "MSG" in name
```
silverjam, line 25: Do we have a test for message sub-type modifications?

scripts/spec_validator.py
```diff
@@ -0,0 +1,123 @@
...
+            continue
+        previous_fields = previous["fields"]
+        current_fields = current["fields"]
+        validate_fields(previous_fields, current_fields)
+
+
+def validate(previous, current):
+    previous_files = glob.glob(f"{previous}/*.yaml")
+    current_files = glob.glob(f"{current}/*.yaml")
+    current_files = filter_files(previous_files, current_files)
```
silverjam, line 101: Not sure we need to handle it in this PR, but it might be worth detecting the (degenerate) scenario where a package was renamed and asserting if that happens.

scripts/spec_validator.py
```diff
@@ -0,0 +1,99 @@
...
+def validate_fields(previous_fields, current_fields):
+    # Returns False if the new definition does not have the
+    # same type of fields in the same order
+    for previous, current in zip(previous_fields, current_fields):
+        previous_key, previous_value = unpack(previous)
+        current_key, current_value = unpack(current)
+        previous_type = previous_value["type"]
+        current_type = current_value["type"]
+        if current_type != previous_type:
+            logging.info("Breaking Message Mutation Detected!")
```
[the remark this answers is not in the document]
isaactorz: I've moved these to the exception message

scripts/spec_validator.py
```diff
@@ -0,0 +1,99 @@
...
+            logging.info(f"Field (previous, current): {previous_key}, {current_key}")
+            logging.info(f"Type (previous, current): {previous_type}, {current_type}")
+            raise RuntimeError
```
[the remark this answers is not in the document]
isaactorz: Yeah that's better.

scripts/validation_test_harness.bash
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env bash
+
+# Checks change in a fields type
+if ! python scripts/spec_validator.py spec/validation/previous/case_one spec/validation/current/case_one ; then
```
notoriaga, line 4: What do you think about changing the `case_one`,`case_two` etc folders to something that describes the test? Like `field_type_changed` for the first one

isaactorz:
> > * The name of a field changes (test case 4)
> 
> I know this isn't a breaking change, but some stuff does rely on the field names, so I think changing the field names should be done carefully. Maybe we could catch changes to names and not outright fail but emit a warning that you could surface as an "annotation" - https://github.community/t/annotations-how-to-create-them/18387

Ok I modified the script to emit a warning when it detects a name change, and surface it via `validate.bash` but I'm not getting the annotation to show up in the UI.

silverjam (changes requested):
I think we’re close, couple things:
- merge master and update the PR branch, should get a clean sets of checks
- I like @notoriaga’s suggestion to name the test case folders descriptively, can we do that unless there’s a good reason not to?

silverjam:
> > Maybe changes to a "flags" field? Like this - #1093
> > Not the best example because the outer `type` field changed which this should catch. But I think modifying any of the sub fields, or changing the order of them would also be a breaking change.
> > Would also be nice to perform some simple validation that the bitfields make sense. For example this pr - #1092. If you look at the diff you can see I changed the system.yaml spec because the lsb/msb range was incorrect. I think we could do some validation that the ranges make sense/fit in the outer type. But this doesn't really fall under "change validation" and is more just a general check on the spec so maybe that can be for another time
> 
> See tests 9 and 10 for the updated handling of bitfields.
> 
> As for validating that the ranges make sense, I'd think that would be something that would be better handled by unit tests than to rely on this script to check.

I think this would be a fairly trivial addition, basically just need to check that the ranges aren’t larger than the type and that they are non overlapping.  How would a unit test catch this error?

isaactorz:
> > > Maybe changes to a "flags" field? Like this - #1093
> > > Not the best example because the outer `type` field changed which this should catch. But I think modifying any of the sub fields, or changing the order of them would also be a breaking change.
> > > Would also be nice to perform some simple validation that the bitfields make sense. For example this pr - #1092. If you look at the diff you can see I changed the system.yaml spec because the lsb/msb range was incorrect. I think we could do some validation that the ranges make sense/fit in the outer type. But this doesn't really fall under "change validation" and is more just a general check on the spec so maybe that can be for another time
> > 
> > 
> > See tests 9 and 10 for the updated handling of bitfields.
> > As for validating that the ranges make sense, I'd think that would be something that would be better handled by unit tests than to rely on this script to check.
> 
> I think this would be a fairly trivial addition, basically just need to check that the ranges aren’t larger than the type and that they are non overlapping. How would a unit test catch this error?

Maybe I'm missing something here but from the example Steve showed, the range in the bitfields was incorrect (`8-16`). It looks like these ranges control the bitmask macros; in this case `SBP_STATUS_REPORT_MAJOR_PROTOCOL_NUMBER_MASK`.

I don't think `0x1ff` is valid hex. So when we invoke the setter, getter macros, we probably would not get the values we expected. Looking at the test suite, there aren't tests for this particular message; `MSG_STATUS_REPORT`.

With that being said I agree that it wouldn't be terribly complicated to implement this check so I can do that if we want to.

isaactorz:
> I think we’re close, couple things:
> 
> * merge master and update the PR branch, should get a clean sets of checks
> * I like @notoriaga’s suggestion to name the test case folders descriptively, can we do that unless there’s a good reason not to?

Yeah no reason not too I'll update the names. Just wanted to get eyes on all the other changes.

notoriaga:
> I don't think `0x1ff` is valid hex. So when we invoke the setter, getter macros, we probably would not get the values we expected. Looking at the test suite, there aren't tests for this particular message; `MSG_STATUS_REPORT`.

I don't think the auto generated test suite covers the bitfield methods at all, we could probably add that to the test yaml files and update the c/rust test codegen if we wanted to cover that

isaactorz:
@silverjam @notoriaga 

Good news the bitfield range check works. Bad news it looks like at least one of our messages so far has an invalid range; `MSG_STARTUP`:`cause`.

Do we want to try to patch these inline or should we open up a separate ticket?

If we want the latter I can modify the script so that it only checks newly added fields .

Regardless I'll find out exactly how many of these degenerate ranges are in our current definitions.

silverjam:
> Good news the bitfield range check works. Bad news it looks like at least one of our messages so far has an invalid range; `MSG_STARTUP`:`cause`.
> 
> Do we want to try to patch these in this PR or should we open up a separate ticket?

Let's fix in another PR and then update this PR once the fix is merged.

silverjam:
> > Good news the bitfield range check works. Bad news it looks like at least one of our messages so far has an invalid range; `MSG_STARTUP`:`cause`.
> > Do we want to try to patch these in this PR or should we open up a separate ticket?
> 
> Let's fix in another PR and then update this PR once the fix is merged.

This message is emitted here: https://cs.github.com/swift-nav/piksi_firmware_private/blob/d94cbfb44e354110d716babea59c0fbe8b780f55/src/main/main.c?q=MSG_STARTUP++language%3Ac#L182

But the flags are all set to zero, so the risk around fixing the bitfield ranges should be minimal (or near zero).

isaactorz:
> https://cs.github.com/swift-nav/piksi_firmware_private/blob/d94cbfb44e354110d716babea59c0fbe8b780f55/src/main/main.c?q=MSG_STARTUP++language%3Ac#L182

The `startup-type` field in this message also needs to be fixed.

isaactorz:
@silverjam @notoriaga FYI still working on surfacing the file that has a field name change so that the warning annotation can be associated with it.

silverjam:
Looking better, need to figure out what this is failing: https://github.com/swift-nav/libsbp/runs/5825726116?check_suite_focus=true -- I'm running `make all` locally to see if I can see what's up.

silverjam (approved):
Looking better, need to figure out why this is failing: https://github.com/swift-nav/libsbp/runs/5825726116?check_suite_focus=true -- I'm running `make all` locally (within the docker container) to see if I can replicate the issue.

silverjam:
> Looking better, need to figure out why this is failing: https://github.com/swift-nav/libsbp/runs/5825726116?check_suite_focus=true -- I'm running `make all` locally (within the docker container) to see if I can replicate the issue.

Never mind on this, I guess the failure is from a test change for this

> @silverjam @notoriaga FYI still working on surfacing the file that has a field name change so that the warning annotation can be associated with it.

isaactorz:
Anything else on this @notoriaga ? I've implemented the warnings annotation for name changes, and it gets associated with the yaml file contains the change.

If not then I think it's ready to be merged.

notoriaga (approved):
Looks great!

PR #1106, opened 2022-04-04
Fix Bitfield Ranges in MSG_STARTUP [DEVINFRA-675]

isaactorz:
# Description

During the development of our message validation script [PR 1102](https://github.com/swift-nav/libsbp/pull/1102), we discovered that the bitfield range for two fields in the `system.MSG_STARTUP` message were incorrect:
- `cause`: type u8, range 0-8
- `startup_type`: type u8, range 0-8

This PR fixes these fields to the correct range

@swift-nav/devinfra

# API compatibility

Technically this is an API break, but it fixes an error in the message specification.

## API compatibility plan

@silverjam has identified at least one place the message is emitted: 

https://github.com/swift-nav/piksi_firmware_private/blob/d94cbfb44e354110d716babea59c0fbe8b780f55/src/main/main.c?q=MSG_STARTUP++language%3Ac#L182

The message is emitted with all 0 value, so patching should be minimal risk.

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-675


isaactorz:
FYI @swift-nav/devinfra I was able to sort out the issue and run `make all` from within the container. Looks like it ran to completion (last output was `Finished running Elm JSON Schema tests`).

Not super familiar with what all is supposed to get updated. Do we only have c and rust bindings for this message?

notoriaga (approved):

silverjam (approved):

woodfell (approved):
LGTM!

The PR title references `MSG_STATUS` but it should be `MSG_STARTUP`?

isaactorz:
> LGTM!
> 
> The PR title references `MSG_STATUS` but it should be `MSG_STARTUP`?

Nice catch, yes it should be `MSG_STARTUP`

PR #1107, opened 2022-04-06
Suppressed warnings and deprecation notices for Java

adrian-kong:
# Description

@swift-nav/devinfra

Adds @SuppressedWarning for unchecked generics and removes SBPStruct#parse return from generic to SBPStruct.

Changes the concurrent for loop removing to maybe prevent ConcurrentModificationException (?) - not sure if needed, also it doesn't look like it was being called anyways

# API compatibility

It should not, it only suppresses warnings and narrowing the scope of return type

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-479

### DIFF
--- a/java/src/com/swiftnav/sbp/client/SBPHandler.java
+++ b/java/src/com/swiftnav/sbp/client/SBPHandler.java
@@ -197,7 +197,7 @@ public class SBPHandler implements Iterable<SBPMessage> {
                     if (msg != null) {
                         return msg;
                     }
-                } catch (InterruptedException ignored) { //NOSONAR
+                } catch (InterruptedException ignored) { // NOSONAR
                 }
             }
             // If we get here finished is set to true so there are no more messages available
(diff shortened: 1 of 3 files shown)


java/src/com/swiftnav/sbp/client/SBPHandler.java
```diff
@@ -201,9 +197,7 @@ protected SBPMessage getNext() {
                     if (msg != null) {
                         return msg;
                     }
-                    continue;
-                } catch (InterruptedException e) {
-                    continue;
+                } catch (InterruptedException ignored) {
```
silverjam, line 200: I think SonarCloud flagged this because you touched this line, if it makes sense we can just fix according to SonarCloud recommendation, if it doesn't make sense we can ignore the warning.

image

adrian-kong: hmm yea, maybe we could log the error as it suggests but is also another thing which the library doesn't look like it uses a logger instead using System.out.println() everywhere

I don't think it should have to throw an InterruptedException and it follows logically like previously:
https://github.com/swift-nav/libsbp/runs/2882927558

notoriaga: I think we should ignore it (with `// NOSONAR` I think) and make a ticket to track switching to a logging facade

silverjam (approved):
Ship it! 🎉

PR #1108, opened 2022-04-06
Add support for SSR_ORBIT_CLOCK_BOUNDS message [OTA-65]

fpezzinosn:
# Description

@swift-nav/devinfra

Adds the SSR_ORBIT_CLOCK_BOUNDS message to the SSR package as described [here](https://swift-nav.atlassian.net/wiki/spaces/ENG/pages/2178286328/WIP+OTA+Interface+Addition+design+doc+-+Integrity+messages).

It also cherrypicks a fix for rust generation https://github.com/swift-nav/libsbp/pull/1109 related to missing `short_desc` on the spec.

# API compatibility

It doesn't. I'm only adding one new message following the procedure.

## API compatibility plan

# JIRA Reference

https://swift-nav.atlassian.net/browse/OTA-65


python/tests/sbp/test_table.py
```diff
@@ -40,7 +40,7 @@ def test_table_count():
   Test number of available messages to deserialize.
 
   """
-  number_of_messages = 204
+  number_of_messages = 205
```
fpezzinosn, line 43: I've updated the number of expected messages by the amount I've added in this PR, but the test still expects 204. Am I missing something since this is a new package?

spec/yaml/swiftnav/sbp/integrity.yaml
```diff
@@ -0,0 +1,94 @@
...
+package: swiftnav.sbp.integrity
+description: Over-the-air Integrity Messages
+stable: True
+public: False
+include:
+  - types.yaml
+  - gnss.yaml
+definitions:
+
+ - OrbitClockBound:
```
fpezzinosn, line 20: In the excel file Orbit Bounds and Clock Bounds are in separate groups. But I think it make more sense to have them combined in one structure, since the satellite ID is only present in the first group.

python/tests/sbp/test_table.py
isaactorz: Think you need to run the `make all` target to generate the various language bindings for new messages.

fpezzinosn: Yes I did. I didn't commit yet the autogenerted code (to not pollute the PR in case there is something wrong in my yaml)

isaactorz: There's a lot of build dependencies so it's probably best to build in the docker image.

fpezzinosn:
I had problems with the following targets:
# C
  - The same thing GH Actions reports, this is a new package and it cannot create the integrity folder in the include directory. In my machine I solved this by manually creating it.

# Python
 - The number of expected messages is still the previous one, apparently it doesn't recognize that I added a new package with one message

# Haskel
 - No issues

# Java
 - No issues

# Rust
I got the following error and I'm not sure what's it about.

```
py run-test: commands[0] | python -m sbpg.generator -i /mnt/workspace/spec/yaml/swiftnav/sbp/ -o /mnt/workspace/rust/ -r 4.1.5-6-gafc2e4ad --rust
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/mnt/workspace/generator/sbpg/generator.py", line 216, in <module>
    main()
  File "/mnt/workspace/generator/sbpg/generator.py", line 202, in main
    rs.render_all(output_dir, parsed, release)
  File "/mnt/workspace/generator/sbpg/targets/rust.py", line 404, in render_all
    render_file(
  File "/mnt/workspace/generator/sbpg/targets/rust.py", line 431, in render_file
    f.write(py_template.render(**args))
  File "/mnt/workspace/.tox/py/lib/python3.8/site-packages/jinja2/environment.py", line 1291, in render
    self.environment.handle_exception()
  File "/mnt/workspace/.tox/py/lib/python3.8/site-packages/jinja2/environment.py", line 926, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "/mnt/workspace/generator/sbpg/targets/resources/rust/sbp_messages_mod.rs", line 114, in top-level template code
    /// (((m.short_desc | commentify(indent=2) )))
  File "/mnt/workspace/generator/sbpg/targets/rust.py", line 147, in commentify
    value = wrap_urls(value)
  File "/mnt/workspace/generator/sbpg/targets/rust.py", line 119, in wrap_urls
    urls = re.findall(r"\(?https?://[^\s\)]+\)?", s)
  File "/usr/lib/python3.8/re.py", line 241, in findall
    return _compile(pattern, flags).findall(string)
TypeError: expected string or bytes-like object
ERROR: InvocationError for command /mnt/workspace/.tox/py/bin/python -m sbpg.generator -i spec/yaml/swiftnav/sbp -o rust -r 4.1.5-6-gafc2e4ad --rust (exited with code 1)
```

# Protobuf
 - No issues

# JSON Schema
 - No issues

python/tests/sbp/test_table.py
fpezzinosn: Correct. I've been using the image indicated in the README. This one `swiftnav/libsbp-build:2021-10-13`.

spec/yaml/swiftnav/sbp/integrity.yaml
```diff
@@ -0,0 +1,94 @@
...
+          type: u8
+          units: m
+          desc: Clock Bound Mean
+      - clock_std:
+          type: u8
+          units: m
+          desc: Clock Bound Standard Deviation
+
+ - MSG_SSR_ORBIT_CLOCK_BOUNDS:
+    id: 1234
```
fpezzinosn, line 62: This id is a placeholder, It must be updated to the final value before merging the PR.

spec/yaml/swiftnav/sbp/integrity.yaml
```diff
@@ -0,0 +1,94 @@
...
+          desc: IOD of the SSR bound. Encoded following RTCM DF413 specification.
+      - ssr_id:
+          type: u8
+          desc: SSR Solution ID. Encoded following RTCM DF415 specification.
+      - constellation:
+          type: u8
+          desc: Constellation ID to which the SVs belong.
+      - num_sat:
+          type: u8
+          desc: Number of satellites. Encoded following RTCM DF387 specification.
```
IsakTjernberg, line 90: @swift-nav/devinfra I notice in most other SBP messages with arrays the number of array entries are not present as a field. I suppose it is just deduced from the message length? Should we follow the normal practice and skip this one then?

IsakTjernberg (commented):
Great start! I have some comments/thoughts on the message definition. I also wonder if we shouldn't put this in the existing SSR section?
They are very much tied to the rest of the SSR format and in the Excel sheet are named such as:
`MSG_SSR_ORBIT_CLOCK_BOUNDS`.

fpezzinosn:
I have moved the message into the existing SSR package.

notoriaga:
@fpezzinosn this should fix the rust generation problem  - https://github.com/swift-nav/libsbp/pull/1109

fpezzinosn:
> @fpezzinosn this should fix the rust generation problem - #1109 (could also add a `short_desc` field to your message. Or maybe you could make what you have as `desc` the `short_desc` and write something longer for the `desc`). Either way feel free to merge those changes into this pr

Yes! That was the issue
Thank you!

fpezzinosn:
The only thing that fails is the code coverage of java. But the other messages in the SSR package aren't covered either.

spec/yaml/swiftnav/sbp/integrity.yaml
notoriaga: The spec allows for both fixed length and variable length arrays, if you don't specify the length then it's variable length. I think if you have a variable length array it has to be at the end of the message (@silverjam can confirm)

notoriaga: With the exception of some older messages like `MSG_FILEIO_WRITE_REQ` that end with a variable length string and an array, which works because the string is null terminated

silverjam: We've generally tried to avoid variable length arrays, if we can avoid them here that'd be ideal.

silverjam: > The spec allows for both fixed length and variable length arrays, if you don't specify the length then it's variable length. I think if you have a variable length array it has to be at the end of the message (@silverjam can confirm)

It does need to be at the end, but again, let's avoid them if we can, they are not well supported by non-LLVM/GNU compilers.

silverjam:
> The only thing that fails is the code coverage of java. But the other messages in the SSR package aren't covered either.

Java is the only place where code coverage is tracked, so any new messages need to be covered (Java acts as a proxy for other languages), but this can be delayed until this PR is ready to merge.

spec/yaml/swiftnav/sbp/integrity.yaml
notoriaga: Seems like another check we could add to the tool @isaactorz just did

silverjam: @notoriaga yeah, good idea, could you add a Jira please?

fpezzinosn:
Added a test spec, unfortunately it fails.
Probably missed one bit when encoding the example, have to check again.

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -802,3 +802,113 @@ definitions:
...
+           type: u8
+           desc: Number of messages in the dataset
+       - id_msg_dataset:
+           type: u8
+           desc: Position of this message in the dataset
+       - ssr_update_interval:
+           type: u8
+           desc: >
+             Update interval between consecutive bounds. Encoded following
+             RTCM DF391 specification.
```
IsakTjernberg, line 897: I think these can be placed in a header type (see for example how it works for the STECHeader here: https://github.com/swift-nav/libsbp/blob/2c9688ad3ae3effc4714cc7e50c2ea994c9546e8/spec/yaml/swiftnav/sbp/ssr.yaml#L375).
The variables here will be similar in the other upcoming messages (SSR_CODE_PHASE_BIASES_BOUNDS, new SSR_STEC_CORRECTION, SSR_GRIDDED_CORRECTION_BOUNDS).
In the header, let's name `nb_msg_dataset` and `id_msg_dataset` as `num_msgs` and `seq_num` instead, then we follow the existing practice already in libsbp.
These are the fields I'd suggest for the header type:
time (GPSTimeSec), num_msgs, seq_num, update_interval, solution_id.
Note that iod is not in the above list because it's naming changes slightly for the gridded and stec corrections.

IsakTjernberg (commented):
This looks great, just some small renaming suggested here from me. I'll let Devinfra have the final say but I'm good with merging this after my comment is adressed.
Let's take the PR out of draft status?

spec/yaml/swiftnav/sbp/ssr.yaml
fpezzinosn: Does the order matter? For example in the definition of MSG_SSR_ORBIT_CLOCK_BOUNDS, IOD comes before solution ID but in MSG_SSR_STEC_CORRECTION it comes after. So the header type isn't exactly the same for both.

IsakTjernberg: Good call, we should update the Excel sheet so the solution ID comes before in both. And then solution ID can be in the common header here.

spec/yaml/swiftnav/sbp/integrity.yaml
fpezzinosn: I went with ID 1502 since the bounds provided in this message are the ones for MSG_SSR_ORBIT_CLOCK (1501)

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -802,3 +802,120 @@ definitions:
...
+           desc: Header of a bounds message.
+       - ssr_iod:
+           type: u8
+           desc: IOD of the SSR bound. Encoded following RTCM DF413 specification.
+       - const_id:
+           type: u8
+           desc: Constellation ID to which the SVs belong.
+       - nb_sat:
+           type: u8
+           desc: Number of satellites. Encoded following RTCM DF387 specification.
```
silverjam, line 916: > Encoded following RTCM DF387 specification.

Is this true for this particular field?

fpezzinosn: Maybe you are right. In the Excel file we are saying that this field is similar to the existing RTCM Data Field. Perhaps `Encoded` isn't the best choice of word. (And also DF387 is only 6 bits long)

silverjam: Maybe "Number of satellites, follows the RTCM DF387 specification." -- if it's only expected to be 6-bit, we should also document the valid ranges (0-63).

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -802,3 +802,120 @@ definitions:
...
+     id: 0x05DE
+     short_desc: Combined Orbit and Clock Bound
+     public: False
+     fields:
+       - header:
+           type: BoundsHeader
+           desc: Header of a bounds message.
+       - ssr_iod:
+           type: u8
+           desc: IOD of the SSR bound. Encoded following RTCM DF413 specification.
```
silverjam, line 910: Are all of these values expected to use all 8 bits?  Would it make sense to pack these values into a larger type as a collection of bitfields?  This would allow use to more closely model the RTCM fields.

silverjam: Something similar to this: https://github.com/swift-nav/libsbp/blob/master/spec/yaml/swiftnav/sbp/system.yaml#L227 but instead of `flags` something like `metadata`

silverjam: Or probably more similar to this: https://github.com/swift-nav/libsbp/blob/master/spec/yaml/swiftnav/sbp/system.yaml#L200 -- since we won't be able to specify the values field in the yaml spec

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -802,3 +802,120 @@ definitions:
...
               essentially the index of the quadrants that contain transitions between
               valid and invalid (and vice versa) are encoded as u8 integers.
+
+ - BoundsHeader:
+     short_desc: Header for the Bounds messages
+     fields:
+       - time:
+           type: GPSTimeSec
+           desc: GNSS reference time of the bound
+       - num_msgs:
```
woodfell, line 812: For consistency, in SBP the prefix `n_` is typically used for these sorts of fields. Such as `n_obs` (observation_header) or `n_group_msgs` (system_meta)

```suggestion
       - n_msgs:
```

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -802,3 +802,120 @@ definitions:
...
+       - header:
+           type: BoundsHeader
+           desc: Header of a bounds message.
+       - ssr_iod:
+           type: u8
+           desc: IOD of the SSR bound. Encoded following RTCM DF413 specification.
+       - const_id:
+           type: u8
+           desc: Constellation ID to which the SVs belong.
+       - nb_sat:
```
woodfell, line 914:
```suggestion
       - n_sats:
```

Same consistency comment here. Made it plural to match `n_sats` in (eg) `pos_ecef`

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -802,3 +802,120 @@ definitions:
...
+           type: u8
+           desc: Number of satellites. Encoded following RTCM DF387 specification.
+       - orbit_clock_bounds:
+           type: array
+           fill: OrbitClockBound
+           map_by: sat_id
+           desc: Orbit and Clock Bounds per Satellite
```
woodfell, line 921: Is the number of elements in this array described by the `n_sats` / `nb_sat` field above? If so then this field needs to have an extra key to relate the two fields to the encoder/decoder can properly handle the variable length array
```suggestion
       - orbit_clock_bounds:
           type: array
           fill: OrbitClockBound
           size_fn: n_sats
           map_by: sat_id
           desc: Orbit and Clock Bounds per Satellite
```

spec/yaml/swiftnav/sbp/integrity.yaml
ljbade: I think the message IDs don't need to be as tightly grouped these days. Considering it is tricky figuring out the next free ID.

spec/yaml/swiftnav/sbp/ssr.yaml
IsakTjernberg: Yes probably we should drop the reference to RTCM here. After all it's just a number?
And I think we won't limit the range to only 6-bit.

spec/yaml/swiftnav/sbp/ssr.yaml
IsakTjernberg: @silverjam do you know how this works for other already existing SBP messages which mimic RTCM fields?
For example SSR_CODE_BIASES here:
https://github.com/swift-nav/libsbp/blob/8a6c6216c105429d86c82976a5a88d5cabf5f190/spec/yaml/swiftnav/sbp/ssr.yaml#L293-L297
The `update_interval` is specified as a `u8`, but it also says it's encoded as DF391 which is only 4 bits.
I would think we still fill out all 8 bits, but the interpretation of it's value is then according to DF391?

spec/yaml/swiftnav/sbp/ssr.yaml
IsakTjernberg: In the other SSR messages (STEC Correction, Gridded Correction) the field is called `num_msgs` though.

spec/yaml/swiftnav/sbp/integrity.yaml
silverjam: > Variable length arrays are well supported in the v4 API but were the root cause of many compiler problems in the past. 

@woodfell oops, thanks for the reminder

spec/yaml/swiftnav/sbp/ssr.yaml
silverjam: I think if we design these messages to more closely match their RTCM equivalents, we'll have an easier time when we go through the conversion effort.  Data fidelity and correctness is more important than user experience.  So... the question is, do we want do this now... or later?

Risks:
- On doing it now: delaying progress on getting the end-to-end solution working with SBP
- On doing it later: it'll get forgotten or we'll be in a situation where we'll lose fidelity or have to drop data when converting to RTCM

In general this also relates to this issue here: https://github.com/swift-nav/libsbp/pull/1108#discussion_r850505904

IsakTjernberg: These messages don't have RTCM equivalents. Any conversion to RTCM will be just wrapping them in the Swift proprietary RTCM message. The fields which are being referenced from RTCM though exist in other RTCM SSR message types which we will need to compare the values against, so there it makes sense to use a similar definition.

silverjam: Alright, @woodfell's suggestion makes sense in that case

fpezzinosn: Removed all the references to RTCM where the field is just a number. Left the ones regarding satellites IDs and update interval since they have some special meaning. The comment now reads `Similar to RTCM DF` for context.

fpezzinosn:
There seems to be an issue with haskell org https://status.haskell.org/
The job is failing because it can't download the dependencies

spec/yaml/swiftnav/sbp/integrity.yaml
fpezzinosn: Does `make mapping` show all the currently used IDs or only the public ones?
Because I was basing it on that info.

IsakTjernberg (approved):

notoriaga (approved):

silverjam (changes requested):
> @swift-nav/devinfra any objections to merging this?

Merging to master canonicalizes and "releases" the message.  This should mean that the message is finalized and unlikely to change.  If the message needs to change then we need to deal with retiring the old message for any changes that we need to make.

I think we need a strategy for handing these messages that will work with our existing testing and quality workflow while allowing us to iterate on these messages without incurring the overhead of having to rev and retire messages.

woodfell (approved):

fpezzinosn:
I've created a new `staging` branch as discussed and the PR now points to that branch.

silverjam (approved):

PR #1111, opened 2022-04-08
Add a rust sbp iterator variant with timeout[CPP-721]

john-michaelburke:
# Description

@swift-nav/devinfra

Adds an iter_messages variant with a timeout which triggers when no preamble has been found in the read buffer for some duration. An example of when this behavior is frequently observed is when connecting to a serial device with an unexpected baudrate.

# API compatibility

This introduces some duplication which would need to be maintained if iter_messages or the SBPDecoder changes in the future but likely minor upkeep. I have added two unittests to hopefully catch any breaking changes to this feature.

# JIRA Reference

https://swift-nav.atlassian.net/browse/CPP-721

### DIFF
--- a/rust/sbp/src/de.rs
+++ b/rust/sbp/src/de.rs
@@ -216,16 +216,6 @@ impl Decoder for SbpDecoder {
     }
 }
 
-#[derive(Debug)]
-pub struct TimeoutError;
-
-impl std::error::Error for TimeoutError {}
-impl std::fmt::Display for TimeoutError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "timeout waiting for valid message",)
-    }
-}
-
 struct TimeoutSbpDecoder {
     decoder: SbpDecoder,
     timeout_duration: Duration,
@@ -270,7 +260,10 @@ impl Decoder for TimeoutSbpDecoder {
 
 #[cfg(test)]
 mod tests {
-    use std::{convert::TryInto, io::Cursor};
+    use std::{
+        convert::TryInto,
+        io::{Cursor, Write},
+    };
 
     use bytes::BufMut;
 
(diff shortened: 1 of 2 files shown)


notoriaga:
Out of curiosity have you gotten a chance to test this out in the console/does it fix the issue you were seeing?

john-michaelburke:
> Out of curiosity have you gotten a chance to test this out in the console/does it fix the issue you were seeing?

Yeah it is working well. The selection of the timeout may need some tuning but the app does not lock up which is a huge win. 

I was thinking having some form of a wrapper as you suggested would be potentially cleaner approach but I'm not sure how you would pass the SbpDecoder through FramedRead. I think I would need to hold a reference to an SbpDecoder inside my new struct. Maybe thats not too bad?

notoriaga:
> > Out of curiosity have you gotten a chance to test this out in the console/does it fix the issue you were seeing?
> 
> Yeah it is working well. The selection of the timeout may need some tuning but the app does not lock up which is a huge win.

Nice!

> I was thinking having some form of a wrapper as you suggested would be potentially cleaner approach but I'm not sure how you would pass the SbpDecoder through FramedRead. I think I would need to hold a reference to an SbpDecoder inside my new struct. Maybe thats not too bad?

I think I have something that might work, I'll push a different branch in a few and you can take a look

rust/sbp/src/de.rs
```diff
@@ -196,6 +216,58 @@ impl Decoder for SbpDecoder {
     }
 }
 
+#[derive(Debug)]
+pub struct TimeoutError;
```
notoriaga, line 220: This can be removed right?

rust/sbp/src/de.rs
```diff
@@ -206,6 +278,59 @@ mod tests {
...
+            for b in buf.iter_mut() {
+                self.idx += 1;
+                if self.idx % PREAMBLE as usize == 0 {
+                    self.idx += 1;
+                }
+                *b = (self.idx % u8::MAX as usize) as u8;
+            }
+            Ok(buf.len())
+        }
+    }
```
notoriaga, line 318: could this just be 

```rust
    impl std::io::Read for NeverPreambleReader {
        fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
            buf.write_all(&[0]);
            Ok(1)
        }
    }
```

rust/sbp/src/de.rs
```diff
@@ -206,6 +278,59 @@ mod tests {
 
     use super::*;
 
+    struct ZeroReader;
+
+    impl std::io::Read for ZeroReader {
+        fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+            Ok(buf.len())
+        }
+    }
```
notoriaga, line 287: Should probably be 

```rust
    impl std::io::Read for ZeroReader {
        fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
            Ok(0)
        }
    }
```

if buf had some bytes in it before for some reason, this read impl would say it read some bytes even though it didn't

rust/sbp/src/de.rs
john-michaelburke: 😬  Good catch removed!

rust/sbp/src/de.rs
john-michaelburke: Done.

rust/sbp/src/de.rs
john-michaelburke: Your suggested reader actually gets handled currently. What I had originally was simply just returning the buffer assuming it was filled with zeros. This was just abusing the same corner case in the other testcase, so per your suggestion I have merged the two. I decided to keep the "NeverPreambleReader" as I think it describes the corner case better than ZeroReader unless you have a better suggestion?

notoriaga (approved):
:shipit:

rust/sbp/src/de.rs
notoriaga: Ahhh yeah that makes sense, sounds good to me

PR #1113, opened 2022-04-12
Updating libraries and license to 2022

adrian-kong:
# Description

@swift-nav/devinfra

Small update to the libraries and the license header to 2022

# API compatibility

It should still be compatible, no changes in API.


joelynch (commented):
Looks good to me.  Is there a JIRA task for this one?

adrian-kong:
> Looks good to me. Is there a JIRA task for this one?

nope just a quick and tiny update

joelynch (approved):
Looks good to me

samvrlewis (approved):

silverjam (changes requested):
Since most files in this repository are generated you'll need to update the template file that generates everything in order for this change to persist please see here for the correct location to update the generated files:

https://github.com/swift-nav/libsbp/blob/master/generator/sbpg/targets/resources/sbp_java.java.j2#L2

After updating the template you can run the generator in order to check that it generates the correct output, in your checkout of the libsbp repo, run the following:

```
docker run  -v $PWD:/mnt/workspace -i -t swiftnav/libsbp-build:2021-10-13 /bin/bash -c "make java"
```

adrian-kong:
oops i think i broke the branch

PR #1119, opened 2022-04-20
Publishing to maven central

adrian-kong:
# Description

@swift-nav/devinfra

Releases java with

```
gradle publish
```

requires the account which should be in lastpass as well as a gpg key 

and usage would be:

```
<dependency>
  <groupId>com.swiftnav</groupId>
  <artifactId>sbp</artifactId>
  <version>1.0.0</version>
</dependency>
```

I think it should be searchable at https://search.maven.org/artifact/com.swiftnav/sbp when the site gets indexed

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-294

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,34 +1,34 @@
-name: Build docker image
-on:
-  workflow_dispatch:
-  pull_request:
-    paths:
-      - Dockerfile
-      - Makefile
-      - .github/workflows/docker.yaml
-jobs:
-  build:
-    strategy:
-      matrix:
-        ubuntu-release:
-          - bionic
-          - focal
-#          - jammy
-    name: Build docker image
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
-          submodules: recursive
-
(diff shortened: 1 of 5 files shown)


adrian-kong:
removed installing gradle from workflow because it was downloading 4.5.1 which didn't support java method for the generator test. not sure if it broke anything

silverjam (commented):
A few requests (from our meeting):
- Let's update docs with what's needed to release (what accounts, gpg keys, etc)
- The process for releasing, even if it's partially manual
- Make released version number match the version number in the repository

silverjam:
@adrian-kong let's move the Java publishing instructions to the HOWTO.md in this repo

notoriaga:
@silverjam we are getting errors from apt-get when building the docker image with ubuntu groovy. groovy is end of life, can we just remove it? or should we look into it more

silverjam:
> @silverjam we are getting errors from apt-get when building the docker image with ubuntu groovy. groovy is end of life, can we just remove it? or should we look into it more

Yeah, feel free to remove it, let's add a Jira introduce Jammy (22.04) at some point too.

silverjam (approved):

PR #1122, opened 2022-04-26
Runnable sbp java example

adrian-kong:
# Description

Added how to use and runnable build for examples in a different module instead of the same gradle project.

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-692


java/build.gradle
```diff
@@ -20,7 +20,6 @@ dependencies {
 sourceSets {
     test {
         java {
-            srcDir 'example'
```
john-michaelburke, line 23: Is there a separate way that we could add the examples into this build file as optional? libsbp's current docker doesn't include mvn. Or maybe there would be value in including a separate gradle file just for building the examples?

john-michaelburke, line 23: Is there a separate way that we could add the examples into this build file as optional? libsbp's current docker doesn't include mvn. Or maybe there would be value in including a separate gradle file just for building the examples?

adrian-kong: hmmmmm could swap it to use gradle instead of maven but wanted to just write a sample outline project of how to implement it, and could also be runnable by using `mvn install` if needed.

but switching entirely to gradle would be fine as well, was thinking like an example project not within the current `build.gradle`, doesn't matter on whether its maven or gradle - (not necessary in building it with docker, just for demo for how to implement it + a runnable demo if needed)

john-michaelburke: I think a separate Gradle sounds reasonable and could add a step to the Makefile to make java-examples or something as well.

adrian-kong:
some issues with docker.yaml file similar to the maven central PR

adrian-kong:
split out the docker.yaml thing to a different issue, all checks pass now 👍

john-michaelburke:
Looks like you may want to rebase your branch. That being said I think you could have commented out this line to prevent the crash on groovy but keep the workflow for the other distros:
https://github.com/swift-nav/libsbp/blob/031635acb3ab0f55bb4908b482ebc719129188b0/.github/workflows/docker.yaml#L16

adrian-kong:
> Looks like you may want to rebase your branch. That being said I think you could have commented out this line to prevent the crash on groovy but keep the workflow for the other distros:
> 
> https://github.com/swift-nav/libsbp/blob/031635acb3ab0f55bb4908b482ebc719129188b0/.github/workflows/docker.yaml#L16



> Looks like you may want to rebase your branch. That being said I think you could have commented out this line to prevent the crash on groovy but keep the workflow for the other distros:
> 
> https://github.com/swift-nav/libsbp/blob/031635acb3ab0f55bb4908b482ebc719129188b0/.github/workflows/docker.yaml#L16

i tried that in a diff PR 

https://github.com/swift-nav/libsbp/runs/6204715987

it had some error that wasn't relevant to the changes so wasn't sure what was happening for it

john-michaelburke (approved):

PR #1125, opened 2022-04-27
Generate missing tests via JSON

adrian-kong:
# Description

Draft PR on top of missing.py to generate tests for json inputs

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-676


generator/json2test.py
```diff
@@ -0,0 +1,144 @@
...
+            is_real_message = hasattr(cls, "_parser")
+            if is_real_message:
+                if cls_name.endswith("Dep") or cls_name[-4:-1] == "Dep":
+                    deprecated_missing += 1
+                if cls_name.endswith("Req") or cls_name.endswith("Resp"):
+                    req_resp_missing += 1
+                print("Missing {}: {}".format(msg_type, cls_name))
+                classes_without_tests[cls_name] = cls
+
+    print(f"Total missing: {len(classes_without_tests)} ({req_resp_missing} Req/Resp; {deprecated_missing} Dep.)")
```
silverjam, line 110: We don't need any of the code here that looks for missing test cases, so can probably just delete lines 83-110

generator/json2test.py
```diff
@@ -0,0 +1,144 @@
...
+    while True:
+        try:
+            print("=====")
+            inp = input("input: ")
```
silverjam, line 114: This should come from the command line, use sys.argv to fetch input arguments (from jira): `python ./testgen.py <MSG_ID> <JSON_FILE> <YAML_OUTPUT>`

generator/json2test.py
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+
+def write_test(msg, test_set_name="swiftnav"):
+    module = msg.__class__.__module__
+    name = msg.__class__.__name__
+    path_parts = (
+        [ROOTPATH, test_set_name] + module.split(".") + ["test_{}.yaml".format(name)]
+    )
+    filename = os.path.join(*path_parts)
```
silverjam, line 66: Pass in the filename as a parameter to the function

generator/json2test.py
```diff
@@ -0,0 +1,144 @@
...
+                print("invalid format, no msg_type found")
+                continue
+            clazz = lookup(msg_type)
+            print("found class: " + str(clazz.__name__))
+            msg_obj = clazz(**msg)
+
+            # sets crc, length, payload
+            msg_obj.to_binary()
+            if clazz not in classes_without_tests.values():
+                print("test is not missing")
```
silverjam, line 127: Can delete any code related to "missing message" detection

generator/json2test.py
```diff
@@ -0,0 +1,144 @@
...
+                continue
+            print("found missing test data")
+            cls_name = clazz.__name__
+            try:
+                write_test(msg_obj)
```
silverjam, line 132: Pass in filename to write from command line here

silverjam (changes requested):
Good start but need to rebase against master, create the code for cli input and remove any code related to "missing messages"

adrian-kong:
minor change in the usage where it reads from the json msg if id isn't provided

```
python ./json2test.py --in vel_cog.json --out test_vel_cog.yaml
```

generator/json2test.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Swift Navigation SBP Test Generator.")
+    parser.add_argument("-id", "--msgid", default="-1", help="specify message id")
+    parser.add_argument("-in", "--input", help="path to input json file", required=True)
+    parser.add_argument("-out", "--output", help="path to output yaml file", required=True)
+    args = parser.parse_args()
+
+    f = open(args.input)
```
silverjam, line 66: Use `with` statement here, e.g.

```python
with open(args.input) as f:
  msg = ...
```

generator/json2test.py
```diff
@@ -0,0 +1,85 @@
...
+    test_package.write(output, exists_ok=True)
+    print("Wrote {}".format(output))
+    assert_package(output)
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Swift Navigation SBP Test Generator.")
+    parser.add_argument("-id", "--msgid", default="-1", help="specify message id")
+    parser.add_argument("-in", "--input", help="path to input json file", required=True)
+    parser.add_argument("-out", "--output", help="path to output yaml file", required=True)
```
silverjam, line 63: Can just drop the single letter argument variants:

```suggestion
    parser.add_argument("--msg-id", default=None, help="specify message id")
    parser.add_argument("--input", help="path to input json file", required=True)
    parser.add_argument("--output", help="path to output yaml file", required=True)
```

(Note this also changes the default for `--msg-id` to align with the other suggestion below).

generator/json2test.py
```diff
@@ -0,0 +1,85 @@
...
+    args = parser.parse_args()
+
+    f = open(args.input)
+    msg = json.loads(f.read())
+    f.close()
+    msg_type = args.msgid
+    if args.msgid == "-1":
```
silverjam, line 70: More "pythonic" idiom here is:

```suggestion
    if args.msg_id is None:
```

generator/json2test.py
```diff
@@ -0,0 +1,85 @@
...
+        msg_type = msg.get("msg_type")
+    clazz = lookup(msg_type)
```
silverjam, line 72: need to print and error and exit with a failure status code class isn't found:
```python
if clazz is None:
   # print error and exit
```

generator/json2test.py
```diff
@@ -0,0 +1,85 @@
...
+    print("found class: " + str(clazz.__name__))
+    msg_obj = clazz(**msg)
+
+    msg_obj.to_binary()  # sets crc, length, payload
+    try:
+        write_test(msg_obj, args.output)
+    except Exception as e:
+        print("Failed creating test for {}: {}".format(clazz.__name__, e))
+        raise
```
silverjam, line 81:
```suggestion
        raise Exception("Failed creating test for {}: {}".format(clazz.__name__, e)) from e
```

silverjam:
Some brief docs on these two scripts in `HOWTO.md` would be great too.

generator/json2test.py
adrian-kong: i was thinking this as well but wont it keep the file open until the very end

silverjam: The `with` statement only needs to have the `json.loads` call inside of it.  Any variable created inside the `with` will be visible throughout the function.  In Python variable scopes are introduced by functions, not by blocks (as they are in Rust/Java/C/C++).

adrian-kong: oh i did not know that!

silverjam (approved):

PR #1127, opened 2022-04-28
Add support for MSG_SSR_CODE_PHASE_BIASES_BOUNDS message [OTA-119]

fpezzinosn:
# Description

@swift-nav/devinfra

Adds the SSR_CODE_PHASE_BIASES_BOUNDS message to the SSR package.

# API compatibility

It doesn't. It adds a new message.

## API compatibility plan

# JIRA Reference

https://swift-nav.atlassian.net/browse/OTA-119


fpezzinosn:
@swift-nav/devinfra Any feedback on this? This should be very similar to https://github.com/swift-nav/libsbp/pull/1108 and I've incorporated here all the feedback I got there.

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -920,3 +920,57 @@ definitions:
            size_fn: n_sats
            map_by: sat_id
            desc: Orbit and Clock Bounds per Satellite
+
+ - CodeAndPhaseBiasesSatelliteSignal:
```
silverjam, line 924: Any way to shorten this name?

silverjam (approved):
One nit, but otherwise looks good

spec/yaml/swiftnav/sbp/ssr.yaml
fpezzinosn: I can go with `CodePhaseBiasesSatSig`. With a previous name I hit the limit, so its a good idea to keep it short.

silverjam: 👍

PR #1128, opened 2022-04-29
Fixing unconsumed 4096 error sbp2json

adrian-kong:
# Description

Unconsumed 4096 error on sbp2json

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-510

### DIFF
--- a/python/sbp/sbp2json.py
+++ b/python/sbp/sbp2json.py
@@ -107,9 +107,7 @@ def sbp_main(args):
                         m = sbp.table.dispatch(m)
                         dump(args, m)
                     consumed = header_len + m.length + 2
-                except (UnpackError, StreamError):
-                    break
-                except ValueError:
+                except (UnpackError, StreamError, ValueError):
                     consumed = 1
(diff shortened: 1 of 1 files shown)


adrian-kong:
the suggested changes stops the unconsumed 4096 but runs for a long time, not sure how large the file is supposed to be generated but testing this change creates a 1gb json file for a 0.5gb sbp - runs for a long time (was interrupted)

notoriaga:
at some point we removed the JIT support for python and the bindings got _really_ slow, so it might actually be working

silverjam:
What is the behavior of the tool now?

adrian-kong:
we were playing around with handler instead but that generated more messages than the rust version which was odd but the tool is fixed i think, it worked fine with the 10mb file without unconsumed 4096

adrian-kong:
i also made another pull request #1140 which takes out all the manual reading and also fixes the unconsumed 4096

silverjam (approved):

PR #1126, opened 2022-04-28
Add MSG_STATUS_JOURNAL [4.1.x]

silverjam:
# Description

@swift-nav/devinfra

Adds new MSG_STATUS_JOURNAL, see https://swift-nav.atlassian.net/l/c/yYraQXHA

# API compatibility

_Does this change introduce a API compatibility risk?_ -> No, just introducing a new message.

# Related PRs
- https://github.com/swift-nav/libsbp/pull/1123

PR #1129, opened 2022-05-02
ESD-2413 Journaled status reports (master)

RReichert:


RReichert:
Going to close and use https://github.com/swift-nav/libsbp/pull/1126

PR #1131, opened 2022-05-05
Add Atmospheric Corrections and Bounds [OTA-120]

fpezzinosn:
# Description

@swift-nav/devinfra

Added 3 new messages related to Atmospheric Corrections and Bounds
- MSG_SSR_STEC_CORRECTION (new evolution of SBP 1531): One Message per Tile
- MSG_SSR_GRIDDED_CORRECTION_BOUNDS (new extension of SBP 1532, TROPO + Residuals)
- MSG_SSR_TILE_DEFINITION (new evolution of SBP 1526)

# API compatibility

No, just adding new messages.

## API compatibility plan

# JIRA Reference

https://swift-nav.atlassian.net/browse/OTA-120


spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -379,6 +379,53 @@ definitions:
...
+       - update_interval:
+           type: u8
+           desc: >
+             Update interval between consecutive bounds. Similar to
+             RTCM DF391.
+       - sol_id:
+           type: u8
+           desc: SSR Solution ID.
+
+ - MSG_SSR_STEC_CORRECTION_NEW:
```
fpezzinosn, line 403: I'll look for a better name to differentiate it with the existing message, _NEW isn't the best of suffixes

silverjam: Standard procedure here is to add `_DEP` (or `_DEP_A` if `_DEP` already exists) for the existing message and have this new message occupy the old name. I would go ahead and do that in this PR since these messages should be treated like the real thing.

fpezzinosn: And what happens with the IDs of the messages?
The message validator threw the following `Previous ID: 1531, Current ID: 1533`, so I guess I should have kept the original ID.

silverjam: @isaactorz ^^ this is a bug with the message validator that we didn't notice during review, it's a common pattern to rename messages (the ID and contents of the message must remain the same)

silverjam (approved):

PR #1132, opened 2022-05-05
Add Integrity Flags messages [OTA-121]

fpezzinosn:
# Description

@swift-nav/devinfra

These messages are:
- MSG_SSR_FLAG_HIGH_LEVEL
- MSG_SSR_FLAG_SATELLITES
- MSG_SSR_FLAG_TROPO_GRID_POINTS
- MSG_SSR_FLAG_IONO_GRID_POINTS
- MSG_SSR_FLAG_IONO_TILE_SAT_LOS
- MSG_SSR_FLAG_IONO_GRID_POINT_SAT_LOS

# API compatibility

No, just adding new messages to a new package.

## API compatibility plan

# JIRA Reference

https://swift-nav.atlassian.net/browse/OTA-121


silverjam (approved):
@fpezzinosn -- FYI, we need a follow-up task to review the PDF documentation for these messages, see https://github.com/swift-nav/libsbp/blob/staging/docs/sbp.pdf and check out the "units" columns in particular (mainly, the units need to very simple otherwise they'll overflow the layout in the PDF, so any "usage notes" or "caveats" for the units need to move into the message description).

fpezzinosn:
> @fpezzinosn -- FYI, we need a follow-up task to review the PDF documentation for these messages, see https://github.com/swift-nav/libsbp/blob/staging/docs/sbp.pdf and check out the "units" columns in particular (mainly, the units need to very simple otherwise they'll overflow the layout in the PDF, so any "usage notes" or "caveats" for the units need to move into the message description).

Ok, will do. I'll also enable the others that are currently set as private to check how their documentation looks.

PR #1140, opened 2022-05-09
Python sbp2json unconsumed 4096 refactor

adrian-kong:
# Description

Refactors sbp2json to use inbuilt handler to parse file, 
* fixes the unconsumed 4096 but doesn't terminate because file doesn't get closed? 
* also needs to use that __iter__ method to manually iterate through to maintain the order
* also differs from the rust version because the float have 9 extra digits

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-510

### DIFF
--- a/python/sbp/sbp2json.py
+++ b/python/sbp/sbp2json.py
@@ -2,26 +2,17 @@ from __future__ import print_function
 
 import sys
 
-import io
-
 import json
 
-from construct.core import StreamError
+from sbp.client import Handler, Framer
 
-import sbp.msg
-import sbp.table
-
-from sbp.msg import SBP_PREAMBLE
-from sbp.msg import UnpackError
-
-
-DEFAULT_JSON='rapidjson'
-JSON_CHOICES=['json', 'rapidjson']
+DEFAULT_JSON = 'rapidjson'
+JSON_CHOICES = ['json', 'rapidjson']
 try:
     import rapidjson
 except ImportError:
-    DEFAULT_JSON='json'
-    JSON_CHOICES=['json']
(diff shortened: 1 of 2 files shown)


python/sbp/sbp2json.py
```diff
@@ -69,56 +60,12 @@ def dump(args, res):
...
-    if not args:
-        sys.exit(1)
-    try:
-        sbp_main(args)
-    except KeyboardInterrupt:
-        pass
+    if args := get_args():
+        with Handler(Framer(args.file.read, None, verbose=True)) as source:
+            try:
+                it = source.__iter__()
```
joelynch, line 67: It's considered unpythonic (whatever that means) to manually call "dunder" methods.  Use `iter(source)` over `source.__iter__()`.  Better yet use a for loop here:

```python
for msg_meta in source:
    dump(msg_meta)
```

or 

```python
for msg_meta in source:
    if msg_meta is None:
        break
    dump(msg_meta)
```

instead depending on what behaviour you want.

Also don't use the walrus operator unless you are sure that this file will only run on Python >= 3.9!

notoriaga:
> * also differs from the rust version because the float have 9 extra digits

I think that should be expected, python only has 64 bit floats and rust has 32 and 64 so I imagine you're looking at a field that is actually supposed to be a 32 bit float

python/sbp/sbp2json.py
adrian-kong: i tried doing that before but the order kept changing when i ran it, but now when i run it, it changes regardless 

also didnt know it was called the Walrus Operator

PR #1142, opened 2022-05-10
Added input and output parameters to rust sbp2json

adrian-kong:
# Description

`--input` and `--output` parameters to rust sbp2json, json2json, json2sbp

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-738

### DIFF
--- a/rust/sbp2json/src/bin/json2json.rs
+++ b/rust/sbp2json/src/bin/json2json.rs
@@ -1,4 +1,7 @@
+use std::fs::File;
 use std::io;
+use std::io::{Read, Write};
+use std::path::PathBuf;
 
 use clap::Parser;
 use sbp::json::{CompactFormatter, HaskellishFloatFormatter};
@@ -12,10 +15,18 @@ static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
 ///
 /// Typical usage:
 ///
+///     json2json --input [PATH] --output [PATH]
+///
 ///     cat console-json-log.json | json2json
 #[derive(Debug, Parser)]
 #[clap(name = "json2json", verbatim_doc_comment)]
 struct Options {
+    /// Path to input file
+    input: Option<PathBuf>,
+
+    /// Path to output file
+    output: Option<PathBuf>,
+
     /// Print debugging messages to standard error
     #[clap(long)]
     debug: bool,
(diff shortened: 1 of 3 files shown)


rust/sbp2json/src/bin/sbp2json.rs
```diff
@@ -46,8 +59,15 @@ fn main() -> Result<()> {
 
     env_logger::init();
 
-    let stdin = io::stdin();
-    let stdout = io::stdout();
+    let stdin: Box<dyn Read> = match options.input {
+        Some(path) => Box::new(File::open(path).unwrap()),
```
notoriaga, line 63: Should be able to change these `unwrap()`s to `?`s which will print a nicer error message

adrian-kong:
is there a way to make clap do without the -- parameter? i just saw the comment on the ticket, would it have to instead use std::env::args hmm

notoriaga:
> is there a way to make clap do without the -- parameter? i just saw the comment on the ticket, would it have to instead use std::env::args hmm

yeah if you don't specify long/short it should be a positional argument instead of a flag

rust/sbp2json/src/bin/json2json.rs
```diff
@@ -12,10 +15,20 @@ static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
...
 /// Typical usage:
 ///
+///     json2json --input [PATH] --output [PATH]
+///
 ///     cat console-json-log.json | json2json
 #[derive(Debug, Parser)]
 #[clap(name = "json2json", verbatim_doc_comment)]
 struct Options {
+    /// Path to input file
+    #[clap()]
```
notoriaga, line 25: I think you can omit the whole attribute `#[clap()]` if you're not passing anything to it

notoriaga (approved):

rust/sbp2json/src/bin/json2json.rs
```diff
@@ -12,10 +15,18 @@ static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
 ///
 /// Typical usage:
 ///
+///     json2json --input [PATH] --output [PATH]
```
silverjam, line 18: Need to update these example usage strings since we're not using option args (using positional args instead), example:

```
json2json [<INPUT> [<OUTPUT>]{
```

notoriaga: What does --help show if we remove that completely? I feel like that's pretty similar to what clap will generate anyway

silverjam: Yeah, probably better to just remove this

PR #1149, opened 2022-05-18
Clean up the documentation for the new messages [OTA-150]

fpezzinosn:
# Description

@swift-nav/devinfra

Cleaned up the documentation for the new staged messages.

# API compatibility

No, just docstrings updates

## API compatibility plan

# JIRA Reference

https://swift-nav.atlassian.net/browse/OTA-150


fpezzinosn (commented):
@silverjam After fixing the overflow in the units columns, the message structure table for `MSG SSR GRIDDED CORRECTION BOUNDS`, `MSG SSR ORBIT CLOCK BOUNDS` and `MSG SSR ORBIT CLOCK BOUNDS DEGRADATION` still clips the page. I guess a solution could be to write shorter descriptions, but then I'm afraid some information will be missing from the document.

silverjam:
> @silverjam After fixing the overflow in the units columns, the message structure table for `MSG SSR GRIDDED CORRECTION BOUNDS`, `MSG SSR ORBIT CLOCK BOUNDS` and `MSG SSR ORBIT CLOCK BOUNDS DEGRADATION` still clips the page. I guess a solution could be to write shorter descriptions, but then I'm afraid some information will be missing from the document.

We will need to customize rendering for these messages or move the information elsewhere.  Would it make sense to create a dedicated page/table that explains things in more detail or do we need to leave everything in this table?

fpezzinosn:
> > @silverjam After fixing the overflow in the units columns, the message structure table for `MSG SSR GRIDDED CORRECTION BOUNDS`, `MSG SSR ORBIT CLOCK BOUNDS` and `MSG SSR ORBIT CLOCK BOUNDS DEGRADATION` still clips the page. I guess a solution could be to write shorter descriptions, but then I'm afraid some information will be missing from the document.
> 
> We will need to customize rendering for these messages or move the information elsewhere. Would it make sense to create a dedicated page/table that explains things in more detail or do we need to leave everything in this table?

I've moved the repeated explanations to notes in the description and then added a See Note 1 for those. I also abbreviated some common terms. Now all the tables fit.

silverjam:
Or... maybe need to add a special case for stubbed messages to the validator.

PR #1152, opened 2022-05-25
add parallel tox, remove python 2.7 support [DEVINFRA-771]

silverjam:
# Description

@swift-nav/devinfra

  - We can introduce tox parallel mode to speed up the test-python target in libsbp, but we’ll need to upgrade the GHA job to use Ubuntu 20.04 to get a tox that’s new enough to run it

  - (Apparently) this exposes some bugs with Python 2.7 support in Ubuntu, this works locally but could not figure out a fix for GHA… so I propose we try to drop support for this again, Python 2 has been obsolete/discontinued for 2 years now.  Removing support for Python 2.7 ~and Python 3.5~ opens the door for us to introduce type annotations to our Python code.

# API compatibility

_Does this change introduce a API compatibility risk?_  Does not introduce compatibility issues in SBP messages or the API but drops support for Python 2.7 ~and Python 3.5~ (which are both obsolete).

## API compatibility plan

_If the above is "Yes", please detail the compatibility (or migration) plan:_ This is a "wait and see" compatibility plan, the last customer we had that was using Python 2 was able to migrate to Python 3 after some prodding, so hopefully we'll OK this time around.

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-771

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -361,7 +361,7 @@ test-c-v4:
 
 test-python:
 	$(call announce-begin,"Running Python tests")
-	cd $(SWIFTNAV_ROOT)/python/ && tox -vv --parallel auto --skip-missing-interpreters
+	cd $(SWIFTNAV_ROOT)/python/ && tox --parallel auto --skip-missing-interpreters
 	$(call announce-end,"Finished running Python tests")
 
 test-javascript:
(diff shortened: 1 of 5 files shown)


silverjam:
> Get's rid of python 2 basically?

Yes, updated the PR description

adrian-kong (approved):

PR #1153, opened 2022-05-25
docker: add back apt, focal only, tox fixes

silverjam:
# Description

@swift-nav/devinfra

Previous docker image removed apt and a number of other things that are useful for debugging the image after the image has been built, so add these back.

In the interest of simplicity, only support `focal` for the build container.  The purpose of the build container is to provide a consistent build environment for generating bindings, not to provide a build recipe for a bunch of different Ubuntu releases.

Removes unused Docker build files, these haven't been used for a while.

# API compatibility

_Does this change introduce a API compatibility risk?_  No.

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-771


isaactorz (approved):
What was the reason for removing apt to begin with? Save size?

silverjam:
> What was the reason for removing apt to begin with? Save size?

Possibly

PR #1155, opened 2022-06-02
Updates messages to ICD v1.2 [OTA-161]

fpezzinosn:
# Description

@swift-nav/devinfra

Updates the name of two navigation messages to better reflect their purpose.
Also updates how the bitmask field should be interpreted in the Integrity Flags messages.
There is also an update on how to interpret the Bound values for the Orbit and Clock Bound message.
The frame transformation message was updated to use fixed-length null-terminated strings for the frames names.
The Gridded Correction Bounds message has two new fields.

# API compatibility

There's no change in the fields of the messages themselves, only on how the data should be interpreted.

## API compatibility plan

# JIRA Reference

https://swift-nav.atlassian.net/browse/OTA-161


spec/yaml/swiftnav/sbp/navigation.yaml
```diff
@@ -2479,27 +2479,23 @@ definitions:
           units: s
           desc: Leap second count after insertion.
 
- - MSG_ITRF:
+ - MSG_REFERENCE_FRAME_PARAM:
      id: 0x0244
      short_desc: Reference Frame Transformation Parameter
```
jtec, line 2484: nit: "Reference Frame Transformation Parameter**s**"

silverjam (approved):
Assuming message content has already been approved then this looks good to me.  Hopefully the PDF docs look reasonable too?

silverjam:
> Assuming message content has already been approved then this looks good to me. Hopefully the PDF docs look reasonable too?

I did some spot checks on things that looked questionable and it looks fine, but not sure if a more thorough read through has been done already.

fpezzinosn:
> > Assuming message content has already been approved then this looks good to me. Hopefully the PDF docs look reasonable too?
> 
> I did some spot checks on things that looked questionable and it looks fine, but not sure if a more thorough read through has been done already.

The two new fields for Gridded Correction Bounds made the table overflow (again), so I had to abridge some of the descriptions (Hopefully not beyond the point of usefulness). I also did a comparison with the current ICD, and unless I missed something, everything looks good.

PR #1156, opened 2022-06-02
Update navigation stubbed msgs names [OTA-161]

fpezzinosn:
# Description

@swift-nav/devinfra

Updates the names of two of the stubbed navigation messages.
GPSLeapSecond -> UTCLeapSecond
ITRF -> FrameTransformationParams
These are more appropriate names for the proposed new messages.

# API compatibility

Two messages are changing names, but since they are stubbed messages not yet published, no one should be consuming them.

## API compatibility plan

# JIRA Reference

https://swift-nav.atlassian.net/browse/OTA-161


fpezzinosn:
The generated artifacts is failing at the `make html-python` step. There isn't any useful error message in GA, but when running in my local docker image I get:
```
Extension error:
Could not import extension sphinxcontrib.spelling (exception: The 'enchant' C library was not found and maybe needs to be installed.
See  https://pyenchant.github.io/pyenchant/install.html
for details
)
```
This was solved by installing `python-enchant`.

The other check that is failing is sonarcloud. It doesn't like the how one of the functions in the template is using `var` to create a variable. It's marking that as a code smell.

fpezzinosn:
I fixed the sonarcloud check by updating the javascript template to use `let` instead of `var`.
And regarding the target `html-python`, Sphinx had language set to `None`, which raised a warning that was treated as error. I fixed this by setting the language to English

Now there's a new failure related to `py36-sbp2json-construct2_9_33` and only that target, all the other python runs pass.

python/tox.ini
```diff
@@ -29,7 +29,7 @@ sitepackages = False
 [coverage:run]
 branch = True
 include = sbp/*
-parallel = true
+parallel = True
```
fpezzinosn, line 32: This is why `py36-sbp2json-construct2_9_33` was failing.

silverjam (approved):

notoriaga (approved):

PR #1157, opened 2022-06-03
Python 3.10 update [DEVINFRA-769]

adrian-kong:
# Description

@swift-nav/devinfra

Releases sbp v4.3.2 so python 3.10 shows on languages in sbp

https://pypi.org/project/sbp/

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-769


python/deploy.py
```diff
@@ -73,7 +73,7 @@ def run_bdist(deploy_dir):
 
     shutil.copytree(os.path.join(repo_dir, ".git"), ".git")
 
-    shutil.copy(os.path.join(script_dir, ".coveragerc"), "module/.coveragerc")
+    # shutil.copy(os.path.join(script_dir, ".coveragerc"), "module/.coveragerc")
```
joelynch, line 76: Did we work out if we need to add the converagerc back in?

adrian-kong: nope i think we dont need it but im not sure if i should leave it as it was when i released it

python/sbp/RELEASE-VERSION
```diff
@@ -1 +1 @@
-4.3.1-alpha
\ No newline at end of file
+4.3.1
```
silverjam, line 1: Should drop this change, anything that's not a released version needs to maintain the `-alpha` tag.

python/deploy.py
silverjam: Just delete this like, no since in maintaining cruft

silverjam (changes requested):

python/sbp/RELEASE-VERSION
adrian-kong: should i update that to 4.3.2-alpha instead of 4.3.1, since i deployed it as 4.3.2, (couldnt deploy as 4.3.1 for some reason)

silverjam: After release there is supposed to be a "prep for next release" process which bumps this to `<next_bugfix_version>-alpha` -- so if 4.3.2 was just released, this should be `4.3.3-alpha` -- this happens automatically if you rerun the generator on a non-tagged commit, e.g. https://github.com/swift-nav/libsbp/commit/b423ed3e487c8a6890dc57ef842989ed3ac58e10

silverjam (approved):

PR #1161, opened 2022-06-06
Locked stdin and out for rust sbp2json

adrian-kong:
# Description

Locks the stdin and stdout for sbp2json

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-790

### DIFF
--- a/rust/sbp2json/src/bin/json2json.rs
+++ b/rust/sbp2json/src/bin/json2json.rs
@@ -53,12 +53,12 @@ fn main() -> Result<()> {
 
     let stdin: Box<dyn Read> = match options.input {
         Some(path) => Box::new(File::open(path)?),
-        _ => Box::new(io::stdin()),
+        _ => Box::new(io::stdin().lock()),
     };
 
     let stdout: Box<dyn Write> = match options.output {
         Some(path) => Box::new(File::create(path)?),
-        _ => Box::new(io::stdout()),
+        _ => Box::new(io::stdout().lock()),
     };
 
     if options.float_compat {
--- a/rust/sbp2json/src/bin/json2sbp.rs
+++ b/rust/sbp2json/src/bin/json2sbp.rs
@@ -48,12 +48,12 @@ fn main() -> Result<()> {
 
     let stdin: Box<dyn Read> = match options.input {
         Some(path) => Box::new(File::open(path)?),
-        _ => Box::new(io::stdin()),
+        _ => Box::new(io::stdin().lock()),
     };
 
     let stdout: Box<dyn Write> = match options.output {
         Some(path) => Box::new(File::create(path)?),
-        _ => Box::new(io::stdout()),
+        _ => Box::new(io::stdout().lock()),
     };
 
     json2sbp(stdin, stdout, options.buffered, options.fatal_errors)
--- a/rust/sbp2json/src/bin/sbp2json.rs
+++ b/rust/sbp2json/src/bin/sbp2json.rs
@@ -57,12 +57,12 @@ fn main() -> Result<()> {
 
     let stdin: Box<dyn Read> = match options.input {
         Some(path) => Box::new(File::open(path)?),
-        _ => Box::new(io::stdin()),
+        _ => Box::new(io::stdin().lock()),
     };
 
     let stdout: Box<dyn Write> = match options.output {
         Some(path) => Box::new(File::create(path)?),
-        _ => Box::new(io::stdout()),
+        _ => Box::new(io::stdout().lock()),
     };
 
     if options.float_compat {


notoriaga (approved):
for edification this avoids acquiring a lock on a mutex for each read/write. the lock is always uncontested so in reality it's not a big deal (especially after the std mutex improvements). you still see a small per boost though

```
hyperfine 'sbp2json <./tests/data/regression/2020-06-25-BayLoop/STEP_v2_SR_L2_OSR_staging_10Hz_GNSSOnly_AmotechL2/25-190704.sbp >/dev/null'
Benchmark 1: sbp2json <./tests/data/regression/2020-06-25-BayLoop/STEP_v2_SR_L2_OSR_staging_10Hz_GNSSOnly_AmotechL2/25-190704.sbp >/dev/null
  Time (mean ± σ):      5.132 s ±  0.027 s    [User: 4.896 s, System: 0.236 s]
  Range (min … max):    5.105 s …  5.194 s    10 runs

hyperfine 'sbp2json <./tests/data/regression/2020-06-25-BayLoop/STEP_v2_SR_L2_OSR_staging_10Hz_GNSSOnly_AmotechL2/25-190704.sbp >/dev/null'
Benchmark 1: sbp2json <./tests/data/regression/2020-06-25-BayLoop/STEP_v2_SR_L2_OSR_staging_10Hz_GNSSOnly_AmotechL2/25-190704.sbp >/dev/null
  Time (mean ± σ):      4.962 s ±  0.044 s    [User: 4.712 s, System: 0.249 s]
  Range (min … max):    4.900 s …  5.037 s    10 runs
```

notoriaga:
I think the dockerfile in the rust folder needs to be updated to a newer version of rust. I'm guessing since 1.55 some borrowing rule was relaxed because the code compiles in the other workflows which use the latest version

rust/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.55-slim-buster
+FROM rust:1.61-slim-buster
```
silverjam, line 1: there's a benchmark dockerfile that probably needs to be updated too: https://github.com/swift-nav/libsbp/blob/master/python/Dockerfile.benchmark

adrian-kong: hmmmmm that shouldnt affect the test error `./rust/bin/sbp2json: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.25' not found (required by ./rust/bin/sbp2json)` but maybe updating that to python3.10? or make it install `apt get musl` as well will have to test that

silverjam: This is using GLIBC (not MUSL) -- this error happens when the glibc the tool is linked against is newer than the glic that the system suports -- try changing `FROM python:3.7-slim-stretch` to `FROM python:3.7-slim-buster`.

silverjam: @adrian-kong Note that the released version of `sbp2json` (the one that goes to the releases page) does use musl, but this is so that we can release a statically linked version of the tool that doesn't require a specific version of glibc.

silverjam: > @adrian-kong Note that the released version of `sbp2json` (the one that goes to the releases page) does use musl, but this is so that we can release a statically linked version of the tool that doesn't require a specific version of glibc.

Actually, this bit about using musl isn't true... I thought it was, but apparently not:
```
✦ ❯ ldd ~/.local/share/swift-cli/packages/libsbp/4.1.6/contents/sbp2json
        linux-vdso.so.1 (0x00007fff1cdce000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f1f6d491000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f1f6d487000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f1f6d464000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f1f6d272000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f1f6d893000)
```

silverjam:
@adrian-kong Next thing we can try is to use MUSL to work around this issue... basically anywhere we call `cargo build --release` add `--target=x86_64-unknown-linux-musl` you'll probably need to reintroduce the installation of `musl-tools` to the Dockerfiles

adrian-kong:
hmmmm if we change the target, it changes the target path from `target/release` -> `target/x86_64-unknown-linux-musl/release`, should we just change all the OS to use their specific binary targets so it wont just be `target/release `

scripts/ci_build_rust.bash
```diff
@@ -19,7 +19,11 @@ else
...
 fi
 
-cargo build --all --release
+if [ "$RUNNER_OS" == "Linux" ]; then
+  cargo build --all --release --target=x86_64-unknown-linux-musl
+else
+  cargo build --all --release
+fi
 
 cd target/release
```
adrian-kong, line 28: changing this temporarily for now

```suggestion
cd target/x86_64-unknown-linux-musl/release
```

silverjam:
> hmmmm if we change the target, it changes the target path from `target/release` -> `target/x86_64-unknown-linux-musl/release`, should we just change all the OS to use their specific binary targets so it wont just be `target/release `

I think it only does this if you're cross compiling, or manually specifying the build target.

silverjam:
I think the latest failure is from the upgrade of hyperfine, the format of the json it outputs has probably changed... you'll need to run it locally to see what the new structure looks like (or rollback the upgrade if there wasn't a reason it was needed).

scripts/ci_benchmark_install.bash
```diff
@@ -2,7 +2,7 @@
 
 set -ex
 
-HYPERFINE_RELEASE=v1.9.0
+HYPERFINE_RELEASE=v1.14.0
```
adrian-kong, line 5:
```suggestion
HYPERFINE_RELEASE=v1.9.0
```

silverjam (approved):
Some slight slowdowns due to the switch to musl but I think we can live with that to get more universally compatible binaries

PR #1163, opened 2022-06-08
update 'target' path

silverjam:
This should be the next change that's needed @adrian-kong


adrian-kong (approved):
oh i forgot to update the `rust_bins` line in the other pr

PR #1164, opened 2022-06-08
only on linux

silverjam:
cc @adrian-kong


adrian-kong:
do we even need this if `ci_benchmark.bash` doesnt call this build script? we could just leave it as it was before and only keep the target changes in benchmark but its missing parameter now for benchmark

silverjam:
> do we even need this if `ci_benchmark.bash` doesnt call this build script? we could just leave it as it was before and only keep the target changes in benchmark but its missing parameter now for benchmark

We don't strictly need it, but I think we should benchmark what we distribute and it'll be an added bonus to release a statically linked tool, so we're not dependent on glibc.

adrian-kong (approved):

PR #1166, opened 2022-06-09
docker and doc fix-ups

silverjam:
# Description

- numpy is technically needed for the fftmonitor, but since that's the only thing that uses it I think we've intentionally omitted it from our dependency list... this breaks the documentation generator though, since it tries import the package-- previously this was installed via the generator (somehow) -- now we're explicitly installing the dependency in the tox.ini command that's used to run the documentation generator

- Don't run autoremove in the dockerfile, this seems to be uninstalling `enchant` or something else that the `pyenchant` package needs

- Fix Sphinx config error seen in recent release:
```
 sphinx-build -b spelling -d build/doctrees  -W source build/spelling
Running Sphinx v5.0.1
Warning, treated as error:
Invalid configuration value found: 'language = None'. Update your configuration to a valid langauge code. Falling back to 'en' (English).
Creating file ./sbp.rst.
Creating file ./sbp.acquisition.rst.
Creating file ./sbp.bootload.rst.
```

# API compatibility

N/A


silverjam:
Force merging since this has to do with stability of the build

PR #1167, opened 2022-06-09
fixing rust build script

adrian-kong:
# Description

right now it doesn't upload the linux build

since linux specifies binary target, directory becomes
`/target/x86_64-unknown-linux-musl/release`
where normally it should be
`/target/release`

and we want to zip it up to end up in the top folder 
`/`
so we will just zip it up with an extra level

this is due to #1161


silverjam (approved):

silverjam:
@adrian-kong Where we able to validate this?  Or do we need to wait until the next release?

adrian-kong:
> @adrian-kong Where we able to validate this? Or do we need to wait until the next release?

here https://github.com/swift-nav/libsbp/runs/6806332387?check_suite_focus=true
i've uploaded the binary to the 4.3.3 release

PR #1168, opened 2022-06-09
Add IMU type for ICM-42670 [ESD-2445]

woodfell:
# Description

@swift-nav/devinfra

Add a documentation for a new IMU type value (3) in IMU_AUX messages for TDK ICM-42670

Mark IMU type 2 as "reserved" - it is currently in use but we haven't been making it public so far

# API compatibility

Documentation change only

## API compatibility plan

N/A

# JIRA Reference

https://swift-nav.atlassian.net/browse/ESD-2445


spec/yaml/swiftnav/sbp/imu.yaml
```diff
@@ -86,6 +86,8 @@ definitions:
                   values:
                       - 0: Bosch BMI160
                       - 1: ST Microelectronics ASM330LLH
+                      - 2: Reserved
```
woodfell, line 89: This value is currently used for Android but I'm not sure we need to make it public - @reimerix ?

silverjam (approved):

jithk (approved):
Thanks for doing this Matt

notoriaga (approved):

PR #1176, opened 2022-06-27
Add signature and certificate messages [GV2-74]

samvrlewis:
# Description
Adds the initial version of the SBP signature and certificate messages
as first proposed in https://swift-nav.atlassian.net/wiki/spaces/ENG/pages/2234843801/Secure+Corrections+Client+Interface+Design

# API compatibility
Messages aren't 100% finalized yet, so shouldn't be promoted from staging yet. Having them here will allow for easier prototyping though.

# JIRA Reference

https://swift-nav.atlassian.net/browse/GV2-74


spec/yaml/swiftnav/sbp/signing.yaml
```diff
@@ -0,0 +1,50 @@
...
+package: swiftnav.sbp.signing
+description: Messages relating to signatures
+stable: False
+public: False
+include:
+  - types.yaml
+  - gnss.yaml
+definitions:
+  - MSG_ED25519_SIGNATURE:
+      id: 0x0BD2
```
samvrlewis, line 20: Not sure if there's a rule of thumb when selecting these IDs but I just chose the next available.

spec/yaml/swiftnav/sbp/signing.yaml
```diff
@@ -0,0 +1,55 @@
...
+            fill: u32
+            desc: CRCs of signed messages.
+  - MSG_ED25519_CERTIFICATE:
+      id: 0x0BD3
+      short_desc: ED25519 certificate, split over multiple messages
+      fields:
+        - message_number:
+            type: u8
+            desc: Message number out of total_messages (0 indexed).
+        - total_messages:
```
silverjam, line 44: Do we need 2 bytes for this information?  Can we use half a byte?  15*(255-20-1) would allow 3210 bytes for a certificate -- given the Ed25519 cert that I generated manually was about 250 bytes it seems like this should be enough.

spec/yaml/swiftnav/sbp/signing.yaml
silverjam: let's see if we can establish a "group" of message by selecting a unique first byte so that we can have related messages grouped together by ID, for example 0xEF01, 0xEF02 (no idea if these message IDs are actually available though).

spec/yaml/swiftnav/sbp/signing.yaml
```diff
@@ -0,0 +1,55 @@
...
+            fill: u8
+            size: 64
+            desc: ED25519 signature for messages.
+        - fingerprint:
+            type: array
+            fill: u8
+            size: 20
+            desc: SHA-1 fingerprint of the associated certificate.
+        - signed_messages:
+            type: array
```
silverjam, line 34: I think we should split this into two messages so that we can provide the client with a message that will allow them to size a buffer that's needed for signing and to easily identify when a group of messages is complete (potentially without having to consult something like a hash map).

So first message would contain something like:
- Count of message and/or size of the signature buffer
- List of CRCs included in signature
- Some sort of signature ID (random or possibly a hash of the CRCs or the signature buffer itself).

Second message would contain:
- Signature ID
- Fingerprint of cert used to generate signature
- The signature itself

spec/yaml/swiftnav/sbp/signing.yaml
samvrlewis: Good idea, will do

spec/yaml/swiftnav/sbp/signing.yaml
samvrlewis: Could do. As far as I can see, libsbp doesn't have a nice way of encoding integers less than 8 bit though, so I  guess we'd have to do something in the spec like:

```
        - message_num:
            type: u8
            fields:
                  - 7-4:
                      desc: Total number of messages
                          - 0: 0 messages total
                          - 1: 1 messages total
                          - 2: 2 messages total
                          - ...etc..
                  - 0-3:
                      desc: Message number out of total_messages (0 indexed).
                      values:
                          - 0: 1st message
                          - 1: 2nd message
                          - 2: 3rd message
                          - ..etc..
```

Unless there's a nicer way of doing it? Maybe not worth polluting the spec with all that info to save a byte?

spec/yaml/swiftnav/sbp/signing.yaml
samvrlewis: Isn't this only helpful/useful if we can assume the message stream is well ordered? Or have I misunderstood?

If we can rely on the message stream being ordered then I think we could achieve a similar thing by sending the signature message prior to the messages that it signs, avoiding the need for two messages.

I was of the understanding that we shouldn't/couldn't rely on ordering for RTCM/SBP (even if we might be able to say its OK for the particular customer that this is for).

spec/yaml/swiftnav/sbp/signing.yaml
```diff
@@ -0,0 +1,55 @@
...
+      short_desc: ED25519 certificate, split over multiple messages
+      fields:
+        - message_number:
+            type: u8
+            desc: Message number out of total_messages (0 indexed).
+        - total_messages:
+            type: u8
+            desc: Total number of messages.
+        - fingerprint:
+            type: array
```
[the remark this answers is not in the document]
samvrlewis: Same question about ordering here I guess.

Maybe a different way around having to send the 20 byte buffer each time would be to associate each group of cert messages with a random byte. Considering that we're only rarely going to rotate certs, this should probably give us enough of a guarantee of avoiding clashes.

spec/yaml/swiftnav/sbp/signing.yaml
silverjam: When we've done this in other messages (MSG_OBS) it's typically just indicated in the docs that the top nibble is the index and the bottom nibble is the count.

spec/yaml/swiftnav/sbp/signing.yaml
silverjam: Yeah, getting the feeling I didn't think this through all the way.  There was a question about latency introduced and buffer sizes from the client and was hoping we could come up with a way to size the buffer before the messages arrive while allowing us to pass through the messages.  Obviously if we sent the signature ahead of the messages we could do this easily but it'd mean we'd need to hold these messages in a buffer, generate the signature, then send the signature message ahead of the message group.

One way to do this might be to assume a signature buffer with a fixed maximum size (that is, it'd always be sized to hold 44 messages).  Each block in the buffer would be 255 byte blocks that would be filled by messages.

To find where a message would slot into the buffer we could use the CRC has a hash value, then use open indexing with linear probing to find where to slot the message: https://www.geeksforgeeks.org/implementing-hash-table-open-addressing-linear-probing-cpp/.

This means we'd always be signing a buffer with a fixed size, and many of the bytes would be zero in the signature buffer for smaller groups.

This would require always having 11220 (44*255) bytes for the sig buffer, but this is the upper maximum that would be needed anyway.

samvrlewis: Interesting approach! 

> 11220 (44*255) bytes for the sig buffer,

We need to sign RTCM messages, not SBP messages, so I think this would actually be 44KB (44*1024).

I also wonder if it is worth building this level of complexity into the protocol? I think a client could implement a similar scheme themselves, if they wanted to, by having a similar fixed-size hashmap (which, if I'm understanding correctly, is essentially what we'd be building server side in your proposal)? In this case, they'd implement something like (excuse the very rough sketch):

```
void on_msg(uint32_t crc, void* buf) {
     fixed_hashmap->insert(crc, buf);
}

bool on_signature(uint32_t* crcs, uint32_t* sig) {
     uint8_t buf[max_size];
     for (crc in crcs) {
       buf.insert(fixed_hashmap->get(crc));
     }

     return verify_sig(sig, buf);
}
```

And in this way, it doesn't preclude clients that are happy to use dynamic memory from implementing the verification in a simpler fashion, if they chose to (by using a standard hashmap).

And not to reroute to start talking about message ordering again, but if we need to account for to non-guaranteed message ordering then I think a client could still need to buffer messages from other groups if the full group hasn't arrived yet. I guess in practice, if a client needed to use fixed buffers, you'd have to start dropping the oldest (by arrival time?) messages once you've buffered as much as possible.

samvrlewis: > Obviously if we sent the signature ahead of the messages we could do this easily but it'd mean we'd need to hold these messages in a buffer, generate the signature, then send the signature message ahead of the message group.

Also not sure if I see the problem with this. We need to hold messages in a buffer to generate the signature, regardless and the current proposal is to sign messages that are generated together, together.

If latency is the issue, I think this would practicably only be an issue if we started grouping messages that are generated at different times together, which I don't think we're currently intending to do (or at least isn't how I've implemented this for now).

silverjam: > > Obviously if we sent the signature ahead of the messages we could do this easily but it'd mean we'd need to hold these messages in a buffer, generate the signature, then send the signature message ahead of the message group.
> 
> 
> 
> Also not sure if I see the problem with this. We need to hold messages in a buffer to generate the signature, regardless and the current proposal is to sign messages that are generated together, together.
> 
> 
> 
> If latency is the issue, I think this would practicably only be an issue if we started grouping messages that are generated at different times together, which I don't think we're currently intending to do (or at least isn't how I've implemented this for now). (Added you to the go code PR in case it helps get a sense of how this is working ATM)

Yeah, it's maybe a moot point, but I was thinking this approach would essentially allow us to call `send(msg)` on messages as we're putting them into a buffer for a signature, the client would them be able to build its signature buffer in lock step with the server with this approach (and would be able to validate the signature with this "prefilled" buffer as soon as we send it).  But again, the latency improvements here are probably really small.

spec/yaml/swiftnav/sbp/signing.yaml
samvrlewis: Claimed the `0x0C` "namespace" for these messages

spec/yaml/swiftnav/sbp/signing.yaml
samvrlewis: Have changed to do this in the same way MSG_OBS does it now

spec/yaml/swiftnav/sbp/signing.yaml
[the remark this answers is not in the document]
samvrlewis: I've left this as it is for now, as I feel like it makes it a bit more straight-forward and consistent to have the same fingerprint passed around as in the signature messages. My thinking is that the cert messages aren't going to be sent super often so we can live with the bit of overhead for consistency and easier server/client implementations.

spec/yaml/swiftnav/sbp/signing.yaml
samvrlewis: Given there's not likely to be a big improvement in latency for the amount of complexity this brings, I think I'll just leave it as it is. Sending the signature first I think should allow clients to have a simpler implementation. (Though I think there is likely some complexity in properly handling edge cases like messages in a group not arriving - but I don't think there's any nice way around that).

spec/yaml/swiftnav/sbp/signing.yaml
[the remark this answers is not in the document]
silverjam: 👍

spec/yaml/swiftnav/sbp/signing.yaml
silverjam: Sounds good

silverjam (approved):

silverjam:
Will need to add a test vector for the new messages before SonarCloud will pass.  The json2test tool that @adrian-kong  wrote will help with this.

silverjam:
@samvrlewis can we introduce stubs for these messages on master?

samvrlewis:
> @samvrlewis can we introduce stubs for these messages on master?

Will do :+1:

PR #1177, opened 2022-07-01
Add example json2test instructions

samvrlewis:
Also installs some dependencies to the Dockerfile so that json2test can
run in the libsbp docker image.

# Description

@swift-nav/devinfra

# API compatibility

No

# JIRA Reference


silverjam (approved):
🤘

PR #1178, opened 2022-07-01
Add a tip to the HOWTO for new message groups

samvrlewis:
# Description

I was surprised that these files weren't auto-generated, but listing
this as an extra step will hopefully avoid confusion for developers
adding new message groups in the future.

# API compatibility

No

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -40,7 +40,11 @@ process. This is likely to change in the future.
 2. Increment `number_of_messages` in `python/tests/sbp/test_table.py`
    by the corresponding number of new messages.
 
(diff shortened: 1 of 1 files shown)


samvrlewis:
IMO this should probably be part of the code generation step, as it's really easy to forget adding this (and seemingly only caught if you add message specific tests - so luckily I did that!), so I added a couple of bugs to track: https://swift-nav.atlassian.net/browse/DEVINFRA-804?jql=issueKey%20in%20(DEVINFRA-803%2CDEVINFRA-804)

silverjam (approved):

PR #1179, opened 2022-07-01
[DEVINFRA-806] Updated CHANGELOG and prep for next release

adrian-kong:
# Description

Updates changelog and prep for release 4.5.1

PR #1180, opened 2022-07-04
Javadoc escaping < >

adrian-kong:
# Description

@swift-nav/devinfra

Generator escaping < > 

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-805


generator/sbpg/targets/java.py
```diff
@@ -67,7 +67,7 @@ def commentify(environment: Environment,
   """
   Builds a comment.
   """
-  return indented_wordwrap(environment, value, indent=(" " * indent) + " * ", first=False)
+  return indented_wordwrap(environment, value.replace("<", "{@code <}").replace(">", "{@code >}"), indent=(" " * indent) + " * ", first=False)
```
samvrlewis, line 70: Wasn't `@literal` the one you originally proposed to use? It seems to me to be more appropriate here than `code`?

samvrlewis, line 70: What do you think about defining another function for stripping out the tags? Just to avoid having to repeat the `replace` calls in multiple places?

Maybe something like

```python
def escape_comment_tags(comment: str):
  """
  Escape tags in a Javadoc comment string
  """
  return comment.replace("<", "{@literal <}").replace(">", "{@literal >}")
```

Then I think you could either call that in `commentify` or in the templates.

samvrlewis (approved):
LGTM!

samvrlewis (commented):
Actually, looks like the tests are out of date? Do you need to `make all`?

silverjam:
> > Actually, looks like the tests are out of date? Do you need to `make all`?
> 
> Yep, would it be better to merge this into #1179 and do the make all over there

👍

silverjam (approved):

PR #1181, opened 2022-07-05
Enforce Java 11 in build gradle

adrian-kong:
# Description

Should resolve Jacoco issue where `gradle jacocoTestReport` fails to run due to major version upgrade


silverjam (approved):

notoriaga:
What does this mean for consumers of the library, do they have to use java 11 or is this just for our ci?

notoriaga (approved):

PR #1182, opened 2022-07-05
Adds javadoc generator test

adrian-kong:
# Description

Checks whether javadocs can be built successfully to prevent gradle errors

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-807

### DIFF
--- a/generator/sbpg/targets/java.py
+++ b/generator/sbpg/targets/java.py
@@ -60,6 +60,13 @@ def classnameify(s):
   return ''.join(w if w in ACRONYMS else w.title() for w in s.split('_'))
 
 
+def escape_comment_tags(comment: str):
+    """
+    Escape tags in a Javadoc comment string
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -504,6 +504,11 @@ java-examples:
 	cd $(SWIFTNAV_ROOT)/java/example && gradle clean fatjar
 	$(call announce-end,"Finished building fat jar")
 
+javadoc javadocs:
```
silverjam, line 507: We could make the dependency of the `Java` target so that it runs locally-- would there be any downsides to that?

silverjam: If errors are rare then this is a good place to verify, making everyone check it locally sounds unnecessary

silverjam (approved):

notoriaga (approved):

PR #1190, opened 2022-07-14
make serde feature depend on serde-big-array

silverjam:
# Description

@swift-nav/devinfra

Makes `serde` and `serde-big-array` go together for the `serde` feature, since some users only want to enable `serde` and not the `json` feature.

# API compatibility

_Does this change introduce a API compatibility risk?_

Should not, no messages are changed, but enables just the "serde" feature of the Rust library to be used in isolation from the "json" feature.

## API compatibility plan

_If the above is "Yes", please detail the compatibility (or migration) plan:_

n/a

# References

## Jira
No Jira, but example breakage here: https://github.com/swift-nav/gnss-converters-private/runs/7347817014?check_suite_focus=true

## Related PRs
- https://github.com/swift-nav/gnss-converters-private/pull/1226

### DIFF
--- a/rust/sbp/Cargo.toml
+++ b/rust/sbp/Cargo.toml
@@ -20,7 +20,8 @@ readme = "../../README.md"
 [features]
 default = []
 async = ["futures", "dencode/async"]
-json = ["serde", "serde_json", "serde-big-array", "base64"]
+serde = ["dep:serde", "serde-big-array"]
+json = ["serde", "serde_json", "base64"]
 float_roundtrip = ["serde_json/float_roundtrip"]
 link = ["slotmap"]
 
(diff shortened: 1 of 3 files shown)


notoriaga (approved):

rust/sbp/Cargo.toml
```diff
@@ -20,7 +20,8 @@ readme = "../../README.md"
 [features]
 default = []
 async = ["futures", "dencode/async"]
-json = ["serde", "serde_json", "serde-big-array", "base64"]
+serde = ["dep:serde", "serde-big-array"]
```
samvrlewis, line 23: Ah, nice, I was wondering how you could do something like this

samvrlewis (approved):

PR #1196, opened 2022-07-24
Added json sbp rust tests

adrian-kong:
# Description

Adding tests for converting json to sbp messages

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-821

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5546,9 +5546,9 @@
       }
     },
     "node_modules/terser": {
-      "version": "4.6.6",
-      "resolved": "https://registry.npmjs.org/terser/-/terser-4.6.6.tgz",
-      "integrity": "sha512-4lYPyeNmstjIIESr/ysHg2vUPRGf2tzF9z2yYwnowXVuVzLEamPN1Gfrz7f8I9uEPuHcbFlW4PLIAsJoxXyJ1g==",
+      "version": "4.8.1",
+      "resolved": "https://registry.npmjs.org/terser/-/terser-4.8.1.tgz",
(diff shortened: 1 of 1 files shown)


silverjam:
I re-kicked Python, not sure why it's failing

silverjam:
> I re-kicked Python, not sure why it's failing

Still failing unfortunately

silverjam:
> > I re-kicked Python, not sure why it's failing
> 
> Still failing unfortunately

I disabled the failing step. we should log a jira to revisit

silverjam (approved):
@adrian-kong can you manually modify one of the generated tests so that we can verify that it fails?  Or have you already done this?

silverjam:
> > @adrian-kong can you manually modify one of the generated tests so that we can verify that it fails? Or have you already done this?
> 
> I'm not sure what you mean by that because the generated tests don't fail as ...

Part of writing tests should be seeing them fail at least once, so a quick tweak to one of the generated tests just to see one of the asserts fire would be enough to verify that the tests are actually validating something.  Might seem silly, but it's a good sanity check to get in the habit of.

silverjam:
> https://github.com/swift-nav/libsbp/runs/7351274963?check_suite_focus=true
> 
> which previously passes suddenly stops passing...

This action isn't pinning the versions of anything, so things can change out from underneath us, which likely what happened here.

silverjam:
> also the benchmark is weird not sure why that fails sometimes

Benchmarks are pretty heavily dependent on the underlying machine type and it's load.  We've tried to control some of this by using docker and restricting the number of CPUs that can be used but it's still an approximate measure.

The main thing we need to make sure is that the benchmarks can catch big slowdowns.  So I've been tweaking values to get things passing as needed when there's a perf change.  The recent addition around float precision changed perf (and the benchmark CI check worked).  But we might need more tweaking to find a stable value.

PR #1197, opened 2022-07-25
Deprecate SSR gridded correction message

mfine:
# Description

Add a deprecated SSR gridded correction message so that there are not deprecated and non-deprecated SSR messages composing the SSR atmospheric message.

@swift-nav/devinfra

# API compatibility

This removes a API compatibility risk with using a non-deprecated message amongst deprecated messages.

## API compatibility plan

# JIRA Reference

### DIFF
--- a/c/include/libsbp/cpp/message_traits.h
+++ b/c/include/libsbp/cpp/message_traits.h
@@ -6074,6 +6074,43 @@ struct MessageTraits<sbp_msg_ssr_gridded_correction_dep_a_t> {
 };
 
 template <>
+struct MessageTraits<sbp_msg_ssr_gridded_correction_dep_t> {
+  static constexpr sbp_msg_type_t id = SbpMsgSsrGriddedCorrectionDep;
+  static const sbp_msg_ssr_gridded_correction_dep_t &get(const sbp_msg_t &msg) {
+    return msg.ssr_gridded_correction_dep;
+  }
+  static sbp_msg_ssr_gridded_correction_dep_t &get(sbp_msg_t &msg) {
+    return msg.ssr_gridded_correction_dep;
+  }
+  static void to_sbp_msg(const sbp_msg_ssr_gridded_correction_dep_t &msg,
+                         sbp_msg_t *sbp_msg) {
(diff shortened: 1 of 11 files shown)


mfine:
Not all language files have been generated yet.

joelynch:
Hey Mark I was chatting to @theureau recently and discovered that I had misinterpreted the messages intended to be produced from the SsrAtmo message in the integrity project.

Apparently we need to be emitting
- `msg_ssr_gridded_correction_bounds_t`
or
- `msg_ssr_gridded_correction_t`

rather than what I had assumed as

- `msg_ssr_gridded_correction_bounds_t`
- `msg_ssr_gridded_correction_t`
or
- `msg_ssr_gridded_correction_t`

So I think we should deprecate `msg_ssr_gridded_correction_t` but not introduce a new message as it is being replaced by `msg_ssr_gridded_correction_bounds_t`.

Can you confirm this is the case @fpezzinosn and @lkloh ?

I should have a PR up soon that will make some modifications to the SbpDecoder to support this.

mfine:
@joelynch if `msg_ssr_gridded_correction_t` won't be used in the new message path, then you wouldn't need to deprecate it - you'll be able to tell whether you're using deprecated or non-deprecated messages without a "dummy" deprecated message. (even if messages are no longer used, they don't get deprecated - only when there's another message replacing it).

fpezzinosn:
> Apparently we need to be emitting
> 
> * `msg_ssr_gridded_correction_bounds_t`
>   or
> * `msg_ssr_gridded_correction_t`

This is correct. MSG_SSR_GRIDDED_CORRECTION_BOUNDS (1534) is an extension of MSG_SSR_GRIDDED_CORRECTION (1532). So that one message contains the corrections + bounds, instead of only bounds (like the other integrity messages).

mfine:
> This is correct. MSG_SSR_GRIDDED_CORRECTION_BOUNDS (1534) is an extension of MSG_SSR_GRIDDED_CORRECTION (1532). So that one message contains the corrections + bounds, instead of only bounds (like the other integrity messages).

Nice. Then disregard this PR.

PR #1198, opened 2022-07-26
Added sbp2json test

adrian-kong:
# Description

Adds `fn test_sbp2json_*` test for JsonEncoder (used in sbp2json) for serializing SBP messages into json strings.
Tests for `sbp::json::CompactFormatter`
This test also assumes the original `fn test_*` passes, which does not use `json` feature.

As per last pull request: https://github.com/swift-nav/libsbp/pull/1196
We added some more tests for `json2sbp` which tests deserializing of raw json (using methods used in sbp2json) and checks the sbp fields.

Basically TLDR:
`fn test_json2sbp` - tests from-fields is identical to sbp created via payload && contains all the fields
`fn test_sbp2json` - tests sbp constructed from payload serialized into json is identical (basically roundtrip, json2sbp from-fields)

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-821

### DIFF
--- a/test_data/benchmark_main.py
+++ b/test_data/benchmark_main.py
@@ -11,8 +11,8 @@ SLUSH_PERCENTAGE = 0.25
 
 # How much faster Rust should be than other implementations
 RATIOS_SBP2JSON = {
-    "haskell": 1.51,
-    "python": 14.10,
+    "haskell": 2.12,
+    "python": 19.46,
 }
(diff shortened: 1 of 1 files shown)


silverjam (approved):
Lgtm

notoriaga (approved):

PR #1199, opened 2022-07-28
Automatically create alias targets in cmake - Submodule update [ESD-2467]

lucasle-sn:
# Jira ticket 
https://swift-nav.atlassian.net/browse/ESD-2467

# Design
* The main content for auto ALIAS creation is in `cmake`. This PR only removes manual ALIAS addition for sbp.

# Related PRs


isaactorz (commented):
This just a test PR?

Looks good regardless. I think we should be ok regardless of which of these changes lands first as they peculate up the submodule tree (cmake or the one here).

lucasle-sn:
> This just a test PR?
> 
> Looks good regardless. I think we should be ok regardless of which of these changes lands first as they percolate up the submodule tree (cmake or the one here).

Hi @isaactorz, the new update in `cmake` will automatically create alias for all swift targets, under naming convention `swiftnav::${target}`. So the line `add_library(swiftnav::sbp ALIAS sbp)` becomes redundant. But it will not cause any issue.

silverjam (commented):
@lucasle-sn removing from my review queue, retag if necessary

antrikshsrivastava (approved):

RReichert (approved):

PR #1203, opened 2022-08-03
Hooking package-register updater

adrian-kong:
Calls to create PR in package registry upon release

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-852

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -82,13 +82,6 @@ jobs:
         id: build
         run: ./scripts/ci_build_rust.bash
         shell: bash
-        
-      - name: Debugging
-        shell: bash
-        run: |
-          find .
-          echo "====ls"
-          ls .
 
       - name: Upload Release
 #        if: startsWith(github.ref, 'refs/tags/v')
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -88,7 +88,7 @@ jobs:
         uses: svenstaro/upload-release-action@v1-release
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


silverjam (approved):

.github/workflows/rust.yaml
```diff
@@ -83,28 +83,44 @@ jobs:
...
+  registry:
+    if: startsWith(github.ref, 'refs/tags/v')
+    needs: [ build ]
+    name: Publish rust release to package-registry
+    runs-on: ubuntu-latest
+    steps:
+      - name: Hook into package registry
+        uses: swift-nav/swift-cli-publisher@v2
+        env:
+          DL_LINUX: ${{ needs.build.outputs.LINUX_RELEASE_ARCHIVE }}
```
adrian-kong, line 117: Seems like we can't access the output of previous jobs if it is dynamically modified (as per https://stackoverflow.com/questions/59175332/using-output-from-a-previous-job-in-a-new-one-in-a-github-action as well)

silverjam: Per discussion we probably can share data by using well known artifact names then uploading/downloading in different jobs, please re-request a review once you have that working

adrian-kong: Just switched them both to enforce binary targets and hardcoded in the paths

silverjam:
Looks fine, let's do a test release to validate

adrian-kong:
https://github.com/swift-nav/libsbp/runs/7662941672?check_suite_focus=true

https://github.com/swift-nav/package-registry/pull/35

looks to release ok and creates PR,
version is incorrect though

silverjam (approved):

PR #1124, opened 2022-04-26
Update sbp Docs to Include L5 Signals

isaactorz:
# Description

The gnss signals that are enumerated by `GnssSignal` are incomplete, and thus the table that gets rendered for them in our pdf doc is incomplete.

This change adds a new section to the documentation that enumerates every gnss signal. This table is hardcoded into our latex files.

The messages that contain the fields of `GnssSignal` will no longer create the table for the bitfields. Instead we a page reference is appended to the description of this field that points readers to the table where the full list of signals is defined.

The new signals table is currently on Page 4, and you can see an example on ~ pg 43 of the field's description referencing this new table.

@swift-nav/devinfra

# API compatibility

No API compatibility implications for this change

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-661

PR #1204, opened 2022-08-05
Update SBP Docs to Include GNSS Signals Table [DEVINFRA-661] (#1124)

isaactorz:
Adds a table to the libsbp pdf that explicitly lists all the GNSS Signals
we support (or intend to support).

The message fields that referenced the old, incomplete table (derived from
a bitfield), have been updated to have a link to the new table.

Cherry-picked from: 7763b73f2aa6891a57dd47656677bb6bfe2668f5 per request from @larrychangSN 

# Description

@swift-nav/devinfra

# API compatibility

No this only changes the documentation.

## API compatibility plan

# JIRA Reference

No jira ticket.


isaactorz:
@silverjam Any idea what this build error is ^^ ? `make html-python` ran successfully for me in the build container.

silverjam:
A lot of things are not containerized in GHA, so we occasionally get hit with build breakages from non-backwards compatible tool updates... in this case I think the build error is this:

```
Warning, treated as error:
Invalid configuration value found: 'language = None'. Update your configuration to a valid language code. Falling back to 'en' (English).
Creating file ./sbp.rst.
Creating file ./sbp.acquisition.rst.
Creating file ./sbp.bootload.rst.
```

silverjam:
In this branch we use `langauge = None` see:
https://github.com/swift-nav/libsbp/blob/libsbp-v3.4.11-release/python/docs/source/conf.py#L73

In master we use `language = "en"` see:
https://github.com/swift-nav/libsbp/blob/master/python/docs/source/conf.py#L81

silverjam:
just close to close this for now since CI needs some work

PR #1205, opened 2022-08-09
Update deprecating ubuntu 18.04

adrian-kong:
# Description

For deprecating ubuntu 18.04 not needed for like another 4 months
https://github.com/actions/runner-images/issues/6002


silverjam:
> # Description
> For deprecating ubuntu 18.04 not needed for like another 4 months [actions/runner-images#6002](https://github.com/actions/runner-images/issues/6002)

Ooops, hit the wrong button, didn't mean to close the PR... for gcc-6, we're going to have to switch to docker for that step, or just drop support for gcc-6 here (it'll be implicitly checked elsewhere).

For example: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-docker-hub-action

.github/workflows/c.yaml
```diff
@@ -52,18 +52,21 @@ jobs:
           { build_type: "Debug", compiler: { c: "gcc-6", cxx: "g++-6" }, build_shared_libraries: false }
         ]
     name: "Ubuntu 18.04 (Build: ${{ matrix.build_type }}, Compilers: ${{ matrix.compiler.c }}/${{ matrix.compiler.cxx }}, Shared Library: ${{ matrix.build_shared_libraries }})"
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
+    container: ubuntu:18.04
```
silverjam, line 56: Maybe we could use our existing libsbp-build container for this?

adrian-kong: Think we just need to upgrade cmake to 3.2x, having some issues with the keys from downloading via kitware

adrian-kong:
Superseded by #1250

PR #1216, opened 2022-09-01
SbpString helpers

adrian-kong:
# Description

Adds string helpers for parsing sbp string to vec

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-867


rust/sbp/src/sbp_string.rs
```diff
@@ -37,6 +39,30 @@ where
...
 }
 
+impl<T: AsRef<[u8]>> SbpString<T, Multipart> {
+    pub fn parse(&self) -> Result<Vec<std::borrow::Cow<str>>, MultipartError> {
+        parse_to_vec(self.as_bytes(), 1)
+    }
+}
+
+impl<T: AsRef<[u8]>> SbpString<T, NullTerminated> {
+    pub fn parse(&self) -> Result<std::borrow::Cow<str>, MultipartError> {
```
notoriaga, line 49: I think this should be 


```suggestion
    pub fn parts(&self) -> impl Iterator<Item=&[u8]> {
```

It's more flexible (doesn't allocate and doesn't parse as a string) but you can replicate the current behavior with it. Also I don't think Result should be involved here, you should assume that the string is a valid multipart string (and then provide constructors for multipart string that makes sure they are valid). I was thinking of adding a constructor for each encoding on SbpString so like SbpString::multipart which would be fallible

notoriaga (approved):

notoriaga:
Sorry didn't mean to approve, that's what I get for reviewing from my phone

rust/sbp/src/sbp_string.rs
adrian-kong: Something like:

```rust
impl<T: AsRef<[u8]>, E> SbpString<T, E> {
    pub fn parts(&self) -> impl Iterator<Item = &[u8]> {
        self.as_bytes().split(|b| b == &0)
    }
}
```

and have 

```rust
pub struct PartsDecoder<T>(T);

impl PartsDecoder<DoubleNullTerminated>{
    
    fn parse(parts: impl Iterator<Item = &[u8]>) -> Result<Vec<Cow<str>>, MultipartError>{
        parts_to_vec(parts, 2)
    }
    
}
```
and have `parts_to_vec` as a helper method to parse iterator according to the nulls, return an error if the number of nulls dont match?

notoriaga: I was thinking the API could look something like 

```rust
/// Fixed or variable length string and its encoding.
#[derive(Debug, PartialEq, Clone)]
pub struct SbpString<T, E = Unterminated> {
    data: T,
    encoding: PhantomData<E>,
}

impl<T, E> SbpString<T, E>
where
    T: AsRef<[u8]>,
{
    /// Creates a new SbpString with a given encoding. Does not check if the data is valid
    /// for the given encoding.
    pub fn new_unchecked(data: T) -> Self { // rename new to reflect that it doesn't check encodings
        Self {
            data,
            encoding: PhantomData,
        }
    }

    /// Returns a byte slice of this SbpString's contents.
    pub fn as_bytes(&self) -> &[u8] {
        self.data.as_ref()
    }

    /// Returns a byte vector of this SbpString's contents.
    pub fn to_vec(&self) -> Vec<u8> {
        self.data.as_ref().to_vec()
    }
}

impl<T: AsRef<[u8]>> SbpString<T, Unterminated> {
    // I think this one can be infallible
    pub fn unterminated(data: T) -> Self {
        todo!();
    }
}

impl<T: AsRef<[u8]>> SbpString<T, NullTerminated> {
     // maybe doesn't need to return a result we could also just append the null byte if missing
    pub fn null_terminated(data: T) -> Result<Self, NullTerminatorError> {
        todo!();
    }
}

// from calling .split() on a slice/vec
// we should reuse this iterator because it impls a lot of nice things
// like DoubleEndedIterator. And its a std type so it's fine to have it be in the public API
// We could also wrap it up in a newtype if you want
pub type Parts<'a> = std::slice::Split<'a, u8, fn(&u8) -> bool>;

// Multipart and DoubleNullTerminated all variable length so we don't need to
// be generic over the container we can just assume Vec<u8>

impl SbpString<Vec<u8>, Multipart> {
    pub fn multipart(data: impl Into<Vec<u8>>) -> Result<Self, MultipartError> {
        todo!();
    }

    pub fn from_parts(parts: impl IntoIterator<Item = impl AsRef<[u8]>>) -> Self {
        todo!();
    }

    pub fn parts(&self) -> Parts<'_> {
        todo!();
    }
}

impl SbpString<Vec<u8>, DoubleNullTerminated> {
    pub fn double_null_terminated(data: impl Into<Vec<u8>>) -> Result<Self, DoubleNullTerminatedError> {
        todo!();
    }

    pub fn from_parts(parts: impl IntoIterator<Item = impl AsRef<[u8]>>) -> Self {
        todo!();
    }

    // cant share an impl with Multipart bc you need to handle the extra null at the end
    pub fn parts(&self) -> Parts<'_> {
        todo!();
    }
}
```

The one thing I don't like about this API is that if you wanna build a new string with `from_parts` I think you'll still need to specify the encoding.

```rust
// is this multipart or double null terminated?
let x = SbpString::from_parts(["a", "b", "c"]);
// will have to do something like
let x = SbpString::<_, Multipart>::from_parts(["a", "b", "c"]);
```

Not sure if there is a nice way around that, other than by naming the functions differently. Also something to consider is adding `FromIterator` impls for the multipart strings

```rust
impl<T: AsRef<[u8]>> FromIterator<T> for SbpString<Vec<u8>, Multipart> {
    fn from_iter<I: IntoIterator<Item=T>>(iter: I) -> Self {
        todo!();
    }
}
```

which would let people use `.collect` to create multipart strings

```rust
let parts = ["a", "b", "c"];
let string: SbpString<_, Multipart> = parts.into_iter().collect();
```

Could just do that instead of `from_parts` (or could do both and impl one via the other)

> and have parts_to_vec as a helper method to parse iterator according to the nulls, return an error if the number of nulls dont match?

Creating the string from the parts should be something like 

```rust
fn from_parts(it: ...) -> Self {
    // check that none of the parts contain a null byte already
    // and return an error (I think a case could also be made to panic instead, or we could have the code
    // remove the null bytes) 
    // once you've checked it just
    Self {
        data: it.intersperse(b"0").collect(),
        encoding: PhantomData,
    }
}
```

silverjam (changes requested):
Let's see if we can implement @notoriaga 's suggestions before merging

rust/sbp/src/sbp_string.rs
```diff
@@ -37,6 +42,98 @@ where
...
+        }
+    }
+}
+
+// pub type Parts<'a> = std::slice::Split<'a, u8, fn(&u8) -> bool>;
+
+impl SbpString<Vec<u8>, Multipart> {
+    pub fn multipart(data: impl Into<Vec<u8>>) -> Result<Self, MultipartError> {
+        let vec = data.into();
+        if let [.., one] = vec.as_slice() {
```
adrian-kong, line 81: is this basically the idea?

- we don't actually check if the next is non null so a multipart could be a double null... - should we handle that case as well or assume it should be multipart

rust/sbp/src/sbp_string.rs
```diff
@@ -37,6 +39,100 @@ where
...
+    for i in parts {
+        for j in i.as_ref() {
+            data.push(*j)
+        }
+        data.push(0);
+    }
+    data
+}
+
+// pub type Parts<'a> = std::slice::Split<'a, u8, fn(&u8) -> bool>;
```
adrian-kong, line 82: @notoriaga wasn't sure how to use this type...

maybe we could store the parts in SbpString i.e...?

```rust
#[derive(Debug, PartialEq, Clone)]
pub struct SbpString<T, E = Unterminated> {
    data: T,
    encoding: PhantomData<E>,
    parts: Vec<Cow<str>>
}
```

rust/sbp/src/sbp_string.rs
```diff
@@ -37,6 +38,93 @@ where
...
+///
+/// Ex. [a, b, c] => [a, 0, b, 0, c, 0]
+///
+/// Returns null terminated vector
+fn alt_null<R, I>(parts: I) -> Vec<u8>
+where
+    R: AsRef<[u8]>,
+    I: IntoIterator<Item = R>,
+{
+    let mut data = vec![];
```
joelynch, line 71: I'd be tempted to write some iterator nonsense here but this seems really readable actually 👌

joelynch (approved):
Looks good to me Adrian

rust/sbp/src/sbp_string.rs
joelynch:
```rust
    parts
        .into_iter()
        .flat_map(|r| {
            r.as_ref()
                .iter()
                .flat_map(|i| vec![*i, 0])
                .collect::<Vec<_>>()
        })
        .collect()
```

there "fixed" 😆

adrian-kong: would have loved `Iterator::intersperse` but it is nightly :(

silverjam (approved):
Looks great, is the intent to utilize these new helpers in sbp-settings?  Should we create follow-on tasks for that?

rust/sbp/src/sbp_string.rs
```diff
@@ -37,6 +38,94 @@ where
     }
 }
 
+impl<T: AsRef<[u8]>> SbpString<T, Unterminated> {
+    /// Checked unterminated SbpString builder, verify last byte is non null
+    pub fn unterminated(data: T) -> Result<Self, UnterminatedError> {
+        match data.as_ref().last() {
+            Some(l) if l != &0 => Ok(Self::new(data)),
```
notoriaga, line 45: I think an unterminated string ending in a null byte is actually okay? @silverjam do you know?

rust/sbp/src/sbp_string.rs
```diff
@@ -218,6 +307,15 @@ macro_rules! forward_payload_vec {
     };
 }
 
+#[derive(Debug, PartialEq)]
+pub struct UnterminatedError;
```
notoriaga, line 311: These should all impl Error (and Display). Otherwise they won't compose with things like Box<dyn Error> or anyhow::Error

notoriaga (changes requested):

rust/sbp/src/sbp_string.rs
silverjam: I don't think we enforce this anywhere in firmware so this really depends on intended usage-- if we expect to use this construct unterminated SbpStrings from a remote device then I think this will fail because some areas where a string is expected to be "unterminated" we'll end up inserting a null terminated one just because that's the convention in C.

rust/sbp/src/sbp_string.rs
```diff
@@ -37,6 +38,94 @@ where
...
+    parts
+        .into_iter()
+        .flat_map(|r| {
+            r.as_ref()
+                .iter()
+                .flat_map(|i| vec![*i, 0])
+                .collect::<Vec<_>>()
+        })
+        .collect()
+}
```
notoriaga, line 80: How about 

```suggestion
fn alt_null<R, I>(parts: I) -> Vec<u8>
where
    R: AsRef<[u8]>,
    I: IntoIterator<Item = R>,
{
    parts.into_iter().fold(Vec::new(), |mut acc, part| {
        acc.extend_from_slice(part.as_ref());
        acc.push(0);
        acc
    })
}
```

which avoids the intermediate vec

rust/sbp/src/sbp_string.rs
```diff
@@ -37,6 +38,94 @@ where
...
+                return Ok(SbpString::new(vec));
+            }
+        };
+        Err(DoubleNullTerminatedError)
+    }
+
+    pub fn from_parts(parts: impl IntoIterator<Item = impl AsRef<[u8]>>) -> Self {
+        let mut alt = alt_null(parts);
+        alt.push(0);
+        SbpString::double_null_terminated(alt).unwrap()
```
notoriaga, line 120: I think `SbpString::new(alt)` would also work here

notoriaga: / the other from_parts as well

rust/sbp/src/sbp_string.rs
notoriaga: Also should derive `Clone` or else things like `Result<SbpString, UnterminatedError>` won't be cloneable, which can sometimes be annoying

rust/sbp/src/sbp_string.rs
```diff
@@ -37,6 +38,94 @@ where
...
+}
+
+impl<T: AsRef<[u8]>> SbpString<T, NullTerminated> {
+    pub fn null_terminated(data: T) -> Result<Self, NullTerminatedError> {
+        match data.as_ref().last() {
+            Some(l) if l == &0 => Ok(Self::new(data)),
+            _ => Err(NullTerminatedError),
+        }
+    }
+}
```
notoriaga, line 58: What do you think about having separate constructors for Vec<u8> and [u8; LEN] strings? In the case of the Vec I think it might be nice to have the constructor be infallible. It's a variable length string so we can just add a null byte. And for the fixed strings we could still return an error (in addition to now checking the length of the string (at compile time)). Or we could make everything infallible by padding short strings + setting the last byte to 0  

```rust
impl SbpString<Vec<u8>, NullTerminated> {
    pub fn null_terminated(data: impl Into<Vec<u8>>) -> Self {
        let mut data = data.into();
        if data.last().map_or(true, |n| n != &0) {
            data.push(0);
        }
        Self::new(data)
    }
}

impl<const LEN: usize> SbpString<[u8; LEN], NullTerminated> {
    // accepting an array instead of AsRef<[u8]> checks the length at compile time
    pub fn null_terminated(data: [u8; LEN]) -> Result<Self, NullTerminatedError> {
        match data.last() {
            Some(l) if l == &0 => Ok(Self::new(data)),
            _ => Err(NullTerminatedError),
        }
    }
}
// or if we want an infallible api
impl<const LEN: usize> SbpString<[u8; LEN], NullTerminated> {
    pub fn null_terminated(data: impl AsRef<[u8]>) -> Self {
        let mut s: [u8; LEN] = [0; LEN];
        s.copy_from_slice(data.as_ref());
        s[LEN - 1] = 0;
        Self::new(s)
    }
}
```

what do you think, too magic?

rust/sbp/src/sbp_string.rs
adrian-kong: hmmm should we skip unterminated then 🤔

rust/sbp/src/sbp_string.rs
adrian-kong: hmmm aren't both of these the same?
i.e. `[u8; LEN]` is also just `impl Into<Vec<u8>>`

running
`SbpString::null_terminated([0x1, 0x1]);` would use both implementation (as before test case worked for into vec, but now with [u8;2] as well)

rust/sbp/src/sbp_string.rs
```diff
@@ -37,6 +39,110 @@ where
...
+//     pub fn from_parts(parts: [u8; LEN]) -> Self {
+//         SbpString::new(alt_null(parts))
+//     }
+//
+//     pub fn parts(&self) -> Parts<'_> {
+//         let slice = self.data.as_slice();
+//         slice[0..slice.len() - 1].split(|a| a == &0)
+//     }
+// }
+
```
notoriaga, line 120:
```suggestion
```

rust/sbp/src/sbp_string.rs
```diff
@@ -218,6 +324,47 @@ macro_rules! forward_payload_vec {
...
+impl std::error::Error for NullTerminatedError {}
+
+impl std::fmt::Display for DoubleNullTerminatedError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        writeln!(f, "failed double null terminated string validation")
+    }
+}
+
+impl std::error::Error for DoubleNullTerminatedError {}
```
notoriaga, line 366: Could you group these impls with the struct definitions?

rust/sbp/src/sbp_string.rs
```diff
@@ -37,6 +39,110 @@ where
...
+
+    /// Unchecked from parts builder to construct Multipart SbpString
+    pub fn from_parts(parts: impl IntoIterator<Item = impl AsRef<[u8]>>) -> Self {
+        SbpString::new(alt_null(parts))
+    }
+
+    pub fn parts(&self) -> Parts<'_> {
+        let slice = self.data.as_slice();
+        slice[0..slice.len() - 1].split(|a| a == &0)
+    }
```
notoriaga, line 98:
```suggestion
    /// Returns an iterator over the parts of the string.
    pub fn parts(&self) -> Parts<'_> {
        let slice = self.data.as_slice();
        slice[0..slice.len() - 1].split(|a| a == &0)
    }
```

rust/sbp/src/sbp_string.rs
```diff
@@ -37,6 +39,110 @@ where
...
 
+impl<T: AsRef<[u8]>> SbpString<T, Unterminated> {
+    /// Checked unterminated SbpString builder, verify last byte is non null
+    pub fn unterminated(data: T) -> Result<Self, UnterminatedError> {
+        match data.as_ref().last() {
+            Some(l) if l != &0 => Ok(Self::new(data)),
+            _ => Err(UnterminatedError),
+        }
+    }
+}
```
notoriaga, line 50: Seems like we can drop the check here. I'd keep the function for consistency though

```suggestion
impl<T: AsRef<[u8]>> SbpString<T, Unterminated> {
    /// Constructs a new unterminated string
    pub fn unterminated(data: T) -> Self {
        SbpString::new(data)
    }
}
```

notoriaga (commented):
Sorry this slipped my mind, just a few small things

notoriaga (approved):

PR #1218, opened 2022-09-05
Sbp Raw Decoder

adrian-kong:
# Description

Raw decoder to return iterator of bytes mut instead of consuming and skipping errors

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-866


rust/sbp/src/de.rs
```diff
@@ -174,6 +176,114 @@ impl std::fmt::Display for CrcError {
...
+        todo!()
+    }
+}
+
+pub struct SbpFrame<B> {
+    inner: B,
+    payload_len: usize,
+}
+
+impl SbpFrame<BytesMut> {
```
notoriaga, line 234: these methods should be generic over the buffer type, something like 

```rust
pub struct Frame<B>(B);

impl<B: bytes::Buf> Frame<B> {
    pub fn preamble(&self) -> u8 {
        self.0.chunk().get_u8()
    }

    pub fn msg_type(&self) -> u16 {
        let mut slice = &self.0.chunk()[1..];
        slice.get_u16_le()
    }

    pub fn sender_id(&self) -> u16 {
        let mut slice = &self.0.chunk()[3..];
        slice.get_u16_le()
    }

    pub fn payload_len(&self) -> usize {
        let mut slice = &self.0.chunk()[5..];
        slice.get_u8().into()
    }

    pub fn payload(&self) -> &[u8] {
        &self.0.chunk()[HEADER_LEN..HEADER_LEN + self.payload_len()]
    }

    pub fn crc(&self) -> u16 {
        let mut slice = &self.0.chunk()[HEADER_LEN + self.payload_len()..];
        slice.get_u16_le()
    }

    pub fn check_crc(&self) -> Result<u16, CrcError> {
        let actual = self.crc();
        let data = &self.0.chunk()[1..HEADER_LEN + self.payload_len()];
        if actual == crc16::State::<crc16::XMODEM>::calculate(data) {
            Ok(actual)
        } else {
            Err(CrcError {
                msg_type: self.msg_type(),
                sender_id: self.sender_id(),
                crc: actual,
            })
        }
    }

    pub fn len(&self) -> usize {
        self.0.chunk().len()
    }
}
```

rust/sbp/src/de.rs
```diff
@@ -174,6 +176,114 @@ impl std::fmt::Display for CrcError {
...
+    pub fn new(inner: BytesMut, payload_len: usize) -> Self {
```
notoriaga, line 235: we shouldn't provide a public constructor for `Frame`. you should only be able to get them either from a `Framer` or from an existing message

rust/sbp/src/de.rs
```diff
@@ -174,6 +178,153 @@ impl std::fmt::Display for CrcError {
...
+    }
+
+    pub fn len(&self) -> usize {
+        self.0.chunk().len()
+    }
+}
+
+impl SbpFrame<BytesMut> {
+    pub fn parse(buf: &mut BytesMut) -> Option<Self> {
+        if let Some(index) = buf.iter().position(|b| b == &PREAMBLE) {
```
adrian-kong, line 304: Do we want to maybe wrap this in SbpFrame or just leave these skips as they are and return None?

rust/sbp/src/messages/mod.rs
```diff
@@ -1508,6 +1509,1122 @@ impl<'de> serde::Deserialize<'de> for Sbp {
 }
 
 impl Sbp {
+
+    pub fn from_sbp_frame<B: bytes::Buf>(frame: crate::de::SbpFrame<B>) ->Result<Sbp, PayloadParseError> {
```
adrian-kong, line 1513: Example of `from_frame`, replacement...

rust/sbp/src/de.rs
```diff
@@ -174,6 +178,153 @@ impl std::fmt::Display for CrcError {
...
+    fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        Ok(SbpFrame::<BytesMut>::parse(src))
+    }
+    fn decode_eof(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        todo!()
+    }
+}
+
+#[derive(Debug)]
+pub struct SbpFrame<B>(B);
```
notoriaga, line 257: this should just be `Frame`, I'd try to avoid prefixing things with `Sbp`, it's implied by being inside the `sbp` crate. `SbpFramer` has the prefix just to avoid a name collision, but it's private. `FramerImpl` might be a better name

notoriaga: in other words I think we should replace the current `Frame` struct in this module with the new one. Usage of the current `Frame` struct is pretty low I think. and anything using it will probably want to utilize the `Framer` so they'll need to be updated anyway

rust/sbp/src/de.rs
notoriaga: Actually I think we can skip on the parse functions for now and just do everything in the decoder

```rust
impl Decoder for SbpFramer {
    fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
        let start = match src.iter().position(|b| b == &PREAMBLE) {
            Some(idx) => idx,
            None => {
                src.clear();
                return Ok(None);
            }
        };
        src.advance(start);
        if src.len() < HEADER_LEN {
            src.reserve(MAX_FRAME_LEN);
            return Ok(None);
        }
        let payload_len = src[5] as usize;
        let at = HEADER_LEN + payload_len + CRC_LEN;
        if src.len() < at {
            src.reserve(MAX_FRAME_LEN);
            return Ok(None);
        }
        let frame = SbpFrame(src.split_to(at));
        frame.check_crc()?;
        Ok(Some(frame))
    }
}
```

simplifies things not having to worry about parsing from `&[u8]` and we can add it later if we want it

rust/sbp/src/de.rs
```diff
@@ -294,49 +330,47 @@ mod tests {
...
-        let res = parse_frame(&mut buf).unwrap().unwrap();
-        assert_eq!(res.msg_type, 523); // 0x020B
-        assert_eq!(res.sender_id, 35027); // 0x88D3
-        assert_eq!(
-            res.payload,
-            BytesMut::from(&packet[HEADER_LEN..packet.len() - 1])
-        );
-        assert!(buf.is_empty());
-    }
+    // /// Test parsing when we don't have enough data for a frame message
```
adrian-kong, line 333: keeping this commented for now until the structure has been figured out

rust/sbp/src/de.rs
```diff
@@ -294,49 +330,47 @@ mod tests {
...
 
     #[test]
     fn test_parse_nothing() {
         let data = vec![0u8; 1000];
-        let mut bytes = BytesMut::from(&data[..]);
+        let bytes = BytesMut::from(&data[..]);
         assert_eq!(bytes.len(), 1000);
-        assert!(matches!(SbpDecoder.decode(&mut bytes), Ok(None)));
-        assert!(bytes.is_empty());
+        assert!(matches!(SbpDecode::new(bytes.reader()).next(), None));
```
adrian-kong, line 364: so now we dont mutate the bytesmut... is that a good idea ?

rust/sbp/src/json/de.rs
```diff
@@ -62,11 +62,7 @@ impl JsonDecoder {
         let data = input.into_inner();
         self.payload_buf.clear();
         base64::decode_config_buf(data.payload, base64::STANDARD, &mut self.payload_buf)?;
-        let msg = Sbp::from_frame(Frame {
-            msg_type: data.msg_type,
-            sender_id: data.sender,
-            payload: BytesMut::from(&self.payload_buf[..]),
-        })?;
+        let msg = Sbp::from_field( data.msg_type,data.sender, BytesMut::from(&self.payload_buf[..]), )?;
```
adrian-kong, line 65: refactored into mid, sid, payload

rust/sbp/src/de.rs
```diff
@@ -196,36 +189,80 @@ impl Decoder for SbpDecoder {
...
-            Some(Err(err)) => Err(err.into()),
-            None => {
-                src.reserve(MAX_FRAME_LEN);
-                Ok(None)
-            }
+        if src.len() < HEADER_LEN {
+            src.reserve(MAX_FRAME_LEN);
+            return Ok(None);
         }
+        let payload_len = src[5] as usize;
```
adrian-kong, line 196: maybe should extract 5 as a constant but would be kind of weird just having like `PAYLOAD_POSITION = 5` rather than being constructed via other values. HEADER_LEN - 1 ?

rust/sbp/src/messages/mod.rs
```diff
@@ -1508,1141 +1509,1120 @@ impl<'de> serde::Deserialize<'de> for Sbp {
...
-    ///         payload,
-    ///     };
-    ///     let msg: MsgLog = Sbp::from_frame(frame)?.try_into()?;
-    ///     assert_eq!(msg.sender_id, Some(1));
-    ///     assert_eq!(msg.level, 1);
-    ///     assert_eq!(msg.text.as_bytes(), "hello".as_bytes());
-    ///     Ok(())
-    /// }
-    /// ```
-    pub fn from_frame<B: Buf>(mut frame: crate::Frame<B>) -> Result<Sbp, PayloadParseError> {
```
notoriaga, line 1536: I think if you just made this

```rust
pub fn from_frame<B: Buf>(mut frame: &crate::Frame<B>) -> Result<Sbp, PayloadParseError>;
```

it would work. You wanna avoid the `BytesMut::from(&payload[..])` you have - that copies each message

rust/sbp/src/de.rs
```diff
@@ -196,36 +189,80 @@ impl Decoder for SbpDecoder {
...
+    fn decode_eof(&mut self, _src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        todo!()
+    }
+}
+
+/// The wire representation of an SBP message.
+#[derive(Debug)]
+pub struct SbpFrame<B>(B);
+
+impl<B: bytes::Buf> SbpFrame<B> {
```
notoriaga, line 215: Might be nice to add a `pub fn as_bytes(&self) -> &[u8] { self.0.chunk() }` - would clean up these other helpers and might be nice for an end user

rust/sbp/src/de.rs
notoriaga: I don't this it matters

rust/sbp/src/json/ser.rs
```diff
@@ -194,11 +193,7 @@ impl<F: Formatter + Clone> Encoder<Json2JsonInput> for Json2JsonEncoderInner<F>
     fn encode(&mut self, input: Json2JsonInput, dst: &mut BytesMut) -> Result<(), Self::Error> {
         let formatter = self.formatter.clone();
         let payload = base64::decode(input.data.payload)?;
-        let msg = Sbp::from_frame(Frame {
-            msg_type: input.data.msg_type,
-            sender_id: input.data.sender,
-            payload: BytesMut::from(&payload[..]),
-        })?;
+        let msg = Sbp::from_field(input.data.msg_type, input.data.sender, BytesMut::from(&payload[..]))?;
```
notoriaga, line 196: something like this might help for that other change I mentioned

```suggestion
        let mut frame = BytesMut::new();
       // push header + payload
       let frame = crate::Frame(frame);
       let msg = Sbp::from_frame(&frame)?;
```

rust/sbp/src/messages/mod.rs
notoriaga: It actually makes more sense for the frame to be behind an `&` - we have to copy the data out to fill the struct fields anyway. If we were to do something like cast the frame to a specific message than taking ownership would make more sense

rust/sbp/src/de.rs
notoriaga: Both `HEADER_LEN - 1` and a separate thing sound good to me

rust/sbp/src/de.rs
notoriaga: Hmm or maybe `AsRef<[u8]>`

notoriaga:
@adrian-kong I'm almost done with a pr to add on to what you have here which might help clean things up, so for right now I'd focus on the ubx stuff

rust/sbp/src/messages/mod.rs
adrian-kong: but aren't we removing the old Frame with the new one containing the buf

rust/sbp/src/de.rs
```diff
@@ -238,35 +292,34 @@ impl TimeoutSbpDecoder {
...
     type Error = Error;
 
     fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
         if self.last_msg_received.elapsed() > self.timeout_duration {
             return Err(std::io::Error::new(std::io::ErrorKind::TimedOut, "timeout").into());
         }
-        let res = self.decoder.decode(src);
-        if let Ok(Some(_)) = res {
+        let next = match FramerImpl.decode(src)? {
+            Some(frame) => frame.to_sbp().map(Some),
```
adrian-kong, line 304: map some.... ? got confused with all the naming not even sure if this is ok but it def doesn't look right, mapping everything to some

rust/sbp/src/de.rs
```diff
@@ -174,17 +126,64 @@ impl std::fmt::Display for CrcError {
 
 impl std::error::Error for CrcError {}
 
-#[derive(Debug, Clone, Default)]
-struct SbpDecoder;
+struct Framer<R>(FramedRead<R, FramerImpl>);
```
notoriaga, line 129: pub struct here (also you should `pub use` it from lib.rs (same w decoder))

rust/sbp/src/de.rs
```diff
@@ -196,36 +195,91 @@ impl Decoder for SbpDecoder {
...
+    pub fn to_sbp(&self) -> Result<Sbp, Error> {
+        self.check_crc()?;
+        Ok(Sbp::from_field(
+            self.msg_type(),
+            self.sender_id(),
+            self.payload(),
+        )?)
+    }
+
+    pub fn to_bytes(&self) -> Box<dyn AsRef<[u8]> + '_> {
```
notoriaga, line 270:
```
pub fn as_bytes(&self) -> &[u8];
```
`as_` is more common if the conversion is cheap (this is) and you shouldn't need Box dyn here, that more for if you want to return more than one type dynamically

rust/sbp/src/de.rs
notoriaga: that's pretty normal/idiomatic way to go from `Result<T>` to `Result<Option<T>>`. `Some`/enum variants are just [functions](https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/enums.html#constructors-as-functions) so mapping them is normal

adrian-kong: that is super weird, thought we could bubble up from the frame then instead of doin the ? to unwrap the FramerImpl decode making us manaully cast it to some

rust/sbp/src/de.rs
adrian-kong: says
```
the trait bound `dyn AsRef<[u8]>: std::marker::Sized` is not satisfied [E0277]
```
with AsRef[u8]
instead of Box<dyn AsRef[u8]>

rust/sbp/src/de.rs
adrian-kong: do we want decoder to be public? its just constructed in iter_messages()

rust/sbp/src/de.rs
adrian-kong: nvm fixed using &[u8]

rust/sbp/src/de.rs
notoriaga: I'd like to deprecate `iter_messages`, in hindsight I think exposing a named type is better than the `-> impl Iterator`. You can't store `impl Iterator` in a field of a struct without `Box`. We could also add more helper methods to `Decoder`. And it's not any less convenient to use (it's shorter actually)
```
for m in sbp::Decoder::new(reader) {}
vs
for m in sbp::iter_messages(reader) {}
```

notoriaga:
Suggested edit:
```diff
diff --git a/rust/sbp/src/de.rs b/rust/sbp/src/de.rs
index a781bd57..2414bb81 100644
--- a/rust/sbp/src/de.rs
+++ b/rust/sbp/src/de.rs
@@ -128,7 +128,7 @@ impl std::error::Error for CrcError {}
 
 pub struct Framer<R>(FramedRead<R, FramerImpl>);
 
-impl<R: io::Read> Framer<R> {
+impl<R> Framer<R> {
     pub fn new(reader: R) -> Self {
         Self(FramedRead::new(reader, FramerImpl))
     }
@@ -142,6 +142,18 @@ impl<R: io::Read> Iterator for Framer<R> {
     }
 }
 
+#[cfg(feature = "async")]
+impl<R: futures::AsyncRead + Unpin> futures::Stream for Framer<R> {
+    type Item = Result<Frame, Error>;
+
+    fn poll_next(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        std::pin::Pin::new(&mut self.0).poll_next(cx)
+    }
+}
+
 pub struct Decoder<R>(FramedRead<R, FramerImpl>);
 
 impl<R> Decoder<R> {

```

notoriaga:
Suggested edit:
```diff
diff --git a/rust/sbp/src/de.rs b/rust/sbp/src/de.rs
index a781bd57..2414bb81 100644
--- a/rust/sbp/src/de.rs
+++ b/rust/sbp/src/de.rs
@@ -128,7 +128,7 @@ impl std::error::Error for CrcError {}
 
 pub struct Framer<R>(FramedRead<R, FramerImpl>);
 
-impl<R: io::Read> Framer<R> {
+impl<R> Framer<R> {
     pub fn new(reader: R) -> Self {
         Self(FramedRead::new(reader, FramerImpl))
     }
@@ -142,6 +142,18 @@ impl<R: io::Read> Iterator for Framer<R> {
     }
 }
 
+#[cfg(feature = "async")]
+impl<R: futures::AsyncRead + Unpin> futures::Stream for Framer<R> {
+    type Item = Result<Frame, Error>;
+
+    fn poll_next(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        std::pin::Pin::new(&mut self.0).poll_next(cx)
+    }
+}
+
 pub struct Decoder<R>(FramedRead<R, FramerImpl>);
 
 impl<R> Decoder<R> {

```

notoriaga:
Suggested edit:
```diff
diff --git a/rust/sbp/src/de.rs b/rust/sbp/src/de.rs
index a781bd57..3da68015 100644
--- a/rust/sbp/src/de.rs
+++ b/rust/sbp/src/de.rs
@@ -48,9 +48,9 @@ pub fn iter_messages<R: io::Read>(input: R) -> impl Iterator<Item = Result<Sbp,
 /// for the maximum allowed duration without a successful message.
 pub fn iter_messages_with_timeout<R: io::Read>(
     input: R,
-    timeout_duration: Duration,
+    timeout: Duration,
 ) -> impl Iterator<Item = Result<Sbp, Error>> {
-    TimeoutSbpDecoder::framed_with_timeout(input, timeout_duration)
+    TimeoutDecoder::new(input, timeout)
 }
 
 /// Deserialize the async IO stream into an stream of messages.
@@ -64,9 +64,9 @@ pub fn stream_messages<R: futures::AsyncRead + Unpin>(
 #[cfg(feature = "async")]
 pub fn stream_messages_with_timeout<R: futures::AsyncRead + Unpin>(
     input: R,
-    timeout_duration: Duration,
+    timeout: Duration,
 ) -> impl futures::Stream<Item = Result<Sbp, Error>> {
-    TimeoutSbpDecoder::framed_with_timeout(input, timeout_duration)
+    TimeoutDecoder::new(input, timeout)
 }
 
 /// All errors that can occur while reading messages.
@@ -179,6 +179,46 @@ impl<R: futures::AsyncRead + Unpin> futures::Stream for Decoder<R> {
     }
 }
 
+pub struct TimeoutDecoder<R>(FramedRead<R, TimeoutDecoderImpl<FramerImpl>>);
+
+impl<R> TimeoutDecoder<R> {
+    pub fn new(reader: R, timeout: Duration) -> Self {
+        TimeoutDecoder(FramedRead::new(
+            reader,
+            TimeoutDecoderImpl::new(timeout, FramerImpl),
+        ))
+    }
+}
+
+impl<R: io::Read> Iterator for TimeoutDecoder<R> {
+    type Item = Result<Sbp, Error>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let sbp = match self.0.next()? {
+            Ok(frame) => frame.to_sbp(),
+            Err(err) => return Some(Err(err)),
+        };
+        Some(sbp)
+    }
+}
+
+#[cfg(feature = "async")]
+impl<R: futures::AsyncRead + Unpin> futures::Stream for TimeoutDecoder<R> {
+    type Item = Result<Sbp, Error>;
+
+    fn poll_next(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        let frame = match futures::ready!(std::pin::Pin::new(&mut self.0).poll_next(cx)) {
+            Some(Ok(frame)) => frame,
+            Some(Err(e)) => return std::task::Poll::Ready(Some(Err(e.into()))),
+            None => return std::task::Poll::Ready(None),
+        };
+        std::task::Poll::Ready(Some(frame.to_sbp()))
+    }
+}
+
 #[derive(Debug, Clone, Default)]
 struct FramerImpl;
 
@@ -272,42 +312,35 @@ impl Frame {
     }
 }
 
-struct TimeoutSbpDecoder {
-    timeout_duration: Duration,
+struct TimeoutDecoderImpl<D> {
+    timeout: Duration,
     last_msg_received: Instant,
+    inner: D,
 }
 
-impl TimeoutSbpDecoder {
-    fn new(timeout_duration: Duration) -> Self {
+impl<D> TimeoutDecoderImpl<D> {
+    fn new(timeout: Duration, inner: D) -> Self {
         Self {
-            timeout_duration,
+            timeout,
             last_msg_received: Instant::now(),
+            inner,
         }
     }
-    fn framed_with_timeout<W>(
-        writer: W,
-        timeout_duration: Duration,
-    ) -> FramedRead<W, TimeoutSbpDecoder> {
-        FramedRead::new(writer, Self::new(timeout_duration))
-    }
 }
 
-impl dencode::Decoder for TimeoutSbpDecoder {
-    type Item = Sbp;
-    type Error = Error;
+impl<D: dencode::Decoder> dencode::Decoder for TimeoutDecoderImpl<D> {
+    type Item = D::Item;
+    type Error = D::Error;
 
     fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
-        if self.last_msg_received.elapsed() > self.timeout_duration {
-            return Err(std::io::Error::new(std::io::ErrorKind::TimedOut, "timeout").into());
+        if self.last_msg_received.elapsed() > self.timeout {
+            return Err(io::Error::new(io::ErrorKind::TimedOut, "timeout").into());
         }
-        let next = match FramerImpl.decode(src)? {
-            Some(frame) => frame.to_sbp().map(Some),
-            None => return Ok(None),
-        };
-        if let Ok(Some(_)) = next {
+        let next = self.inner.decode(src)?;
+        if next.is_some() {
             self.last_msg_received = Instant::now();
         }
-        next
+        Ok(next)
     }
 }
 

```

rust/sbp/src/lib.rs
```diff
@@ -186,11 +189,20 @@ pub use messages::Sbp;
...
 pub use crate::messages::SbpMessage;
 
+#[doc(inline)]
+pub use crate::de::Decoder;
+
+#[doc(inline)]
+pub use crate::de::Framer;
+
+#[doc(inline)]
+pub use crate::de::Frame;
```
notoriaga, line 199: nit: you can combine all these with the other `pub use de::*` below

notoriaga (approved):
Looks great! I might try using this in one of the projects that would benefit from this (like sbp-filter) in order to make sure the api is right/you have access to everything you need in a downstream crate. But you could also address anything in a separate pr

rust/sbp/src/lib.rs
adrian-kong: Does that include 
```rust
#[cfg(feature = "async")]
#[doc(inline)]
pub use de::{stream_messages, stream_messages_with_timeout};
```
so replacing all the `pub use de::...`  to `pub use de::*`
or just pub use `crate::de::*`


i.e.

```rust
#[doc(inline)]
pub use de::{Error as DeserializeError, *};

#[cfg(feature = "async")]
#[doc(inline)]
pub use de::{stream_messages, stream_messages_with_timeout};
```

rust/sbp/src/de.rs
```diff
@@ -196,77 +245,203 @@ impl Decoder for SbpDecoder {
...
+    }
+
+    /// Checks frame size, used in [`RawDecoder`] to specify
+    /// whether frame is dummy struct to contain bytes
+    fn check_frame(&self) -> bool {
+        self.0.len() > HEADER_LEN
+    }
+
+    pub fn to_sbp(&self) -> Result<Sbp, Error> {
+        if !self.check_frame() {
```
adrian-kong, line 349: passing an invalid frame as dummy for raw decoder, not sure if ideal

adrian-kong: nvm... seems like i broke something

adrian-kong: reverted and just merging previous decoder

PR #1223, opened 2022-09-14
Finish porting test suite to Bazel [BUILD-311]

krisukox:
# Description

@swift-nav/devinfra

This PR adds the following bazel targets:
- sbp-v4-test
- sbp-cpp-v4-test
- sbp-string-test
- sbp-cpp-legacy-test

In the sbp-cpp-legacy-test, `SbpStdioTest.ReadsSbpFiles` and `SbpStdioTest.WritesToSbpFiles` targets have undefined behavior. Assignment of msg to payload in the `MsgObsHandler::handle_sbp_msg` function causes this issue.
It works better with gcc-11 but still, these two testcases shouldn't be run together. In order to run all the test cases:
`CC=gcc-11 CXX=g++-11 bazel test --test_output=all //c:all --test_arg=--gtest_filter=-SbpStdioTest.WritesToSbpFiles`
`CC=gcc-11 CXX=g++-11 bazel test --test_output=all //c:sbp-cpp-legacy-test --test_arg=--gtest_filter=SbpStdioTest.WritesToSbpFiles`

# JIRA Reference

https://swift-nav.atlassian.net/browse/BUILD-311

### DIFF
--- a/c/BUILD.bazel
+++ b/c/BUILD.bazel
@@ -47,11 +47,11 @@ cc_library(
         "src/v4/vehicle.c",
     ],
     hdrs = SBP_INCLUDE + SBP_INCLUDE_INTERNAL,
+    copts = ["-Ic/src/include"],
     includes = [
         "include",
-        "src/include",
-    ],  # FIXME - Don't expose this path publically.
-    visibility = [ "//visibility:public" ],
+    ],
+    visibility = ["//visibility:public"],
 )
 
 SBP_LEGACY_C_SOURCES = glob(["test/legacy/auto*.c"])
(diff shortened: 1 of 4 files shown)


c/BUILD.bazel
```diff
@@ -65,9 +65,73 @@ cc_test(
...
+    srcs = [
+        "test/legacy/cpp/test_sbp_stdio.cc",
+    ] + SBP_CPP_LEGACY_SOURCES,
+    data = [
+        "test/legacy/cpp/sbp_data/gnss_data.sbp",
+    ],
+    deps = [
+        ":sbp",
+        "@my-googletest//:gtest_main",
+    ],
```
isaactorz, line 136:
```suggestion
    ],
    tags = ["exclusive"],
```
This attribute will make it so that the test run serially after the other tests. I don't think it's as granular as making the individual tests themselves run in serial but it might be worth a shot.

isaactorz (approved):
I'm going to chalk the test failure up to the bazel submodule not being public. Not super concerning since we're just merging into a branch.

I tested locally without your command line flags and I didn't run into any errors with either gcc-11 or clang-14.

c/BUILD.bazel
isaactorz: Doc here: https://bazel.build/reference/be/common-definitions#common-attributes-tests

krisukox: Today, even when I run all of the tests at once, it works for me with gcc-11.
Clang-14 still doesn't work, even with the "exclusive" tag.
The behavior is so unpredictable. I will create a ticket about it.

PR #1224, opened 2022-09-19
Add bazel build for libsbp c library

isaactorz:
# Description

Adds support for building libsbp c library using bazel. Adds a submodules that contains common helper scritps.

@swift-nav/devinfra

# API compatibility

No API compatibility risks.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -18,3 +18,15 @@ load("@rules_foreign_cc//foreign_cc:repositories.bzl", "rules_foreign_cc_depende
 
 rules_foreign_cc_dependencies()
 
+# Hedron's Compile Commands Extractor for Bazel
+# https://github.com/hedronvision/bazel-compile-commands-extractor
+http_archive(
+    name = "hedron_compile_commands",
+    sha256 = "4b251a482a85de6c5cb0dc34c5671e73190b9ff348e9979fa2c033d81de0f928",
+    strip_prefix = "bazel-compile-commands-extractor-5bb5ff2f32d542a986033102af771aa4206387b9",
+    url = "https://github.com/hedronvision/bazel-compile-commands-extractor/archive/5bb5ff2f32d542a986033102af771aa4206387b9.tar.gz",
+)
+
+load("@hedron_compile_commands//:workspace_setup.bzl", "hedron_compile_commands_setup")
+
+hedron_compile_commands_setup()
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -6,6 +6,11 @@ new_local_repository(
     path = "c/third_party/check",
 )
 
+local_repository(
+    name = "my-googletest",
+    path = "c/third_party/googletest",
+)
+
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,0 +1,7 @@
+workspace(name = "root")
+
+new_local_repository(
+    name = "my-check",
+    build_file = "bazel/check.BUILD",
+    path = "c/third_party/check",
+)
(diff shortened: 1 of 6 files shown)


c/test/legacy/cpp/CMakeLists.txt
```diff
@@ -13,4 +13,4 @@ swift_add_test(test-libsbp-cpp-legacy
 swift_set_language_standards(test-libsbp-cpp-legacy)
 swift_set_compile_options(test-libsbp-cpp-legacy ADD -Wno-error=array-bounds)
 
-file(COPY sbp_data/ DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+file(COPY sbp_data/ DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/c/test/legacy/cpp/sbp_data/)
```
isaactorz, line 16: Bazel copies all files into a sandbox for building and execution. Data file dependencies are declared using the `data` attribute and are copied into the sandbox with the same path they have relative to the project root. This makes it so that the cmake version of the tests and the bazel version of the test have the files in the same place at runtime.

WORKSPACE
```diff
@@ -0,0 +1,37 @@
...
+    name = "rules_foreign_cc",
+    strip_prefix = "rules_foreign_cc-c65e8cfbaa002bcd1ce9c26e9fec63b0b866c94b",
+    url = "https://github.com/bazelbuild/rules_foreign_cc/archive/c65e8cfbaa002bcd1ce9c26e9fec63b0b866c94b.tar.gz",
+)
+
+load("@rules_foreign_cc//foreign_cc:repositories.bzl", "rules_foreign_cc_dependencies")
+
+rules_foreign_cc_dependencies()
+
+# Hedron's Compile Commands Extractor for Bazel
```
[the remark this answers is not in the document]
isaactorz: Yes we're using these rules to output a `compile_commands.json` file.

silverjam (approved):

c/test/legacy/cpp/test_sbp_stdio.cc
```diff
@@ -97,11 +96,12 @@ class SbpStdioTest : public ::testing::Test {
 };
 
 TEST_F(SbpStdioTest, ReadsSbpFiles) {
-  EXPECT_EQ(num_entries_in_file("gnss_data.sbp"), 3);
+  EXPECT_EQ(num_entries_in_file("c/test/legacy/cpp/sbp_data/gnss_data.sbp"), 3);
```
achubbic-snav, line 99: Please replace magic number `3` or leave a comment describing why 3 entries in the file are expected.

Comments indicating what this test seeks to validate and documentation of `num_entries_in_file` would also be appreciated

c/test/legacy/cpp/test_sbp_stdio.cc
```diff
@@ -97,11 +96,12 @@ class SbpStdioTest : public ::testing::Test {
...
 }
 
 TEST_F(SbpStdioTest, WritesToSbpFiles) {
-  write_to_file("gnss_data.sbp", "gnss_data_output.sbp");
+  write_to_file("c/test/legacy/cpp/sbp_data/gnss_data.sbp",
+                "gnss_data_output.sbp");
   EXPECT_EQ(num_entries_in_file("gnss_data_output.sbp"), 9);
```
achubbic-snav, line 105: Ditto regarding magic numbers and documentaion

achubbic-snav (changes requested):
Left some comments about magic numbers and documentation

c/test/legacy/cpp/test_sbp_stdio.cc
isaactorz: I'm just updating the path here. I don't know the context behind the magic numbers in these tests.

PR #1233, opened 2022-10-25
Add message stream counter fields [SKYL-34]

notoriaga:
# Description

@swift-nav/devinfra

Adds two message counter fields to `MSG_ED25519_SIGNATURE`

# API compatibility

Yes

## API compatibility plan

It's not a publicly used message so it should be fine

# JIRA Reference

https://swift-nav.atlassian.net/browse/SKYL-34


silverjam:
@notoriaga please see discussion on slack

spec/yaml/swiftnav/sbp/signing.yaml
```diff
@@ -17,9 +17,20 @@ include:
   - gnss.yaml
 definitions:
   - MSG_ED25519_SIGNATURE:
-      id: 0x0C01
+      id: 0x0C03
```
woodfell, line 20: Don't we need to retain the old definition and message number as a `_DEP` message?

silverjam: Yes

spec/yaml/swiftnav/sbp/signing.yaml
```diff
@@ -17,9 +17,20 @@ include:
...
       short_desc: ED25519 signature for groups of RTCM messages
       fields:
+        - stream_counter:
+            type: u8
+            desc: >
+              Signature message counter. Zero indexed and incremented with each signature message.
```
silverjam, line 26: Since SBP in general is supposed to be stateless, we should indicate that the starting value can be anything.  This will allow us to not worry about connection state (as much) in the implementation and clients can jump into something "mid stream" and work under assumption that first value isn't necessarily zero.

This would mean clients would have three states to implement:
- no counter seen -> record given value, assume stream valid
- counter seen and less than 255, compare with previous counter, if diff >1 then stream is invalid
- previous counter is 255, next counter should be 0 else stream is invalid

c/include/libsbp/v4/signing/MSG_ED25519_SIGNATURE.h
```diff
@@ -39,6 +39,22 @@ extern "C" {
...
+  u8 stream_counter;
+
+  /**
+   * On demand message counter. Zero indexed and incremented with each signature
+   * message sent in response to an on demand message. The counter will roll
+   * over after 256 messages. Upon connection, the value of the counter may not
+   * initially be zero.
+   */
+  u8 on_demand_counter;
+
```
Blast545, line 57: I think one `sbp_msg_ed25519_signature_t` would have either of the counter fields, maybe those should be defined as an `union`?

c/src/v4/signing.c
```diff
@@ -244,3 +245,143 @@ int sbp_msg_ed25519_certificate_cmp(const sbp_msg_ed25519_certificate_t *a,
...
 }
+
+bool sbp_msg_ed25519_signature_encode_internal(
+    sbp_encode_ctx_t *ctx, const sbp_msg_ed25519_signature_t *msg) {
+  if (!sbp_u8_encode(ctx, &msg->stream_counter)) {
+    return false;
+  }
+  if (!sbp_u8_encode(ctx, &msg->on_demand_counter)) {
+    return false;
+  }
```
Blast545, line 256: Requirement says: 

"The proposal is to have two separate counters.  One to count the messages in the stream and a separate counter for burst or on-demand messages. "

It's not clear to me whether having both (being one static) is desired for this feature.

c/test/legacy/auto_check_sbp_signing_MsgEd25519Signature.c
```diff
@@ -691,7 +692,8 @@ START_TEST(test_legacy_auto_check_sbp_signing_MsgEd25519Signature) {
       test_msg_len += sizeof(test_msg->signed_messages[0]);
     }
     test_msg->signed_messages[24] = 10616;
-    sbp_payload_send(&sbp_state, 0xC01, 9876, test_msg_len, test_msg_storage,
+    test_msg->stream_counter = 1;
+    sbp_payload_send(&sbp_state, 0xC03, 66, test_msg_len, test_msg_storage,
                      &dummy_write);
```
Blast545, line 697: I undestand that the change from 0xC01 to 0xC03 comes from the new msg type definition, can you explain where the change from 9876 to 66 comes from?

c/include/libsbp/legacy/signing.h
```diff
@@ -46,6 +46,26 @@ typedef struct SBP_ATTR_PACKED {
...
+                               incremented with each signature message sent
+                               in response to an on demand message. The
+                               counter will roll over after 256 messages.
+                               Upon connection, the value of the counter may
+                               not initially be zero. */
+  u8 signature[64];       /**< ED25519 signature for messages. */
+  u8 fingerprint[20];     /**< SHA-1 fingerprint of the associated
+                               certificate. */
+  u32 signed_messages[0]; /**< CRCs of signed messages. */
+} msg_ed25519_signature_t;
```
Blast545, line 67: Should this be defined in a non legacy header?

Moving this definition from /libsbp/legacy/signing.h to /libsbp/legacy/signing.h ?

Blast545: I just saw that using #include <libsbp/v4/signing.h> includes the new msg definition, so I'm not sure if this should be moved there

notoriaga: It should be in both, AFAIK we support both apis

c/test/legacy/auto_check_sbp_signing_MsgEd25519Signature.c
notoriaga: When I regenerated the test yaml with json2test it used 42 as the sender id https://github.com/swift-nav/libsbp/blob/c0327571b0bc3fb479f4bfaf25911b3d8a45a87d/spec/tests/yaml/swiftnav/sbp/signing/test_MsgEd25519Signature.yaml#L143

c/src/v4/signing.c
notoriaga: I think the situation in which you'd use both is - you have "normal" groups coming in response to a CRA. via the counter you know you missed one, you send a new CRA message with the correction mask set so that you can request whatever you missed, at this point you'd be getting both "normal" and on demand groups

notoriaga: Not entirely sure if that's how it'll work though. Like if you send two CRAs like that, would the second one "cancel" the first one? If so then one counter would be enough I think

c/include/libsbp/v4/signing/MSG_ED25519_SIGNATURE.h
notoriaga: I don't think the spec/codegen have support for that. A 16 byte bitfield would be kind of close. But, the requirements specify two separate fields

spec/yaml/swiftnav/sbp/signing.yaml
```diff
@@ -54,3 +56,35 @@ definitions:
             type: array
             fill: u8
             desc: ED25519 certificate bytes.
+
+  - MSG_ED25519_SIGNATURE:
```
RReichert, line 60: These changes will need to be reflected [here](https://github.com/swift-nav/rtcm2signed/blob/f05b2ca3a4b177dd3d63f876e9220855ea98ca6b/rtcm2signed/rtcm/frame_signer.go#L22-L29) right?

RReichert (approved):
I'd say this looks good, but please get a sign off from the rest of the folks that have commented.

spec/yaml/swiftnav/sbp/signing.yaml
notoriaga: Yeah, although we are pretty close to moving the signing from there, so I don't think we need to update that

c/src/v4/signing.c
notoriaga: Agreed

silverjam (approved):

PR #1244, opened 2022-11-21
Error logging in converters [DEVINFRA-1018]

adrian-kong:
# Description

Logs error, terminates the unterminating json bug

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-1018


silverjam (commented):
Can we do the same thing for fatal errors too?

adrian-kong:
> Can we do the same thing for fatal errors too?

yep but will need to take until inclusive or chain (?) or just panic on the filter map 🤔 which i have just opted for the panic ^^

another bug (?) that arises json2sbp never terminates i.e.
```
echo {"crc":10555,"length":17,"msg_type":2304,"payload":"xYZjQJhhAWwAYhDy/3v+DgA=","preamble":85,"sender":51228,"tow":1080264389,"tow_f":152,"acc_x":353,"acc_y":108,"acc_z":4194,"gyr_x":1-14,"gyr_y":-389,"gyr_z":14} | cargo run --bin json2sbp
```
will not terminate which might look into (?) but we have `--fatal-errors` option


the above might be the reason why sometimes it doesn't produce back the entire sbp file, hanging the decoder

rust/sbp2json/src/lib.rs
```diff
@@ -99,9 +99,9 @@ where
     I: Iterator<Item = std::result::Result<M, E>> + 'a,
     E: std::error::Error + 'a,
 {
-    if fatal_errors {
-        Box::new(messages.take_while(|m| m.is_ok()).map(|m| m.unwrap()))
-    } else {
-        Box::new(messages.filter_map(|m| m.ok()))
-    }
+    let log_err = move |e| match fatal_errors {
```
silverjam, line 102: clippy will complain about this code, just use an `if` statement: https://rust-lang.github.io/rust-clippy/master/#match_bool

rust/sbp2json/src/lib.rs
```diff
@@ -99,16 +99,9 @@ where
...
-            if let Ok(m) = m {
-                Some(m)
-            } else {
-                eprintln!("{:?}", m.err());
-                None
-            }
-        }))
-    }
+    let log_err = move |e| match fatal_errors {
+        true => panic!("{:?}", e),
```
silverjam: Not sure we should panic, this is a pretty big behavior change, with fatal_errors I'm assuming you could catch the errors further up the chain, without "fatal errors" we just swallow errors?

adrian-kong: yea think --fatal-errors just makes it so it stops after error,
which more weird is not sure if this was just to fix the case of unterminating iterator because looks like running json2sbp on a broken json will not terminate

adrian-kong: maybe we just have --fatal-errors not log error ? since we take until OK, so we bascially assume the next one must be a parse error anyways

rust/sbp/src/json/de.rs
```diff
@@ -75,11 +76,15 @@ impl Decoder for JsonDecoder {
     type Error = JsonError;
 
     fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
-        let value = match decode_one::<JsonInput>(src)? {
+        let ret = decode_one::<JsonInput>(src);
+        if ret.is_err(){
```
adrian-kong, line 80: Hacky fix to terminate the ParserError loop

rust/sbp/src/json/de.rs
```diff
@@ -10,6 +10,7 @@ use crate::{
     messages::Sbp,
     BUFLEN,
 };
+use crate::messages::logging::MsgLog;
```
silverjam, line 13: Remove?

adrian-kong: oops

adrian-kong, line 13:
```suggestion
```

rust/sbp/src/json/de.rs
```diff
@@ -112,6 +116,6 @@ where
     match value.transpose() {
         Ok(v) => Ok(v),
         Err(e) if e.is_eof() => Ok(None),
-        Err(e) => Err(e),
+        Err(e) => Err(e), // returns error, on json errors
```
adrian-kong, line 119: this is the tip of the loop issue - returning error inside decode

silverjam:
If we don't dump the buffer on EOF errors, we terminate processing but we also don't capture the error, for example:

```
❯ echo -e '{"msg_type":1' | cargo run --quiet --bin json2sbp --
# ... no output ...

❯ echo -e '{"msg_type":1' | cargo run --quiet --bin json2sbp -- --fatal-errors
# ... no outputs ...
```

Versus previously
```
❯ echo -e '{"msg_type":1' | cargo run --quiet --bin json2sbp --
SerdeJsonError(Error("EOF while parsing an object", line: 2, column: 0))

❯ echo -e '{"msg_type":1' | cargo run --quiet --bin json2sbp -- --fatal-errors
SerdeJsonError(Error("EOF while parsing an object", line: 2, column: 0))
```

silverjam:
@adrian-kong You're change is probably correct, but it would be nice if we could detect a "real" EOF error and an EOF error where we're at the true end of a stream.

silverjam:
@adrian-kong I wonder how hard it would be to add some integration tests for the error message?

rust/sbp/src/json/de.rs
```diff
@@ -44,12 +44,14 @@ pub fn stream_messages<R: futures::AsyncRead + Unpin>(
...
 struct JsonDecoder {
     payload_buf: Vec<u8>,
+    eof_seen: bool,
 }
 
 impl JsonDecoder {
     fn new() -> Self {
         JsonDecoder {
             payload_buf: Vec::with_capacity(BUFLEN),
+            eof_seen: false,
```
adrian-kong, line 54: was trying to figure out a better way instead of holding this - seemed really hacky but i guess it works 👍 
- or maybe just `eprintln` inside decode, thought the dencode errors meant retry decoding so we needed two error types

rust/sbp2json/tests/test_error.rs
```diff
@@ -0,0 +1,27 @@
+mod common;
+
+use crate::common::{data_path, find_project_root};
+use assert_cmd::cargo::CommandCargoExt;
+use assert_cmd::Command;
+
+/// Asserts termination of json decoder on broken inputs
+#[test]
+fn test_json_converters_termination() {
```
adrian-kong, line 9: added a couple of simple tests for the termination case that wasn't previously handled and some errors
think we won't need much of the error message tests since `serde_json` should cover them

@silverjam

silverjam (approved):

rust/sbp/src/json/de.rs
silverjam: Agree that the solution is a bit hacky.  However, since these are library functions, it didn't feel right to add `eprintln` there (printing errors should be an application layer behavior).

PR #1245, opened 2022-11-21
Add message name to sbp2json output [DEVINFRA-1019]

ljbade:
# Description

@swift-nav/devinfra

In the JSON output of `sbp2json`, a new field `msg_name` is added which gives the SBP message name. This improves the readability of the output by avoiding the need to grep for message numbers.

# API compatibility

I don't think this will break the API as it adds a new field, and the field should be ignored when reading the output with older version of `json2sbp`.

@silverjam is this correct?

## API compatibility plan

N/A

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-1019


silverjam (approved):

silverjam:
> I don't think this will break the API as it adds a new field, and the field should be ignored when reading the output with older version of `json2sbp`.
> 
> @silverjam is this correct?

Yeah, it'll just be ignored

ljbade:
@silverjam do you know why the `test_continue_on_error` is failing? It seems that it is finding more SBP messages than it expects, but I don't understand why this PR would have changed that.

silverjam (approved):

PR #1250, opened 2022-11-23
ci: bump/pin to ubuntu-20.04

silverjam:
# Description

CI: bump and pin to Ubuntu 20.04 to fix recent CI breakage

# API compatibility

n/a

# JIRA Reference

n/a

### DIFF
--- a/.github/workflows/c.yaml
+++ b/.github/workflows/c.yaml
@@ -14,7 +14,7 @@ on:
 jobs:
   coverage:
     name: Code Coverage
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/c.yaml
```diff
@@ -34,9 +34,9 @@ jobs:
       - name: Build examples
         run: make -C c/build examples
 
-      - name: Upload coverage to codecov.io
-        run: |
-          bash <(curl -s https://codecov.io/bash) -s c/build || echo "Codecov did not collect coverage reports";
+#      - name: Upload coverage to codecov.io
+#        run: |
+#          bash <(curl -s https://codecov.io/bash) -s c/build || echo "Codecov did not collect coverage reports";
```
ljbade, line 39: Why is this commented out?

silverjam: We aren't using codecov.io anymore (we closed our account) -- so this should be migrate to SonarCloud eventually

.github/workflows/c.yaml
```diff
@@ -12,34 +12,9 @@ on:
       - "c/**"
       - .github/workflows/c.yaml
 jobs:
-  coverage:
-    name: Code Coverage
```
silverjam, line 16: Removed this stage since it isn't doing anything, we've deleted our code cov account

ljbade (approved):

PR #1254, opened 2022-11-25
Implement HasTime for Frames [DEVINFRA-516]

adrian-kong:
Intended for `with_rover_time`, important to not do redundant `to_sbp` calls after, so this just assumes that we want

`(Frame,  Time)`
since we cant easily convert from sbp -> frame,

or maybe alternatively we can retain frame and sbp.

This is so that we can do 
`Framer::new(...).with_rover_time()` for `(Frame, Time)` tuple
instead of only 
`sbp::iter_messages(...).with_rover_time()` for `(Sbp, Time)` tuple

### DIFF
--- a/rust/sbp/src/sbp_iter_ext.rs
+++ b/rust/sbp/src/sbp_iter_ext.rs
@@ -119,7 +119,11 @@ where
 mod swiftnav_impl {
     use swiftnav::time::GpsTime;
 
-    use crate::{messages::SbpMessage, time::{GpsTimeError, MessageTime, RoverTime}, Sbp, Frame};
+    use crate::{
+        messages::SbpMessage,
+        time::{GpsTimeError, MessageTime, RoverTime},
+        Frame, Sbp,
+    };
 
     /// See [SbpIterExt::with_rover_time] for more information.
     pub struct RoverTimeIter<I: Iterator> {
--- a/rust/sbp/src/sbp_iter_ext.rs
+++ b/rust/sbp/src/sbp_iter_ext.rs
@@ -119,11 +119,7 @@ where
 mod swiftnav_impl {
     use swiftnav::time::GpsTime;
 
-    use crate::{
-        messages::SbpMessage,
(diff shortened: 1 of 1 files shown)


joelynch (approved):
Seems good

notoriaga:
One thing that bothers me a bit about this is that if you have 

```rust
for (frame, time) in Framer::new(...).with_rover_time() {
    ....
    let msg = frame.to_sbp();
    ...
}
```

you end up converting the frame twice.

> or maybe alternatively we can retain frame and sbp.

would avoid that, but not sure what a clean way to do that is. maybe we could add a method to `Framer` that returns both

```rust
impl Framer {
    pub fn with_messages(self) -> impl Iterator<Item=(Frame, Result<Sbp>)> {
        self.map(|f| (f, f.to_sbp()))
    }
}
```

and then you can impl `HasTime` for `(Frame, Result<Sbp>)` as well. Either way I think it's fine as is

notoriaga (approved):

silverjam:
> 



> One thing that bothers me a bit about this is that if you have
> 
> ```rust
> for (frame, time) in Framer::new(...).with_rover_time() {
>     ....
>     let msg = frame.to_sbp();
>     ...
> }
> ```
> 
> you end up converting the frame twice.
> 
> > or maybe alternatively we can retain frame and sbp.
> 
> would avoid that, but not sure what a clean way to do that is. maybe we could add a method to `Framer` that returns both
> 
> ```rust
> impl Framer {
>     pub fn with_messages(self) -> impl Iterator<Item=(Frame, Result<Sbp>)> {
>         self.map(|f| (f, f.to_sbp()))
>     }
> }
> ```
> 
> and then you can impl `HasTime` for `(Frame, Result<Sbp>)` as well. Either way I think it's fine as is

Let's track with an issue and revisit @adrian-kong

silverjam (approved):

adrian-kong:
created https://swift-nav.atlassian.net/browse/DEVINFRA-1028

PR #1256, opened 2022-11-28
Enriched display [DEVINFRA-389]

adrian-kong:
Generate additional enriched field display and friendly names

Introduces two functions:
`Sbp::friendly_name()`
`Sbp::message_display():`

`friendly_name` mainly follows replacing `MSG_`, `_` and abbreviating some keywords, manual implementation also possible via yaml. Mainly autogenerated

`message_display` aims to produce some sort of debug / format message output. Can be manually implemented via yaml with custom regex grouping `{{ field }}`  to denote `field` and `{{@ field @}}` to represent `self.field`


spec/yaml/swiftnav/sbp/piksi.yaml
```diff
@@ -521,6 +523,7 @@ definitions:
 
  - MSG_CELL_MODEM_STATUS:
     id: 0x00BE
+    message_display: "{{@ signal_strength @}} dBm | {{@ signal_error_rate @}} %"
```
adrian-kong, line 526: I went with a funky regex group capturing

`abc{{X}}` represents `format!("abc{}", X)`
`abc{{@ X @}}` represents `format!("abc{}", self.X)`

Basically `{{@ ... @}}` prepends `self.`
This allows some things that only requires fields to be simplified but might be bit hacky.

Missing functionality to do something like this `format!("{:.2}", self.X)` 
Maybe TODO: add more parsing for some syntaxes like `{{@ ... @}:.3}` where we also retain `{{}:.}` some element after char `:`.

We can parse methods if there are by just calling them like `{{@ some_method() @}}` etc... but maybe really hacky to be inside yaml

generator/sbpg/targets/rust.py
```diff
@@ -351,6 +376,20 @@ def __init__(self, msg, package_specs, field):
         if self.options.get("fields", False):
             self.bitfield = get_bitfield(self)
 
+# pattern to capture {{groups}} for enriched message display
+ENRICH_PAT = re.compile("{{([^}]+)}}[.]*", re.S)
```
joelynch, line 380: Definitely would like to see some tests for all this

joelynch: What is the `[.]*` at the end doing?  followed by 0 or more of any character?  This is meaningless right.

generator/sbpg/targets/rust.py
```diff
@@ -351,6 +376,20 @@ def __init__(self, msg, package_specs, field):
...
+ENRICH_FIELD_PAT = re.compile("@ ([^@]+) @[^\@]*", re.S)
```
joelynch, line 381: Be consistent with escaping the `@`.  The final `[^\@]*` is also a bit confusing here I don't think this part is necessary and `@ ([^@]+) @` should work just as well.

generator/sbpg/targets/rust.py
joelynch: Also the re.S option is a bit weird since these fields shouldn't have newlines in them

joelynch (changes requested):
This is good but I think it needs a few tweaks.

spec/yaml/swiftnav/sbp/piksi.yaml
adrian-kong: The first one could be used something like 
```
{{@ posX @ / 10}}
format!({}, self.posX / 10)
```

There are cases where we do [operations](https://github.com/swift-nav/gt/blob/e9e97373532aa7f39c94cdd7a2abc0c1ac72b12c/sw/mod/proto/sbp/gt_proto_sbp_csv.c#L975) operations on these fields - which might turn out really hacky inside the yaml when it gets really complicated as something like

```
{{if @ bool @ {5} else { 10 }}}
format!("{}", if self.bool { 5 } else { 10 })
```

Maybe there is a better way since these format methods would turn out really funky codegen wise, but avoided manual implementations of `message_display` by these scripting formats

adrian-kong: But yea `@field_name` style might be nice but would be weird to decide where to terminate i.e.
```rust
{{@posX + @posY }} // requiring end char to be ' '
{{@posX+@posY}} // no ' ' termination character
```
*Having `@field_name` might be simpler but was unsure if '@' came up often so might swap to this style instead

joelynch: Ah okay I see.  These YAMLs probably want to be language independent right, we would not be able to use `message_display` in the python library if it is a rust expression.  I think until there are usecases for this it might be simpler to only allow simple string interpolation for now.

silverjam: I think we can use {fmt} style interpolation in most languages, but it'd take some work, and it'd need to be done in a language neutral way, so we couldn't have embedded Rust/whatever syntax:
- Java: https://fmt.dev/5.2.1/index.html
- C++: https://github.com/fmtlib/fmt
- Rust (obviously): https://doc.rust-lang.org/std/fmt/
- https://www.w3schools.com/python/ref_string_format.asp
- JavaScript? https://www.npmjs.com/package/format-unicorn

This feature would not work for languages that don't implement some variant of the {fmt} string interpolation.

adrian-kong:
Going to reopen this in smaller PR, have the more complex `message_display()` seperate

adrian-kong:
split into https://github.com/swift-nav/libsbp/pull/1262

PR #1260, opened 2022-12-07
convert asserts to exception in java [DEVINFRA-1035]

adrian-kong:
handle these assertion errors as SBPBinaryException rather than asserts


silverjam (approved):
Probably should turn off code duplication as a SonarCloud metric for libsbp

PR #1262, opened 2022-12-07
add friendly_name method [DEVINFRA-1045]

adrian-kong:
Add friendly_name method which trims out keywords

``` 
{
        "TRACKING": "TRK",
        "MEASUREMENT": "MEAS",
        "INDEX": "IDX",
        "NETWORK": "NET",
        "EPHEMERIS": "EPH",
        "_": " "
}

and 
"^MSG_"
"_GNSS$"
```


silverjam (changes requested):
To statisfy SonarCloud let's add a simple unit test to Java that asserts that the friend name for each message is not empty.

silverjam (approved):

generator/sbpg/targets/resources/rust/sbp_messages_template.rs
```diff
@@ -140,6 +140,12 @@ impl SbpMessage for (((m.msg_name))) {
     (((m.gps_time_fn)))
 }
 
+impl FriendlyName for (((m.msg_name))) {
+    fn friendly_name() -> &'static str {
+        "(((m.friendly_name)))"
+    }
+}
```
notoriaga, line 147: This could probably also be just a new function on the `SbpMessage` trait, then you could also use it on the `Sbp` enum without having to manually get a reference to the inner message

notoriaga: Oh I see this isn't a method it's an associated function. Might be nice to have `fn friendly_name(&self) -> &'static str` on `SbpMessage`. Otherwise if you want to get the friendly name of a message it would be something like 

```rust
let msg = sbp::iter_messages(r).next().unwrap().unwrap();
match msg {
    Sbp::MsgLog(_) => eprintln!("{}", MsgLog::friendly_name()),
    _ => {}
}
```

vs

```rust
let msg = sbp::iter_messages(r).next().unwrap().unwrap();
eprintln!("{}", msg.friendly_name());
```

notoriaga: Having the associated function version also seems useful for some stuff. So maybe leave this as is and then you could add `fn friendly_name(&self) -> &'static str` to the `SbpMessage` trait (which would just call out to `FriendlyName::friendly_name()`)

notoriaga (approved):
One comment about possibly adding `fn friendly_name(&self) -> &'static str` to `SbpMessage` so you can directly fetch the friendly name, but that can be done later so LGTM!

generator/sbpg/targets/resources/rust/sbp_messages_template.rs
silverjam: 👍

PR #1265, opened 2022-12-11
Add frame iterators

adrian-kong:
To preserve error frames, adds additional stream / timeout iterators


rust/sbp/src/de.rs
```diff
@@ -44,12 +44,29 @@ pub fn iter_messages<R: io::Read>(input: R) -> impl Iterator<Item = Result<Sbp,
...
+    Framer::new(input)
+}
+
 /// Deserialize the IO stream into an iterator of messages. Provide a timeout
 /// for the maximum allowed duration without a successful message.
 pub fn iter_messages_with_timeout<R: io::Read>(
     input: R,
     timeout_duration: Duration,
 ) -> impl Iterator<Item = Result<Sbp, Error>> {
+    TimeoutDecoder::new(input, timeout_duration).map(|f| match f {
```
adrian-kong, line 58: since timeoutdecoder is private, this should not be api breaking

notoriaga (approved):
noice

PR #1272, opened 2022-12-14
Add message display method and fields

adrian-kong:
wip

### DIFF
--- a/generator/sbpg/specs/yaml_schema.py
+++ b/generator/sbpg/specs/yaml_schema.py
@@ -50,6 +50,7 @@ field = Schema({identifier:
 definition = Schema({identifier:
                      {Optional('id'): sbp_identifier,
                       Optional('friendly_name'): description,
+                      Optional('message_display'): description,
                       Optional('short_desc'): description,
                       Optional('desc'): description,
                       Optional('replaced_by'): [identifier],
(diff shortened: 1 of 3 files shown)


silverjam:
Maybe too complicated to have all languages share the same message display implementation....

What if we did something like:

```yaml
message_display:
  python: "thing: {field1:.02f}, other thing: {field2:.03f}"
  rust: "thing: {field1:?}, other thing: {field2:.5}"
```

Then we'd only emit the method for language that support f-strings natively (like Python and Rust), but allow the formats to be slightly different, depending on the specific language's implementation.  Maybe this would be more tractable?

adrian-kong:
> Maybe too complicated to have all languages share the same message display implementation....

i don't think it would be more complicated to share the same display fmt since we are extracting all the field params out as a vec

the main difficulty that i haven't decided was the grammar like inlining if else, for eachs and accessing bit arrays creates a much more complicated `message_display` in yaml
^^ which i believe this issue still stands if we do non-language specific things where we would have massive code blocks in yaml

trying to avoid manual implementations to form an actual grammar / mini scripting language to best generate (and future-proof) this function

im sure there is an tool to translate simple string interpolations from language to other languages (which i also haven't looked into pulling into libsbp or coming up with one)

i haven't worked on much of this PR (seems lower prio from these QOL changes) but mainly those are some ideas and so if we dont need to 1-1 match sbp2report we can probably reduce complexity by a bit

adrian-kong:
closing stale PR

Issue #1240, opened 2022-11-12
[C++] MessageTrait: name attribute


Cerclique:
Do you have any plan to add a new auto-generated attribute that would return the name of the current type as a string ? 

Perhaps something like this: 

```c++
template <>
struct MessageTraits<sbp_msg_utc_time_gnss_t> {
   ...
   static constexpr char* name = "sbp_msg_utc_time_gnss_t";
   ...
};
```

silverjam:
No plans yet, but it should pretty easy to implement. I’ll see if we can get it onto our backlog.

silverjam:
Hi @Cerclique -- this change has been merged, it will be released in the next minor release of the library.

PR #1278, opened 2022-12-20
C++ MessageTrait name attribute [DEVINFRA-1034]

ljbade:
# Description

Implements the request in #1240 for a `name` attribute in the C++ `MessageTrait` class.

It results in something like:
```
template <>
struct MessageTraits<sbp_msg_acknowledge_t> {
  static constexpr sbp_msg_type_t id = SbpMsgAcknowledge;
  static constexpr char *name = "MSG_ACKNOWLEDGE";
...
```

For each message

@swift-nav/devinfra

# API compatibility

No it is adding a new class member only.

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-1034

### DIFF
--- a/c/include/libsbp/cpp/message_traits.h
+++ b/c/include/libsbp/cpp/message_traits.h
@@ -58,6 +58,7 @@ struct MessageTraits;
 template <>
 struct MessageTraits<sbp_msg_acknowledge_t> {
   static constexpr sbp_msg_type_t id = SbpMsgAcknowledge;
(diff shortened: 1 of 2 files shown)


silverjam (approved):

silverjam:
@isaactorz any hints on what's going on with the bazel build? Not an error I've seen before.

isaactorz:
@silverjam @ljbade Might have something to do with 6.0 LTS being released yesterday.

PR #1284, opened 2023-01-09
Add new SSR Satellite APC and Tile Definition message [PC-116]

Davinco:
# Description
A new version of the SSR Satellite APC message and SSR Tile Definition message is added which extends the current messages with a timestamp, update interval, solution id and iod fields. See details here:
- https://swift-nav.atlassian.net/wiki/spaces/SN/pages/2435940383/ICD+-+CR1+-+APC
- https://swift-nav.atlassian.net/wiki/spaces/SN/pages/2452979847/ICD+-+CR3+-+Tile+Definition

# API compatibility
No, this PR adds new messages and the current messages are still functional but labelled as deprecated.

# JIRA Reference
https://swift-nav.atlassian.net/browse/PC-116


Davinco:
Putting up this PR to get some feedback. I got the same test failure when I ran this locally, is there something wrong with how the `spec/tests/yaml/swiftnav/sbp/ssr/test_MsgSsrSatelliteApc.yaml` was generated? or with my input file? I created the json file in [sat_apc.zip](https://github.com/swift-nav/libsbp/files/10377486/sat_apc.zip) and ran the command from the howto `PYTHONPATH="python/" python generator/json2test.py --input sat_apc.json --output spec/tests/yaml/swiftnav/sbp/ssr/test_MsgSsrSatelliteApc.yaml`, and then ran `make gen`.

silverjam:
@Davinco I think the json2test script is probably not catching that the `apc` array is a variable size array, so the "c_decoded_fields" needs to be set to indicate how many elements are in the `apc` array.

Davinco:
> @Davinco I think the json2test script is probably not catching that the `apc` array is a variable size array, so the "c_decoded_fields" needs to be set to indicate how many elements are in the `apc` array.

Ah right, thanks for the help @silverjam

silverjam (approved):

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -751,6 +753,129 @@ definitions:
...
+           type: u16
+           desc: >
+             Number of steps in the longitude direction.
+             
+             
+             See GNSS-SSR-ArrayOfCorrectionPoints field numberOfStepsLongitude.
+       - bitmask:
+           type: u64
+           desc: >
+             Specifies the availability of correction data at the
```
silverjam, line 862: Need to see if we can make this documentation more succinct without reducing clarity to make things a little easier on the PDF, right now we're overflowing by a few lines on the end of the table describing these fields:

image

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -537,7 +537,7 @@ definitions:
            size_fn: n_sats
            desc: Array of STEC polynomial coefficients and its bounds for each space vehicle.
 
- - MSG_SSR_TILE_DEFINITION_DEP:
+ - MSG_SSR_TILE_DEFINITION_DEP_A:
```
Davinco, line 540: @silverjam I'm a bit uncertain about this change. Since this message now will consist of two deprecated versions, I added a suffix to the previous deprecated version but then that name doesn't exist anymore, which may create a compatibility issue? Maybe it's more correct to leave the first deprecated message unchanged and then add a suffix to the current tile definition so the message versions become:
```
MSG_SSR_TILE_DEFINITION_DEP
MSG_SSR_TILE_DEFINITION_DEP_A
MSG_SSR_TILE_DEFINITION
```
instead of 
```
MSG_SSR_TILE_DEFINITION_DEP_A
MSG_SSR_TILE_DEFINITION_DEP_B
MSG_SSR_TILE_DEFINITION
```

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -813,15 +938,43 @@ definitions:
...
+           type: u8
+           desc: >
+             IOD of the SSR correction. A change of Issue Of Data
+             SSR is used to indicate a change in the SSR
+             generating configuration
+       - apc:
+           type: array
+           fill: SatelliteAPC
+           desc: Satellite antenna phase center corrections
+
```
Davinco, line 977: @theureau also here

spec/yaml/swiftnav/sbp/ssr.yaml
silverjam: @Davinco Either way is fine, we should go with whatever precedent we already have with existing messages, if there isn't a consistent precedent then we can pick whatever makes the most logic sense.  The main backwards compatibility that we want to guard against is wire/format compatibility, backwards compat with compilation is nice, but not strictly necessary.  Wire compatibility means: the size and format of bytes in the message cannot change, for bit-field messages we can add new bit-fields and deprecate old ones and in some cases we can change the interpretation of bytes (e.g. from unsignged to signed) but these cases need to be considered carefully.

Davinco: nice, thanks for the clarification :raised_hands:

spec/yaml/swiftnav/sbp/ssr.yaml
```diff
@@ -751,6 +753,129 @@ definitions:
...
+             Update interval between consecutive corrections. Encoded
+             following RTCM DF391 specification.
+       - sol_id:
+           type: u8
+           desc: SSR Solution ID. Similar to RTCM DF415.
+       - iod_ssr:
+           type: u8
+           desc: >
+             IOD of the SSR correction. A change of Issue Of Data
+             SSR is used to indicate a change in the SSR
```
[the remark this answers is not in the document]
theureau: @Davinco , for the field `iod_ssr`, may be it should be better to name it the same way than in the other atmo message `iod_atmo` and se the same description `IOD of the SSR atmospheric correction`

spec/yaml/swiftnav/sbp/ssr.yaml
theureau: Ok for me !

spec/yaml/swiftnav/sbp/ssr.yaml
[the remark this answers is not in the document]
Davinco: Ok, great! I'll change to that :+1:

PR #1285, opened 2023-01-10
Add per SV telemetry message [POS-905]

ismolyakov:
# Description

@swift-nav/devinfra

# API compatibility

## API compatibility plan

# JIRA Reference


spec/yaml/swiftnav/sbp/telemetry.yaml
```diff
@@ -0,0 +1,103 @@
...
+      id: 0x0120
+      short_desc: Per-signal telemetry
+      desc: >
+        This message includes telemetry pertinent to satellite signals available to Starling.
+      fields:
+        - tow:
+            type: u32
+            units: ms
+            desc: GPS Time of Week
+        - n_obs:
```
silverjam, line 85: Do we need this field?  It doesn't look like we are doing anything with obs in the message.

silverjam:
PDF might need some tweaks, there's a cut-off in one of the descriptions

image

---

Other tables look good though:

image
image

silverjam (commented):
Addition of new telemetry module looks correct, PDF documentation might need some tweaks though

PR #1287, opened 2023-01-16
Add test for SBP-1534 STEC list empty

JADC362:
# Description

@swift-nav/devinfra

Add a MsgSsrGriddedCorrectionBounds msg (SBP-1534) test, that checks if an empty STEC list can be encoded / decoded.

Task [PC-313](https://swift-nav.atlassian.net/browse/PC-313)

# API compatibility

It doesn't add compatibility risk. It increases the coverage for a test case.

[PC-313]: https://swift-nav.atlassian.net/browse/PC-313?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


silverjam (approved):

spec/tests/yaml/swiftnav/sbp/ssr/test_MsgSsrGriddedCorrectionBounds.yaml
```diff
@@ -68,3 +68,46 @@ tests:
...
       crc: '0x5277'
+
+  - msg:
+      fields:
+        header:
+          time:
+            tow: 180
+            wn: 3
+          num_msgs: 1
+          seq_num: 1
```
jtec, line 79: I think the sequence number might be zero-based.

JADC362: Mmmm are you sure? I set this field to 1, since the first (original) CorrectionBounds test has 1. I look the SBP documentation and it's not clear if the `seq_num` starts from 0 or 1.

PR #1292, opened 2023-01-27
new rust clippy

adrian-kong:
fixes https://github.com/swift-nav/libsbp/actions/runs/4020003836/jobs/6907469269


pcrumley (changes requested):
did you just run `cargo clippy --fix`?

because it seems like if that was supposed to be it i don't know why it added some bin files.
Screenshot 2023-01-26 at 4 23 55 PM


Seems like there are others too.

i also don't know why a `.rs` file would be a binary.

adrian-kong:
not sure why its a BIN file but all that was changed was the generator file (to enforce `make gen-rust` have same clippys) and `cargo clippy --all-targets --all-features --fix`

pcrumley:
ok checked out the commit locally. Those files are being marked as binary because they contain raw null bytes as strings

pcrumley:
Screenshot 2023-01-26 at 5 17 44 PM

Those files are being marked as binary because they contain raw null bytes in them. It also is occuring in one of the places where the lint was being hit. I think we should fix it and have the null bytes be written as a binary string. i.e. in this case it should be:
```rust
     b"idle\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
```

Let's do that in a different PR. At any rate this code could have been auto-gen'd in so we might have to update it there too.

pcrumley (commented):
So because of the binary files, I can't really approve this code. I pulled it down locally and ran `git diff` on the files and it also just says the file changed.

I am guessing these are auto-gen'd and totally fine to accept but I think someone else who has more context about what can and can't change should review this. I don't feel comfortable approving it.

@silverjam @notoriaga See discussion above

notoriaga:
> So because of the binary files, I can't really approve this code. I pulled it down locally and ran `git diff` on the files and it also just says the file changed.
> 
> I am guessing these are auto-gen'd and totally fine to accept but I think someone else who has more context about what can and can't change should review this. I don't feel comfortable approving it.
> 
> @silverjam @notoriaga See discussion above

This should get rid of the null bytes - https://github.com/swift-nav/libsbp/pull/1293 still shows up as bin for me. Maybe takes a bit to update or has to be merged into main

notoriaga (approved):

pcrumley:
@notoriaga Thanks for fixing that null bytes issue so quickly, i know it was a bit nitpicky but i think it is better this way.

I think the reason it is showing it as a binary file is because the diff against main still shows the non-printable characters. I am not a git expert, but i think we could fix it with a `.gitattributes` file.

pcrumley (approved):

pcrumley:
see this pr which would render even the old files with the null bytes: https://github.com/swift-nav/libsbp/pull/1294

PR #1306, opened 2023-03-16
variable length ecdsa signature [GV2-193]

notoriaga:
# Description

@swift-nav/devinfra

Makes the signature on MSG_ECDSA_SIGNATURE variable length

# API compatibility

This breaks the existing message but it is currently unused

# JIRA Reference

https://swift-nav.atlassian.net/browse/GV2-193


spec/yaml/swiftnav/sbp/signing.yaml
```diff
@@ -154,12 +154,16 @@ definitions:
...
             desc: The last 4 bytes of the certificate's SHA-1 fingerprint
+        - n_signature_bytes:
+            type: u8
+            desc: >
+              Length of the signature in bytes.
         - signature:
             type: array
             fill: u8
-            size: 64
+            size_fn: n_signature_bytes
```
silverjam, line 164: I think this needs to be a fixed size of 73 bytes, the `n_signature_bytes` field will dictate how many bytes of the signature are valid and/or should be used.

notoriaga: Ahh good catch

spec/yaml/swiftnav/sbp/signing.yaml
```diff
@@ -154,12 +154,16 @@ definitions:
...
             desc: The last 4 bytes of the certificate's SHA-1 fingerprint
+        - n_signature_bytes:
+            type: u8
+            desc: >
+              Number of bytes to use of the signature field.
         - signature:
             type: array
             fill: u8
-            size: 64
+            size: 73
```
silverjam, line 164: 72 is the maximum from the research I did, but we can leave it at 73 if you'd like to leave some wiggle room -- see https://snav.slack.com/archives/C04RCE6GWRH/p1678904497727939

spec/yaml/swiftnav/sbp/signing.yaml
```diff
@@ -154,12 +154,16 @@ definitions:
             fill: u8
             size: 4
             desc: The last 4 bytes of the certificate's SHA-1 fingerprint
+        - n_signature_bytes:
```
woodfell, line 157: This is changing the structure of an existing message by inserting a new field part way through rather than at the end. Shouldn't this require a bump to the message ID? Deprecate the old one to maintain backwards compatibility?

notoriaga: @Blast545 and I are the only consumers of the message at the moment so it should be okay

Blast545: Yeah, we had a meeting to talk about this and it shouldn't be any issues changing this.

Blast545 (approved):

silverjam (changes requested):
Sorry for changing my mind on this... but every time we try to sneak one of these in we end up regretting it... so maybe we shouldn't try again.  One risk is we have already published tooling that will break when it sees the new version of the message -- no new `sbp2json` has been published for example, but a new `rtcm32json` has (which now supports decoding SBP) see here: https://github.com/swift-nav/gnss-converters-private/commit/edb13b9e173e933a59d1dcdbe8045b5acaa6e406 (and this was cut after the libsbp submodule was bumped).

notoriaga:
> Sorry for changing my mind on this... but every time we try to sneak one of these in we end up regretting it... so maybe we shouldn't try again. One risk is we have already published tooling that will break when it sees the new version of the message -- no new `sbp2json` has been published for example, but a new `rtcm32json` has (which now supports decoding SBP) see here: [swift-nav/gnss-converters-private@edb13b9](https://github.com/swift-nav/gnss-converters-private/commit/edb13b9e173e933a59d1dcdbe8045b5acaa6e406) (and this was cut after the libsbp submodule was bumped).

No worries, I switched to a new message

spec/yaml/swiftnav/sbp/signing.yaml
```diff
@@ -120,6 +120,63 @@ definitions:
...
+              value of the counter may not initially be zero.
+        - certificate_id:
+            type: array
+            fill: u8
+            size: 4
+            desc: The last 4 bytes of the certificate's SHA-1 fingerprint
+        - n_signature_bytes:
+            type: u8
+            desc: >
+              Number of bytes to use of the signature field.
```
silverjam, line 160:
```suggestion
              Number of bytes to use of the signature field.  The DER encoded
              signature has a maximum size of 72 bytes but can vary between 70
              and 72 bytes in length. 
```

c/test/cpp/auto_check_sbp_signing_MsgEcdsaSignatureDep.cc
```diff
@@ -0,0 +1,1389 @@
+/*
+ * Copyright (C) 2015-2021 Swift Navigation Inc.
```
silverjam, line 2: Looks like we might be missing the new test files for C and Java that cover the new message ID.

silverjam (approved):

PR #1307, opened 2023-03-17
fix signature in MsgCertificateChain [GV2-193] 

notoriaga:
# Description

@swift-nav/devinfra

Fixes the signature field of MsgCertificateChain. Also makes a separate type for the signature because we use in it multiple places

# API compatibility

New messages were made for MsgCertificateChain and MsgEcdsaSignature


spec/yaml/swiftnav/sbp/signing.yaml
```diff
@@ -84,6 +101,40 @@ definitions:
...
+        - corrections_certificate:
+            type: array
+            fill: u8
+            size: 20
+            desc: SHA-1 fingerprint of the corrections certificate
+        - expiration:
+            type: UtcTime
+            desc: >
+              The certificate chain comprised of three fingerprints: root
+              certificate, intermediate certificate and corrections certificate.
```
silverjam, line 126:
```suggestion
              The time after which the signature given is no longer valid.
              Implementors should consult a time source (such as GNSS) to
              check if the current time is later than the expiration time,
              if the condition is true, signatures in the stream should not
              be considered valid.              
```

spec/yaml/swiftnav/sbp/signing.yaml
```diff
@@ -51,6 +51,23 @@ definitions:
             desc: nanoseconds of second
                   (range 0-999999999)
 
+  - ECDSASignature:
+      short_desc: ECDSA signature
+      fields:
+        - len:
```
silverjam, line 57: I think if we move `len` field to after `data` then we technically would not need to rev `MSG_ECDSA_SIGNATURE` since the wire format/order/length of the bytes would need to be the same, but not sure that the message checker in CI would be able to handle that.

silverjam:
FWIW I think revving the messages is fine to fix this, I think once all of these `signing` messages are solidified we should "elide" the messages so there isn't so much cruft in the YAML spec.

E.g.

```
  - MSG_SIGNING_RESERVED_0:
      id: 0x0C08
      short_desc: Reserved
      desc: Reserved
      fields:
        - reserved:
            type: array
            fill: u8
            desc: Rererved
```

silverjam (approved):

PR #1312, opened 2023-03-29
Parse error falls back to Invalid

adrian-kong:
# Description

This PR makes it so that if at any point in the parsing of sbp messages the parsing fails, it has the option of falling back to an InvalidMessage which can then be written to the output. The way this PR acheives this goal is through some major breaking changes. The idea is that when errors are thrown throughout the process, we store all that local information in the constructed error. That way, choosing how to write the output can become as simple as matching on the thrown error and then basically writing `SbpMessage::from(error)`. This enables us to keep the previous behavior of skipping errors or returning on errors simply by refusing to handle the errors.

`sbp2json`, `json2sbp` & `json2json` now have a new optional arg `--error_handler` which allows the following options `skip`, `return` or `ToInvalid`. that skip and return are the equivalent of the false and true passed to fatal-errors arg which as been removed.

Care was taken to remove some panics in the parsing of SBP which could cause some types of invalid messages to be lost.

Finally this PR does not do things like try to recover from being passed in malformed json or given an I/O error. In case of those error, if the `ToInvalid` option is passed, the code will panic. 

Some issues with this PR --
1) it has a lot of breaking changes, and it kind of does a lot. Just want to make sure people like the direction i went in.
2) the lack of recovering from malformed json means I left json2json to panic if given ToInvalid as an option.
3) It has no unit tests. I wanted some eyes on this code and proposed solution before totally getting it over the finish line.

# JIRA Reference

https://swift-nav.atlassian.net/browse/DEVINFRA-1220


rust/sbp/src/de.rs
```diff
@@ -303,11 +306,14 @@ impl Frame {
 
     pub fn to_sbp(&self) -> Result<Sbp, Error> {
```
notoriaga, line 307: Can probably change this to `Result<Sbp, CrcError>`, I think that's the only thing that can go wrong now

rust/sbp/src/de.rs
```diff
@@ -303,11 +306,14 @@ impl Frame {
...
         self.check_crc()?;
-        Ok(Sbp::from_parts(
-            self.msg_type(),
-            self.sender_id(),
-            self.payload(),
-        )?)
+        let sender_id = self.sender_id();
+        let mut payload = self.payload();
+        let msg = Sbp::from_parts(self.msg_type(), sender_id, payload).or_else(|_| {
```
notoriaga, line 311: I think this would mask malformed messages though? Like if the message type is a known message but the payload is too short. Then you'd have an `Unknown` message with a message type that corresponds to a known message which is a little odd. But maybe that's what we want, I guess it's a trade off between being lenient and surfacing errors better

adrian-kong: yea... think we're just treating Unknown as the `Frame` at this point to serialize as json :/

although that makes sense for the purpose of the Unknown message i guess.

maybe a better compromise would to just add a flag like `strict` to address dropping payload errors?

issue with the raw decoder or persisting broken data is where do we actually trim the unnecessary data - which thought everything would propagate up to sbp converters and they would drop those - or maybe we just never drop them 😱

notoriaga: I think the `Decoder` should only output valid SBP (as in CRC is good and the payload is correct), if a user wants access to potentially invalid messages they can go with the `Framer` I think. So for example `sbp2json` should probably be doing something like

```rust
for frame in Framer::new(rdr) {
    if let Ok(msg) = frame.to_sbp() { 
        // valid crc/payload, possibly an Unknown message
        json_encoder.write(&msg)?;
    } else {
        // invalid crc or payload, could also be an Unknown message
        // need some json representation of a frame
        json_encoder.write(&frame)?;
    }
} 
```

rust/sbp2json/src/lib.rs
```diff
@@ -78,7 +79,13 @@ where
     W: Write,
     F: Formatter + Clone,
 {
-    let source = maybe_fatal_errors(sbp::iter_messages(input), fatal_errors);
+    let source = maybe_fatal_errors(sbp::iter_frames(input), fatal_errors).map(|frame| {
+        frame.to_sbp().unwrap_or(Sbp::Unknown(Unknown {
```
notoriaga, line 83: I would probably implement Serialize/Deserialize for Frame instead of using Unknown for the case when the frame is invalid so we can distinguish between an Unknown message type and in invalid message. It should look the same (as in serialize/deserialize to a struct with msg_type, sender_id and payload fields) but then you can tell whether or not its an issue with an outdated version of libsbp or a malformed/corrupted message.

Also you'll need to handle the other direction for this is json2sbp

rust/sbp2json/src/lib.rs
```diff
@@ -78,7 +79,13 @@ where
...
+            msg_id: 0,
+            sender_id: None,
```
notoriaga, line 85: you can set these two fields with `frame.msg_type()` and `frame.sender_id()`

rust/sbp2json/src/lib.rs
notoriaga: although I guess we are using an untagged enum so you wouldn't be able to tell the difference between a Frame and an Unknown hmmm

rust/sbp2json/src/lib.rs
adrian-kong: hmmm how would we distinguish if the msg is an unknown if we pass the msg type as well
would just be a secret corrupt msg when parsing but just looks like its not corrupt 🤔

rust/sbp2json/src/lib.rs
adrian-kong: hmmmm, is unknown actually used for anything tho
could just use unknown for the types of msg that we dont known about

this would make it so we dont need to do anything to json2sbp since the msgs they produce would just be kinda the payload except wrapped with unknown 🤔 

unless we want exact raw corrupt packet without wrapping anything then might be a bit more complicated than this.

adrian-kong: @notoriaga i.e. would it be ok to simply wrap around as an unknown message (since these msgs are indeed 'unknown') or do we want some special form of deserialization in JSON for corrupt messages in addition to the sbp?
i.e.
```json
{
    msg_type: ...
    payload: ...
    sender_id: ....
}
{
    corrupt_data: ...
}
```
or simply just handling corrupt_data inside Unknown packets like
```json
{
    msg_type: 0
    sender_id: 0
    payload: the corrupt msg
}
```

@jbangelo @silverjam

notoriaga: We could also add a msg_name field to the json representation of messages so then we'd have 

```
{
    "msg_type": 123,
    "msg_name": Unknown,
    "sender_id": 4,
    "payload": [...]
}
{
    "msg_type": 321,
    "msg_name": Frame,
    "sender_id": 4,
    "payload": [...]
}
```

could also be nice to have for jq stuff, the message names are bit easier to remember than the ids

adrian-kong: think the message names are alread yhtere for sbp2json, making this field necessary would mean json2sbp only uses this field to check specific case where `msg_name == "FRAME"` 🤔 not sure if theres any better way or...
^^ this also seems much more complicated / invovled tho #1317 WIP
hmmmm....

silverjam:
Changes I pushed make the JSON output "look" better for random garbage data:

```
{"crc":44120,"length":132,"msg_type":17681,"msg_name":"UNKNOWN","payload":"QhFF87pNdUgVvuMBbQj/naRyVCUVbhxuPfguaBx5/tgK2rImLA7g74QTHUg+mRdgUCaUL5PnZUtAO+Gumc0xt1OQwVQpsBhSZcbOPMJlImu4a7j33UA9DhL/X1bbS7nKYwjZISrhWcJM+WMHakvH4bOELQ5igBFSAWPWF6mIi2Vdcsot","preamble":85,"sender":47859}
{"crc":51746,"length":116,"msg_type":32385,"msg_name":"UNKNOWN","payload":"JIF+/Uq4ENNpq+uECTleTOXMbPgZ2w0SWeUh4fs6XzumFAgA2kI3vCduAcdeOe35PfI1ejoSkKJoZ9r/lKujjxbUH1yIX73GO0/DGRw/WpHMwUi0y1y/CdbTcxx50ZAZqrh3vOfeUVCb/g5bpwTxEX1x3bA=","preamble":85,"sender":19197}
{"crc":56169,"length":125,"msg_type":35244,"msg_name":"UNKNOWN","payload":"PKyJ6OgEII2cFCgtjjAz+fNOCt3ikYjqkXzgj8sMqslc2pszXn5j4+ZK66zfkf3JKDIE/b80v+iVCZVSL+V0FBBHTAa8z9v7sG1EzFpQvg8ISDdkvLYtcjgqzCZCACHbmRWeHz0yV3hUkJjUH25neHYWDGTWtK91CGrE13A=","preamble":85,"sender":59624}
```

But tests are having issues and the inverse `json2sbp` doesn't work on this new output:
```
data did not match any variant of untagged enum JsonInput
data did not match any variant of untagged enum JsonInput
error parsing payload
error parsing payload
error parsing payload
error parsing payload
error parsing payload
data did not match any variant of untagged enum JsonInput
error parsing payload
```

silverjam:
Converted to draft to remove from review list until we can make more progress on this.

pcrumley:
@silverjam I'm picking this up now, but i don't know what you mean by inverse `json2sbp`. can you explain that workflow to me? I'd like to capture it as a unit test if possible
> But tests are having issues and the inverse `json2sbp` doesn't work on this new output:

silverjam:
> @silverjam I'm picking this up now, but i don't know what you mean by inverse `json2sbp`. can you explain that workflow to me? I'd like to capture it as a unit test if possible
> 
> > But tests are having issues and the inverse `json2sbp` doesn't work on this new output:

I think I just meant that `json2sbp` isn't handling the output from `sbp2json` anymore.  The main idea with this PR is we wanted `sbp2json | json2sbp` to be "idempotent" even in the face of bad/corrupted data.  I think we've figured out a reasonable way to represent corrupted data in the JSON output, but now need to do the work to get `json2sbp` to understand it.

rust/sbp/src/messages/invalid.rs
```diff
@@ -0,0 +1,131 @@
+//! Invalid messages, maybe don't have enough bytes are Crc Errors.
+
+use bytes::{Buf, BufMut};
+
+use crate::{de::CrcError, messages::SbpMsgParseError, wire_format::WireFormat, SbpMessage};
+
+/// The message returned by the parser when the message type does not correspond to a known message.
```
silverjam, line 7: Or if the CRC is invalid?

pcrumley: I forgot to finish up this module's documentation. let me re-write it rq.

rust/sbp/src/messages/mod.rs
```diff
@@ -8,6 +8,11 @@
 // EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
 // WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 //! SBP message definitions.
+
+//****************************************************************************
+// Automatically generated from yaml/
```
adrian-kong, line 13: really tiny thing, dont think this got generated correctly, `yaml/(((filepath)))` should maybe be `generator/sbpg/targets/resources/rust/sbp_messages_mod.rs`

adrian-kong (commented):
looks good to me @pcrumley 👍

rust/sbp/src/messages/mod.rs
pcrumley: thanks for pointing this out

rust/sbp2json/src/lib.rs
```diff
@@ -2,18 +2,104 @@ use std::io::{Read, Write};
...
+    for ErrorHandler<'_, M, E>
+{
+    type Item = M;
+
+    fn next(&mut self) -> Option<M> {
+        match (self.opts, self.inner_iter.next()) {
+            (_, None) => None,
+            (_, Some(Ok(x))) => Some(x),
+            (ErrorHandlerOptions::ReturnOnFirstError, Some(Err(e))) => {
+                eprintln!("{e}");
```
silverjam, line 77: Are these eprintln's here for debugging purposes, do they need remain here long term?

silverjam: It looks like this maybe replaces `maybe_fatal_errors`?

silverjam: I was thinking this might be a global impl, but since it's limited to the sbp2json library then have `eprintln` in the implementation probably isn't an issue.

rust/sbp2json/src/bin/sbp2json.rs
```diff
@@ -32,10 +32,10 @@ pub struct Options {
     /// Buffer output before flushing
     #[clap(short, long)]
     buffered: bool,
-
-    /// Stop on first error encountered
-    #[clap(long)]
-    fatal_errors: bool,
+    ///
+    /// what to do if encounter an error while parsing.
```
silverjam, line 36: This should follow the convention of the other docstrings, also I'm not sure how the leading whitespace will show up when this is rendered as a help string for the command line options:

```suggestion
    /// What to do if encounter an error while parsing
```

rust/sbp2json/src/bin/json2sbp.rs
```diff
@@ -28,9 +28,9 @@ struct Options {
     #[clap(short, long)]
     buffered: bool,
 
-    /// Stop on first error encountered
-    #[clap(long)]
-    fatal_errors: bool,
+    /// what to do if encounter an error while parsing.
```
silverjam, line 31:
```suggestion
    /// What to do if encounter an error while parsing
```

rust/sbp2json/src/bin/json2json.rs
```diff
@@ -33,9 +33,9 @@ struct Options {
     #[clap(short, long)]
     buffered: bool,
 
-    /// Stop on first error encountered
-    #[clap(long)]
-    fatal_errors: bool,
+    /// what to do if encounter an error while parsing.
```
silverjam, line 36:
```suggestion
    /// What to do if encounter an error while parsing
```

generator/sbpg/targets/resources/rust/sbp_messages_mod.rs
```diff
@@ -89,6 +99,8 @@ pub trait SbpMessage: WireFormat + Clone + Sized {
     fn friendly_name(&self) -> &'static str {
         ""
     }
+    /// Tells you if the message is invalid and may need to be special cased at certain points.
+    fn is_invalid(&self) -> bool;
```
silverjam, line 103: I hate to ask, but is it possible to rephase this as `is_valid` so that we don't have the double negative of `! is_invalid()` that means "valid"?

rust/sbp/src/json/mod.rs
```diff
@@ -123,18 +124,49 @@ impl Formatter for HaskellishFloatFormatter {
...
+            }
+
+            Self::Base64Error(e) => {
+                panic!("Unable to recover from Base64 decoding error {}", e)
+            }
+
+            Self::SerdeJsonError(e) => {
+                // We may need to handle this error path if we want to recover
+                // from errors with Json2Json
+                panic!("Unable to recover from SerdeJsonError {}", e)
```
silverjam, line 154: FWIW, the original use case of `json2json` was to expand "elided" JSON data that was output from the old Swift Console when it was logging (the data output only included the base64 message payload and a limited number of common fields, this was done because it was too slow to dump the message and it's constituent fields).  The modern console is fast enough that it's able to dump the "unelided" form of JSON.

rust/sbp/src/messages/invalid.rs
```diff
@@ -0,0 +1,143 @@
...
+//! it is missing some metadata like msg_type or similar.
+//! 2) The message is invalid because the payload is not large enough and cannot be
+//! parsed into a message. This is the SbpMsgParseError.
+
+use bytes::{Buf, BufMut};
+
+use crate::{de::CrcError, messages::SbpMsgParseError, wire_format::WireFormat, SbpMessage};
+
+/// Invalid messages occur when either the frame or message payload doesn't have enough bytes or
+/// the Crc does not match the messages payload. If the message is well formed with a message
```
silverjam, line 13:
```suggestion
/// the CRC does not match the messages payload. If the message is well formed with a message
```

rust/sbp/src/messages/invalid.rs
```diff
@@ -0,0 +1,143 @@
...
+/// id and crc and payload that all are consistent, but its message type is unknown, it should be
```
silverjam, line 14:
```suggestion
/// id and CRC and payload that all are consistent, but its message type is unknown, it should be
```

rust/sbp/src/messages/invalid.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// crc, msg_id & sender_id, if any
+    pub payload: Vec<u8>,
+}
+
+impl SbpMessage for Invalid {
+    fn message_name(&self) -> &'static str {
+        "INVALID"
+    }
+
+    fn message_type(&self) -> u16 {
```
silverjam, line 53: I think 0xFFFF is a valid message number (MSG_HEARBEAT I think?) -- ideally this iterface would return an `Option` or panic but I'm assuming there's something that makes that sort of change inconvenient?

silverjam: If we're trying to avoid a major version bump of the library I think this sort of change is important enough that it justifies a breaking change (and a major version bump).

rust/sbp/src/wire_format.rs
```diff
@@ -178,7 +180,10 @@ impl WireFormat for f64 {
 }
 
 #[derive(Debug, Clone)]
-pub struct PayloadParseError {}
+pub struct PayloadParseError {
+    /// A vec that just contains all the bytes
+    pub raw_invalid_payload_bytes: Vec<u8>,
```
silverjam, line 185: Variable name is a mouthful, would this work?

```suggestion
    pub raw_bytes: Vec<u8>,
```

rust/sbp/src/messages/invalid.rs
pcrumley: I don't want to panic here. I think I will just make another breaking change as this pr already does that anyway.

silverjam: 👍

rust/sbp2json/src/bin/sbp2json.rs
pcrumley: nice catch.

rust/sbp2json/src/lib.rs
pcrumley: the `eprintln!` here was basically just maintaining the same behavior from `maybe_fatal_errors!` as it prints all errors to stderr. I don't believe i added any print statements in the lib but i will check now.
```rust
fn maybe_fatal_errors<'a, M, I, E>(
    messages: I,
    fatal_errors: bool,
) -> Box<dyn Iterator<Item = M> + 'a>
where
    M: 'a,
    I: Iterator<Item = std::result::Result<M, E>> + 'a,
    E: std::error::Error + 'a,
{
    if fatal_errors {
        Box::new(
            messages
                .take_while(|m| m.as_ref().map_err(|e| eprintln!("{e}")).is_ok())
                .map(|m| m.unwrap()),
        )
    } else {
        Box::new(messages.filter_map(|m| m.map_err(|e| eprintln!("{e}")).ok()))
    }
}
```

rust/sbp/src/wire_format.rs
pcrumley: done

pcrumley:
@silverjam 

I'm mostly happy with the PR now. There is only one question i have, which is that i'm converting the invalid_frame to base64 regardless of if it is going to be serialized into a json. Inside of serde ecosystem it should "just work", and for that matter i believe it only ever use serde to go to json within the lib, but tbh i don't love the solution i took.

see `rust/sbp/src/messages/invalid.rs` for what i am talking about

silverjam:
> @silverjam 
> 
> 
> 
> I'm mostly happy with the PR now. There is only one question i have, which is that i'm converting the invalid_frame to base64 regardless of if it is going to be serialized into a json. Inside of serde ecosystem it should "just work", and for that matter i believe it only ever use serde to go to json within the lib, but tbh i don't love the solution i took.
> 
> 
> 
> see `rust/sbp/src/messages/invalid.rs` for what i am talking about

Yeah, I don't love it but I'm ok to leave it if we can't think of any concrete problems with it.

In terms of preference though some "semantic wrapping" around the string rather than just throwing out a base64 string would be nice.  That is, just wrapping the string with a field like "payload" (that we already use) so that some of the semantic information is maintained.

In the ideal case I think we would have some sort of `IntoCommonJson` trait and have CommonJson be a collector of common fields for the JSON format.  This would allow the serde serialization path to only have to worry about "well formed" messages/frames.

In general I think there's several distinct ideas/concepts that we're trying to wrangle here:
- `serde` streams of objects, these are annotated collections of fields that are compatible with the serde data model-- in some areas it seems we're mostly concerned with serde objects that are well formed and from known SBP messages.  This allows for the cleanest representation as a serde object and nothing from the framing or message encoding layer bleeds through.  Concepts like "payload", message number and CRC don't fit well into this model (in some cases SBP messages have binary data so they don't fit cleanly into this mapping either, but those are special cases). Fields in this layer would be all the named fields and substructures that make up a message.
- Then we have the SBP "message metadata" layer of what we're modeling. That is: the length of the message, the message number, payload, the CRC, etc-- these are well formed messages, but different from the previous layer, "unknown" messages are representable in this layer. Fields in this layer would be "message number", "crc", "sender", "length" and "payload".
- Finally there's the framing layer, where we have just the framing around the encoded messages, here we can model just invalid vs valid frames.  There's a lot of crossover with the fields of the "metadata" layer, but invalid frames need some way to signal that they're bad, so picking something like "drop all the fields except the payload" seems like a clean signal that the other fields aren't valid (that is, because the message was not well formed or the CRC was bad).

So I would expect the final output JSON to look something like this to satisfy the above constraints:

    Known: {msg num, msg name: ..., sender, length, crc, payload, fields}

    Unknown: {msg num, msg name: UNKNOWN, sender, length, crc, payload}

    Invalid: {msg name: INVALID, payload}

(NB: I said I didn't really care how we signal that a frame is invalid, so this exact representation isn't necessary but I think we're already doing something like this in this PR).

The choices we make to accomplish this really fall into a few categories for me:
- How much of the SBP message are we going to lift into the Rust type system?
- What data should a Rust message struct carry around with it (e.g. a message struct probably doesn't need to carry around a copy of the sender id and payload, this can be left in the frame object)?
- How do we make the appropriate trade offs in layering (as described above) for the serialization process vs compromising the "purity" of the design so that the output format satisfies our design constraints without being a pain to implement?

Hopefully this gives a good framework for how to think about the changes in this PR... maybe should have worked through this thinking earlier... but I think we're pretty much "there" with these changes.

Again, not sure exactly that we need to change anything, so maybe worthwhile for us to talk through these changes again at some point.

silverjam:
Looks like we might need to bump the compiler version and MSRV that we're using or remove these let-else blocks:

```
error[E0658]: `let...else` statements are unstable
  --> sbp/src/ser.rs:92:5
   |
92 | /     let Some(msg_type) = msg.message_type() else {
93 | |         // this should be unreachable because things
94 | |         // without message ids should be caught by invalid
95 | |         // messages
96 | |         return Err(WriteFrameError::NoMessageType);
97 | |     };
   | |______^
   |
   = note: see issue #87335 <https://github.com/rust-lang/rust/issues/87335> for more information

error[E0658]: `let...else` statements are unstable
  --> sbp/src/ser.rs:86:5
   |
86 | /     let Some(sender_id) = msg.sender_id() else {
87 | |         // this should be unreachable because things
88 | |         // without sender ids should be written as invalid
89 | |         // messages
90 | |         return Err(WriteFrameError::NoSenderId);
91 | |     };
   | |______^
   |
   = note: see issue #87335 <https://github.com/rust-lang/rust/issues/87335> for more information
```

pcrumley:
I am very close to getting this within spec, it just seems like a CRC is added somewhere, e.g this is what 
`./target/release/sbp2json --error-handler ToInvalid ~/Downloads/corrupted.sbp > out`
write invalid messages as:
```json
{"msg_name":"INVALID","payload":"VQoCw5UibB1EFL/0ELSA4EJAzKTjmU2eXsCYqanZWkCiAAwBDwl23lUO","crc":3669}
```

I'm not sure where that extra crc is coming from. I also removed a few other places where panics could happen by replacing `read_u16_le()` with something that tries to cast the slice to a 2 byte array and then calls `u16::from_bytes_le`.

silverjam:
Seeing some unexpected results when just using garbage data, these might be "expected" given the current parsing behavior, but not what I would expect to happen.

If I test with the `short.sbp` I see INVALID messages being surfaced, which is good:

```
❯ cargo run --bin sbp2json --quiet -- --error-handler to-invalid <test_data/short.sbp 2>/dev/null | grep INVALID | wc -l
9
```

However, if I start by creating a bit of garbage data:

```
❯ dd if=/dev/urandom count=1 bs=64 of=corrupt.bin
1+0 records in
1+0 records out
64 bytes copied, 7.3007e-05 s, 877 kB/s
```

Contents of the corrupt blob:

```
❯ od -tx1 corrupt.bin
0000000 0a 28 61 b2 99 ba d6 1d 42 15 f7 9f 36 f8 ca 72
0000020 97 41 af 29 b9 0b f1 0e d3 1d ef ab d4 70 ac 1e
0000040 02 71 a0 59 c1 78 95 5d f9 e5 f9 00 6a 38 06 69
0000060 06 8d f1 80 a3 e2 a1 3c 6b ab 42 e8 18 0a f0 70
0000100
```

Then attempt to decode that I don't get an invalid message:
```
❯ cargo run --bin sbp2json --quiet -- --error-handler to-invalid <corrupt.bin | grep INVALID | wc -l
0
```

Putting this data at the start and end yields similar results:

```
❯ cat corrupt.bin test_data/roundtrip.sbp >corrupt_start.bin

❯ cargo run --bin sbp2json --quiet -- --error-handler to-invalid <corrupt_start.bin | grep INVALID | wc -l
0

❯ cat test_data/roundtrip.sbp corrupt.bin >corrupt_end.bin

❯ cargo run --bin sbp2json --quiet -- --error-handler to-invalid <corrupt_end.bin | grep INVALID | wc -l
0
```

Similar results for the middle:

```
❯ cargo run --bin sbp2json --quiet -- --error-handler to-invalid <test_data/roundtrip.sbp | head -2500 >head.json

❯ cargo run --bin sbp2json --quiet -- --error-handler to-invalid <test_data/roundtrip.sbp | tail -n +2501 >tail.json

❯ wc -l head.json -l tail.json
   2500 head.json
   2500 tail.json
   5000 total

❯ cargo run --bin json2sbp --quiet <tail.json >tail.bin

❯ cargo run --bin json2sbp --quiet <head.json >head.bin

❯ cat head.bin tail.bin | cargo run --quiet --bin sbp2json | wc -l
5000

❯ cat head.bin corrupt.bin tail.bin >corrupt_middle.bin

❯ cargo run --bin sbp2json --quiet -- --error-handler to-invalid <corrupt_middle.bin | grep INVALID | wc -l
0
```

If I use a corrupt blob that happens to start with an SBP preamble byte I get better results.  But only in certain cases.

To create the bad blob, you can run `xxd` in reverse:

```
xxd corrupt.bin >corrupt.txt
nvim corrupt.txt # change start byte to 55
xxd -r <corrupt.txt >corrupt_sbp_start.bin
```

This should yield:

```
❯ od -tx1 corrupt_sbp_start.bin
0000000 55 28 61 b2 99 ba d6 1d 42 15 f7 9f 36 f8 ca 72
0000020 97 41 af 29 b9 0b f1 0e d3 1d ef ab d4 70 ac 1e
0000040 02 71 a0 59 c1 78 95 5d f9 e5 f9 00 6a 38 06 69
0000060 06 8d f1 80 a3 e2 a1 3c 6b ab 42 e8 18 0a f0 70
0000100
```

As the only chunk of data we don't get anything back:

```
❯ cargo run --bin sbp2json --quiet -- --error-handler to-invalid <corrupt_sbp_start.bin | wc -l
0
```

But in the start and middle test cases we do get invalid messages back:

```
❯ cat corrupt_sbp_start.bin test_data/roundtrip.sbp >corrupt_start.bin

❯ cargo run --bin sbp2json --quiet -- --error-handler to-invalid <corrupt_start.bin 2>/dev/null | grep INVALID | wc -l
1

❯ cat head.bin corrupt_sbp_start.bin tail.bin >corrupt_middle.bin

❯ cargo run --bin sbp2json --quiet -- --error-handler to-invalid <corrupt_middle.bin 2>/dev/null | grep INVALID | wc -l
1
```

We don't get anything if the bad data is at the end though:

```
❯ cat test_data/roundtrip.sbp corrupt_sbp_start.bin >corrupt_end.bin

❯ cargo run --bin sbp2json --quiet -- --error-handler to-invalid <corrupt_end.bin | grep INVALID | wc -l
0
```

Ultimately I think we want these test case to pass, but we might be getting into the weeds of how the SBP parser currently works, so we'll probably want to fix these issues with a new PR.

---

The above blobs are attached here: [corrupt_blobs.zip](https://github.com/swift-nav/libsbp/files/12030844/corrupt_blobs.zip)

pcrumley:
There is something else, where data aren't round tripping from sbp -> json -> sbp properly, even when it seems to create invalid messages.

This PR is at the point now where it needs unit tests on this behavior, but it is so large I am going to do a stacked PR to it, because it is getting too unwieldy. I am hopeful that with unit tests these bugs will be easy to fix.

silverjam:
As long as we can still round trip valid data I'm not sure I want to invest more time on this right now, once we get the new tests in let's think about moving on.  We've made a lot of good progress but this is starting to be a bigger time sink than I expected.

silverjam (approved):

rust/sbp/src/json/ser.rs
```diff
@@ -19,7 +20,7 @@ const BASE64_BUFLEN: usize = BUFLEN * 4;
 pub fn to_writer<W, M>(mut writer: W, msg: &M) -> Result<(), JsonError>
 where
     W: io::Write,
-    M: SbpMessage + Serialize,
+    M: SbpMessage + Serialize + WireFormat + Clone,
```
silverjam, line 23: I think WireFormat was introduced in all of these methods because it was removed form SbpMessage?  Are we sure that all of these methods really need WireFormat?

pcrumley: The call to `crate::ser::to_buffer` on line 239 requires `WireFormat`, which then has to be propagated up to pretty much all of the functions in `json::ser`. I took out `WireFormat` from the `SbpMessage` trait in an abandoned refactoring, so if you prefer i can re-add it.

Finally regarding benchmarks, it seems like the rust version is too fast and I'll need to bump some of the expected values.

silverjam: > I took out WireFormat from the SbpMessage trait in an abandoned refactoring, so if you prefer i can re-add it.

Let's re-add it if it's no longer needed

silverjam:
@pcrumley Can try cherry-picking ef262b452d54a5ec4cfeb8910b09af139a61bf4a to see if it fixes benchmarks here too.

pcrumley:
@silverjam it passes now! 🎉

pcrumley:
@silverjam it passes now! 🎉

silverjam:
@adrian-kong FYI, after this PR is merged, the next release cut of libsbp should be 5.0.0.

pcrumley:
I'm just waiting to merge to let other small changes in if we want to cut two releases. Otherwise I think it is g2g

silverjam:
> I'm just waiting to merge to let other small changes in if we want to cut two releases. Otherwise I think it is g2g

Two releases is a good idea, let's wait for Adrian to cut `4.17.0`, then we can merge this can cut `5.0.0`.

cc @pcrumley @adrian-kong

silverjam:
@pcrumley 4.17.0 was cut I think we're good to merge this

pcrumley:
Thanks for letting me push the button

silverjam:
> 3\. It has no unit tests. I wanted some eyes on this code and proposed solution before totally getting it over the finish line.

Just noticed this, this isn't true right?  I think I saw some unit test updates.

pcrumley:
No, It has unit test from bytes to `Sbp::Invalid` -> https://github.com/swift-nav/libsbp/pull/1347

In terms of unit test coverage, I think it is OK but there should be one that shows an invalid JSON message is parsed properly. I will do that today.

pcrumley:
plz review: https://github.com/swift-nav/libsbp/pull/1352

PR #1327, opened 2023-05-18
docker: pin key python tools (tox, sphinx)

silverjam:
# Description

@swift-nav/devinfra

Attempts to fix docker build by pinning key tools that we use to the last working version.

# JIRA Reference

n/a


adrian-kong (approved):

ljbade (approved):
tested it on my machine and the new Docker images works nicely

Issue #1331, opened 2023-05-25
Cannot access `GPSTime` inside `BaseTime`


marcomaida94:
Hello, 
I am trying to extract a timestamp from each SBPMessage that contains one (Rust). 

I found the function `gps_time()` by enabling the feature `swiftnav`:
```
    // frame: &Sbp
    if let Some(Ok(msg_time)) = sbp::messages::SbpMessage::gps_time(frame) {
        let _tow = match msg_time {
            MessageTime::Rover(RoverTime::Tow(rover_tow)) => rover_tow,
            MessageTime::Rover(RoverTime::GpsTime(gps_time)) => gps_time.tow(),
            MessageTime::Base(BaseTime(gps_time)) => gps_time.tow()
        };
    }
```
However `BaseTime` is a struct that contains `gps_time` as a private field:
https://github.com/swift-nav/libsbp/blob/d122b3b82e048e8a313967332ad97ebf05cc80c3/rust/sbp/src/time.rs#LL92C1-L92C1
So I can't access it.

Is this a bug or am I not supposed to do things this way?

Thanks,
Marco

notoriaga:
Not sure what exactly your use-case is but there is a function that converts `Iterator<Item = Sbp>` to `Iterator<Item = Sbp, Option<Result<GpsTime, GpsTimeError>>>` (also works with `Iterator<Item = Result<Sbp>>`) - https://github.com/swift-nav/libsbp/blob/6997d9fd083e0c71dcbd9d14983954c8aaa37f3e/rust/sbp/src/sbp_iter_ext.rs#L48

It's not showing up in the docs because it requires the `swiftnav` feature (which enables some integrations with https://github.com/swift-nav/swiftnav-rs) and I guess we aren't building the docs with all features enabled.

It might not be want you want because it excludes BaseTime messages, but it takes care of tracking the week number for you which is nice. Either way we have a fix for this on the way 😄

marcomaida94:
Thank you! 

I will patch locally waiting for the fix.

silverjam:
Release v4.14.1 with this fix has been cut (and now available on crates.io).

marcomaida94:
Thank you so much!

PR #1332, opened 2023-05-25
Make the BaseTime inner public

pcrumley:
# Description

@swift-nav/devinfra

This PR makes the inner of the BaseTime tuple struct public. This enables users to access the time inside and makes it similar to how RoverTime can be used.

Some minor notes is that the tuple struct here isn't really doing a whole lot anymore so you could envision just removing it and replacing it with `GpsTime` which would be a major breaking change, or you could only allow access to the inner values with an `to_gps_time()` function. However, I think this solution makes the most sense given how `RoverTime` is used and the desire to not add a breaking change for something so simple. 
 
# API compatibility

It should not, adding a public item, should be a minor change.
[Minor: adding new public items](https://doc.rust-lang.org/cargo/reference/semver.html#minor-adding-new-public-items)
Adding new, public [items](https://doc.rust-lang.org/reference/items.html) is a minor change.## API compatibility plan

N/A

# JIRA Reference

would resolve: https://github.com/swift-nav/libsbp/issues/1331

### DIFF
--- a/rust/sbp/src/time.rs
+++ b/rust/sbp/src/time.rs
@@ -89,10 +89,14 @@ impl From<f64> for RoverTime {
 /// Time received via a message from a base station, like
 /// [MsgObs](crate::messages::observation::MsgObs).
 #[derive(Debug, Clone, Copy, PartialEq, PartialOrd)]
-pub struct BaseTime(pub GpsTime);
+#[non_exhaustive]
+pub enum BaseTime {
+    /// The [GpsTime] of the message.
+    GpsTime(GpsTime)
+};
 
 impl From<GpsTime> for BaseTime {
     fn from(gps_time: GpsTime) -> Self {
-        BaseTime(gps_time)
+        BaseTime::GpsTime(gps_time)
     }
 }
--- a/rust/sbp/src/time.rs
+++ b/rust/sbp/src/time.rs
@@ -89,7 +89,7 @@ impl From<f64> for RoverTime {
 /// Time received via a message from a base station, like
 /// [MsgObs](crate::messages::observation::MsgObs).
 #[derive(Debug, Clone, Copy, PartialEq, PartialOrd)]
-pub struct BaseTime(GpsTime);
+pub struct BaseTime(pub GpsTime);
 
 impl From<GpsTime> for BaseTime {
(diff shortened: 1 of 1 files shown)


notoriaga:
> Some minor notes is that the tuple struct here isn't really doing a whole lot anymore so you could envision just removing it and replacing it with GpsTime which would be a major breaking change, or you could only allow access to the inner values with an to_gps_time() function. However, I think this solution makes the most sense given how RoverTime is used and the desire to not add a breaking change for something so simple.

While technically breaking I doubt there are many users of this, partly because in it's current state it's useless and partly because most usecases (that I'm aware of) want `RoverTime`/use `.with_rover_time()`. So I wouldn't be against removing `BaseTime` and replacing it with just `GpsTime`. 

One caveat would be if we added a message that uses `BaseTime` that only has a tow we'd need to replace this type with an enum similar to `RoverTime`. But that's a breaking change with either set up. The only way I can think to prevent that would be changing `BaseTime` to an enum now, either with a `Tow` variant that is never used or with just a `GpsTime` variant and making it `#[non_exhaustive]`. That being said it seems unlikely that we'd add a message like that

notoriaga:
So I guess what I'm saying is that if you want to remove the tuple struct I think it's probably fine to sneak it in without a major semver change but other people might disagree

pcrumley:
@notoriaga if you look at this diff, you'll see i did in fact think about solving the problem by just switching to an enum with non_exhaustive
https://github.com/swift-nav/libsbp/pull/1332/commits/2dbb98aac3f9ac29cfb0ce94811c55e74eb77e77

I kind of went on the fence because I didn't know if it was a major or minor change. I think it would actually be a minor one because tuple constructors are private if any field is private, but i'm not 100% sure:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cd7d7b0194eb3a66a378fe6129b5a03a

notoriaga:
Oh nice! Yeah I think we're probably fine here

silverjam:
Agree that there are unlikely to be users of this functionality, so breaking changes are probably OK to sneak in.

silverjam (approved):
Looks fine, assuming this aligns with what @notoriaga had in mind

notoriaga (approved):

PR #1334, opened 2023-05-26
Increase SLUSH_PERCENTAGE to 45%

pcrumley:
The actual variation of benchmarks is closer to 45% than 25%. This PR would just increase the tolerance we have for variation in our benchmarks when doing checks against the `main`.

An alternative approach would be to diagnose what is causing this variability and work to decrease it, but [I really like slop!](https://www.amazon.com/Really-Like-Slop-Elephant-Piggie/dp/1484722620)

# Description

@swift-nav/devinfra

Increase slop to 45%

# API compatibility

No, it's only risk would be allowing perf degradations. 

## API compatibility plan

No

# JIRA Reference

None, came from getting an errant warning about my PR failing.

### DIFF
--- a/test_data/benchmark_main.py
+++ b/test_data/benchmark_main.py
@@ -7,7 +7,8 @@ import json
 import subprocess
 
 # If the ratio is off from expected by more than this percentage
-SLUSH_PERCENTAGE = 0.25
+# fail the benchmark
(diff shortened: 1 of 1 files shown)


pcrumley:
Closing this because from internal conversations it is clear that strategy is going to be to try and make the benchmarks more consistent, and not to increase the leeway.

PR #1336, opened 2023-05-30
Generate Kaitai Struct Format Descriptions and run Kaitai Struct Compiler to generate language bindings

dgburr:
# Description

[Kaitai Struct](https://kaitai.io/) is a declarative language used to describe binary data structures.  KSY files containing format specifications for all supported SBP messages can be generated using the `kaitai` target.  The resulting files can be found in the [`kaitai/ksy`](./kaitai/ksy) directory. This target also runs the Kaitai Struct compiler to generate bindings for perl and python.  These bindings (along with test cases) can be found in the [`kaitai/perl`](./kaitai/perl) and [`kaitai/python`](./kaitai/python) subdirectories respectively.

The Kaitai Struct format description files can also potentially be used to generate bindings for the following targets:

* graphviz
* csharp
* rust
* java
* go
* cpp_stl
* php
* lua
* nim
* html
* ruby
* construct
* javascript

The resulting python bindings allow for significantly faster parsing in comparison to the native construct-based bindings included in libsbp. 

# API compatibility

The new bindings do not have any impact on the existing language bindings.  However, due to the nature of Kaitai Struct, the new bindings do not currently allow for serialisation (i.e. construction of SBP objects).   Serialisation for java and python targets is supported as an experimental feature in the development version of Kaitai Struct (> 0.10) so this restriction should hopefully be removed in the future.

It should also be noted that the field names used by the KSY bindings are not 1:1 identical to the native bindings due constraints which Kaitai Struct places on various token types, e.g. it is forbidden to use upper case characters in field names.

## API compatibility plan

N/A


silverjam:
@dgburr Not opposed to this in general, but in the interest of reducing the support burden I would suggest that we scope this down to the languages where we've identified a need for this... so, can we reduce this to just emitting bindings for Python?

README.md
```diff
@@ -263,6 +265,37 @@ provided.  Libraries for JavaScript, TypeScript, and Elm generated by the
...
+
+* graphviz
+* csharp
+* perl
+* java
+* go
+* cpp_stl
+* php
+* lua
+* python
```
silverjam, line 285: Would like to reduce this to just emitting Python, if other languages are needed, presumably one can use the .ksy files to produce a parser?

generator/sbpg/targets/resources/sbp_kaitai_main.ksy.j2
```diff
@@ -0,0 +1,60 @@
...
+
+# Automatically generated with generate.py.
+# Please do not hand edit!
+
+meta:
+  id: sbp
+  file-extension: sbp
+  endian: le
+  imports:
+((* for s in specs *))
```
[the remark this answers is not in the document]
dgburr: I have some test cases ready to go.  But I wanted to get the core part working before pushing them in.

README.md
dgburr: We've had customers asking about C# bindings in the past so I'd like to leave those in.  And ruby/go/lua are fairly common, and we have nothing else to offer users of these languages.  But I wouldn't have any issues with removing the following:

* graphviz
* perl
* java
* cpp_stl
* php
* nim
* html
* javascript

silverjam: Anything where we already have a language implementation should definitely be omitted unless (like Python) there is a clear advantage to the Kaitai alternative.  For other languages I am mostly worried about support, anything that we "officially" generate should either have a test suite or it should be painfully obvious that it's not officially unsupported (having to generate the language bindings yourself should make it obvious that we don't support it).  In general though, we must have a test suite for the devinfra team to be able to support the code.

silverjam: So perhaps "C# support" should come as set of documentation for how to generate the bindings and a disclaimer that it's not officially supported (yet).

dgburr: I'm fine with writing some test cases for C#, although then it would be a question of how we can automatically execute them in CI (I'm guessing mono?).

silverjam: > I'm fine with writing some test cases for C#, although then it would be a question of how we can automatically execute them in CI (I'm guessing mono?).

Note the current test harnesses generate test cases for all `test_*.yaml` files for each language, similar to how we generate language support bindings.  For example see:
- https://github.com/swift-nav/libsbp/blob/master/spec/tests/yaml/swiftnav/sbp/signing/test_MsgEcdsaSignature.yaml
- https://github.com/swift-nav/libsbp/blob/master/generator/sbpg/targets/resources/rust/test/sbp_tests_template.rs

We have access to Windows nodes, but Mono should work too, it's already available on GHA Linux images:

image

See https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md

dgburr:
Benchmark results from `test_benchmarks.py` test case:

construct: 24.84823133982718
ksy: 5.861570787616074
hybrid1: 5.817689431831241
hybrid2: 6.382694503292441

In other words: the KSY-based version of the python SBP parser runs approximately 5 times faster than the original construct-based version.

dgburr:
> Benchmark results from `test_benchmarks.py` test case:
> 
> construct: 24.84823133982718 
> ksy: 5.861570787616074
> hybrid1: 5.817689431831241
> hybrid2: 6.382694503292441
> 
> In other words: the KSY-based version of the python SBP parser runs approximately 5 times faster than the original construct-based version.

Now that I have now hooked up the new kaitai-derived parsers to the existing benchmarking infrastructure, I am seeing the following results:

```
Summary
  './test_data/benchmark/sbp2json_rust.py' ran
    2.14 ± 0.07 times faster than './test_data/benchmark/sbp2json_haskell.py'
    3.99 ± 0.08 times faster than './test_data/benchmark/sbp2json_kaitai_python.py'
   17.00 ± 0.33 times faster than './test_data/benchmark/sbp2json_python.py'
   19.34 ± 0.37 times faster than './test_data/benchmark/sbp2json_kaitai_perl.py'
Benchmark #1: ./test_data/benchmark/json2sbp_rust.py
```
So the speed-up of the kaitai-based python parser compared to the construct-based version appears to be "only" 4.2x rather than 5x as stated above.

kaitai/python/tests/utils_sbp2json.py
```diff
@@ -0,0 +1,132 @@
...
+    obj.payload.length = obj.length
+    obj.payload._io.seek(0)
+    obj.payload.payload = base64.standard_b64encode(obj.payload._io.read_bytes_full()).decode('ascii')
+    obj.payload.crc = obj.crc
+
+    return obj.payload
+
+
+# work-alike of get_next_msg_construct() based upon Kaitai Struct
+def get_next_msg_kaitai(fp):
```
silverjam, line 95: I would think we could generalize and expose it as an API for the kaitai based lib, e.g.

```
from kaitai.python import iter_messages

for msg in iter_messages(fp):
   # Do stuff with msg
```

Here we'd simply import `iter_messages` and compose it with the `get_flattened_msg` function.

dgburr: Yeah - this file is basically intended as a complete working example for anybody who wants to work with the kaitai python bindings.  What about if I move it from `kaitai/python/tests/utils_sbp2json` to `kaitai/python/tests/parser.py`?

silverjam: Sounds good

Dockerfile
```diff
@@ -85,6 +85,8 @@ RUN \
       python3.10 python3.10-dev python3.10-distutils \
       dpkg-dev \
       cmake \
+      libjson-perl \
+      libdigest-crc-perl \
```
silverjam, line 89: So, I think choosing two languages to show the benefit of using KSY is great, and apologies for calling into question the validity of this work but...

I'm not sure about the choice of Perl here, do we have customers that want to parse SBP with Perl?  What's the motivation for supporting it?  Picking something with customer motivation (e.g. C#) would make more sense in my opinion.

Again, I don't want to torpedo this work because I don't agree with the choice of Perl, but if the DevInfra team is going to support this, we should understand the motivations.

Makefile
```diff
@@ -395,6 +425,15 @@ test-protobuf:
 	$(call announce-begin,"Running Protocol Buffer tests")
 	$(call announce-end,"Finished running Protocol Buffer tests")
 
+test-kaitai:
+	$(call announce-begin,"Running Kaitai Struct Python tests")
+	cd $(SWIFTNAV_ROOT) && tox -c kaitai/python/tests/tox.ini kaitai/python/tests/test_*.py
```
silverjam, line 430: Is it possible to generate a module name that would provide a more descriptive name outside of this repo?  This current layout means that we do `from python import ...` and outside of this repo this would look out of place.  Can we change the generated Python code so that imports would look like `from kaitai_sbp import ...`?  Or maybe even `from sbp.kaitai import ...`.

kaitai/python/tests/utils.py
```diff
@@ -0,0 +1,205 @@
+# Utilities for tests cases for the python version of the Kaitai Struct SBP parser
+
+import kaitai.python.sbptable as kaitai_sbptable
+from kaitai.python.parse_utils import iter_messages as get_next_msg_kaitai, get_flattened_msg, serialise, BufferKaitaiStream
```
silverjam, line 4: As mentioned above, if we want to publish this to PyPI, import should look something like this:

```suggestion
from sbp.kaitai.parse_utils import iter_messages as get_next_msg_kaitai, get_flattened_msg, serialise, BufferKaitaiStream
```

kaitai/python/tests/utils.py
```diff
@@ -0,0 +1,205 @@
...
+    elif isinstance(obj, list):
+        return [dictify(x, round_floats) for x in obj]
+    elif round_floats and type(obj) == float:
+        return float('{:.10f}'.format(obj)[:11])
+    else:
+        return obj
+
+
+# adapted from construct-based python version of sbp2json
+def parse_file_construct(fp):
```
silverjam, line 37: Worried that we're duplicating a lot of code here, can we just re-use the function directly?  Possible if we make adjustments to the original code we can just re-use directly here.

Dockerfile
dgburr: I did have a go at implementing a version for C#, but I quickly discovered that there are some issues with the .cs files generated by kaitai-struct-compiler.  I played around with it for a bit but then gave up and moved onto another language which I am more familiar with (I also ended up finding some minor issues in the generated perl code, but I managed to work around those).

Ideally, one of the following will happen in the future:
1. The bugs will get fixed in kaitai-struct-compiler
2. Somebody with sufficient motivation implements fixups for the generated csharp output

But my main motivation was to have working bindings for at least 2 languages (since this forced me to do some substantial refactoring which will make it easier to add even more languages in the future)

Makefile
dgburr: I will re-arrange as follows:

`kaitai/python -> `kaitai/python/sbp`
`kaitai/perl -> `kaitai/perl/Sbp`

silverjam: :+1:

dgburr: Done

kaitai/python/tests/utils.py
dgburr: Done

silverjam (approved):
Not super happy with the fact that we have two python modules that are called `sbp` but I guess we can live with that for now.  If in the future we want to distribute this to PyPI I'm not sure how things will play out if someone wants to install the old Python module and the new module at the same time.  I think it would be a lot cleaner if one was called `katai_sbp` and the original module was just `sbp`.

silverjam:
> > Not super happy with the fact that we have two python modules that are called `sbp` but I guess we can live with that for now. If in the future we want to distribute this to PyPI I'm not sure how things will play out if someone wants to install the old Python module and the new Katai module at the same time. I think it would be a lot cleaner if one was called `katai_sbp` and the original module was just `sbp`.
> 
> Done

😄

PR #1338, opened 2023-06-05
Add tdk-icm42670 to imu.yaml

EricCalgary:
# Description

- add `TDK-ICM42670` to imu.yaml


adrian-kong (commented):
need to run `make all` and commit generated artifacts

EricCalgary:
> 



> need to run `make all` and commit generated artifacts

@adrian-kong I have never done this.  May you please tell how to do it?

adrian-kong:
> @adrian-kong I have never done this. May you please tell how to do it?

yeah the `spec` yaml files autogenerates the code so you'd need to either run `make all` (if that works) or run `make gen-rust gen-c gen-java`  should be the only commands you'd need I believe and commit the generated files i.e. `rust/sbp/src/messages/imu.rs` (check they are all good)

refer: https://github.com/swift-nav/libsbp/blob/master/HOWTO.md#adding-and-testing-new-messages

EricCalgary:
have been in the process of trying to produce `imu.rs`: need to fix some errors by installing some packages

silverjam:
> have been in the process of trying to produce `imu.rs`: need to fix some errors by installing some packages
> 
> * completed this without error:
>   `docker run -v $PWD:/mnt/workspace -i -t swiftnav/libsbp-build:2022-06-14 \ /bin/bash -c "make docs"`
> * got updated `docs/sbp.pdf`

You can just run `docker run -v $PWD:/mnt/workspace -i -t swiftnav/libsbp-build:2022-06-14` it will regenerate everything.

> * trying to remove the following errors when run `make gen-rust gen-c`

If you want to regenerate individual languages, you'll need to run this inside docker, we don't support running natively on macOS.

silverjam (approved):

EricCalgary:
> @EricCalgary no need to run docker now though, I was able to push the necessary updates.

Thank you @silverjam . I have been and still am being challenged with `Mac`

adrian-kong:
> * trying to remove the following errors when run `make gen-rust gen-c`
>   image

@EricCalgary 

huh that looks like something to do with tox version im guessing ( if scrolling down a little has something along the lines of: )
``` 
from tox import hookimpl ...
ImportError: cannot import hookimpl from tox ...
```
then you probably have tox 4 which doesn't have hookimpl, should downgrade to 3.12.1 by something like this...
```
/Users/eric.li/Desktop/2B_tested/libsbp_git/try2/libsbp/.tox/.tox/bin/python  -m pip install --upgrade --force-reinstall tox==3.12.1
```

let me know if this works (more error logs to see whats going on)
can always run via docker if local setup doesn't run well immediately (only if you need to run a lot then local setup feels much faster)

adrian-kong (approved):

PR #1339, opened 2023-06-05
fix tests 

pcrumley:
The minimum set of changes to get the parse unknown falllback tests to pass, but I'm not 100% sure it solves the desired problem


silverjam (approved):
Approving this to move it out of my review queue with the caveat that we discussed what's actually needed here offline: we need to assert the minimum message size, not the minimum payload size (which is zero).

pcrumley:
This is stacked and pointing towards the `unknown-fallback` branch not main, so there is less risk in merging. That said, this diff [0844199](https://github.com/swift-nav/libsbp/pull/1339/commits/084419957177261b26721cdf7d20442981526b84) should resolve issue enforcing a min frame length.

pcrumley:
I am going to merge this into the parent branch, it will be easier to judge the PR as a whole if that is the case.

PR #1342, opened 2023-06-13
c: bump minimum cmake to 3.13

### LINKED ISSUE #1320: Unknown CMake command "add_link_options"
While compiling the SBP bindings for C, I get an `Unknown CMake command "add_link_options"` when using CMake 3.12 which is specified as the minimum required CMake version in the CMakeLists.txt.

I believe the add_link_options was introduced in CMake 3.13.

silverjam:
# Description

Closes #1320 by bumping the minimum cmake to version 3.13.

@swift-nav/devinfra

# API compatibility

_Does this change introduce a API compatibility risk?_

No, just bumping the minimum required version of cmake to match the features we're using.  Specifically `add_link_options` -- see https://cmake.org/cmake/help/latest/command/add_link_options.html.

# JIRA Reference

See issue #1320

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.13)
 project(libsbp C CXX)
 
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake" "${CMAKE_CURRENT_LIST_DIR}/cmake/common")


pcrumley (approved):
Lgtm

silverjam:
cc @isaactorz

isaactorz (approved):

isaactorz:
For some reason this didn't post my review comment I left:

This might break the yocto builds that only have 3.12 - but since we've migrated most of that stuff to bazel my opinion is that we should remove the 3.12 requirement and people should start using bazel to build for those platforms.

silverjam:
> For some reason this didn't post my review comment I left:
> 
> This might break the yocto builds that only have 3.12 - but since we've migrated most of that stuff to bazel my opinion is that we should remove the 3.12 requirement and people should start using bazel to build for those platforms.

Wouldn't our use of the new `add_link_options` feature already break Yocto builds?

isaactorz:
> Wouldn't our use of the new `add_link_options` feature already break Yocto builds?

It's probably not executing that code path but the change to `cmake_minimum_required` might be problematic.

PR #1344, opened 2023-06-26
Add function to convert sbp_msg_type_t to string [ESD-2613]

woodfell:
# Description

By request, add a new function to the C bindings which can convert an instance of `sbp_msg_type_t` to its string representation. Similar to how the message name can be extracted from type traits when the message type is known at compile time, this function allows runtime conversion of a message type to human readable name.

@swift-nav/devinfra

# API compatibility

No. Adding a new function without modification to any existing code

## API compatibility plan

N/A

# JIRA Reference

https://swift-nav.atlassian.net/browse/ESD-2613


lucasle-sn (approved):

peddie (approved):
I hope you auto-generated that!  Bravo.

isaactorz (approved):

silverjam (approved):

PR #1347, opened 2023-07-13
add unit tests for recovering from parsing errors of sbp

pcrumley:
This PR adds a unit test showing that parsing errors fallsback to an invalid message.

Seeing this unit test makes it clear why some of the questions @silverjam  was having about invoking this on certain inputs causes it to pass and others cause it to fail. Additionally the round-tripping on arbitrary bytes is not possible yet because both the iter_frame & iter_messages will only return a message if there is something that starts with the prelude byte, and it returns the bytes that exist between that prelude and HEADER_LEN, whatever the length of the payload is on byte 5 + the CRC_LEN. This means that if there are extra bytes sent in the stream for aliasing for instance, the framer ignores those bytes, and they are never even tried to be parsed.

@silverjam gave two examples that do not throw an error as the unknown fallback is currently implemented. One of those has been added as a unit test here. But the main takeaway is the invalid-fallback as currently written only works on messages that are properly framed but perhaps wrongly labeled or with a CRC error.

### DIFF
--- a/rust/sbp/src/de.rs
+++ b/rust/sbp/src/de.rs
@@ -11,8 +11,8 @@ use dencode::FramedRead;
 use futures::StreamExt;
 
 use crate::{
-    messages::invalid::Invalid, messages::SbpMsgParseError, HandleParseError, Sbp, CRC_LEN,
-    HEADER_LEN, MAX_FRAME_LEN, PAYLOAD_INDEX, PREAMBLE,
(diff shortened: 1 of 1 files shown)


silverjam (approved):

silverjam:
I think we need to adjust benchmark thresholds before the final PR is merged.

PR #1352, opened 2023-07-18
add test to round trip an invalid sbp message

pcrumley:
# Description

@swift-nav/devinfra

This adds an integration test (with some unit tests inside of it) which test the round tripping of an invalid message from invalid sbp -> json with msg_name "INVALID" -> invalid sbp. Covers a gap in test coverage from an earlier PR: https://github.com/swift-nav/libsbp/pull/1312 

Please note, there are no non-test code changes here, so it passed as is.


silverjam (approved):
:tada:

PR #1355, opened 2023-08-22
Remove SBP_SENDER_ID: it is not used anywhere and can conflict with user code

dgburr:
# Description

Remove `SBP_SENDER_ID` since it is not used anywhere and can conflict with user code.

# API compatibility

There is no API risk since the define is not used anywhere.

## API compatibility plan

N/A

# JIRA Reference

N/A

### DIFF
--- a/c/include/libsbp/sbp.h
+++ b/c/include/libsbp/sbp.h
@@ -49,8 +49,6 @@ extern "C" {
 /** Return value indicating an error while decoding an SBP message from wire format */
 #define SBP_DECODE_ERROR (-8)
 
-/** Default sender ID. Intended for messages sent from the host to the device. */
-#define SBP_SENDER_ID 0x42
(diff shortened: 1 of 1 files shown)


silverjam (approved):

woodfell (approved):

evan-flynn-apexai:
@dgburr @woodfell @silverjam looks like the `SBP_SENDER_ID` is still used in the ROS 2 driver here: https://github.com/swift-nav/swiftnav-ros2/blob/master/test/publishers/test_custom_publishers.cpp#L92

What should be the replacement for `SBP_SENDER_ID` now that its removed?

dgburr:
@evan-flynn-apexai: you should copy the line which was removed (`#define SBP_SENDER_ID 0x42`) into the ROS 2 driver.

PR #1359, opened 2023-08-26
fix bug

pcrumley:
# Description

@swift-nav/devinfra

So I just put up a quick PR earlier thinking i had fixed the new lints, and set to auto-merge figuring i could push up new fixes if it didn't work. But what happened was the rust lint stage actually failed and it still let automerge go through!

This PR actually fixes those bugs, and I think the branch protection rules of libsbp need to be reviewed.


silverjam (approved):

silverjam:
Libsbp automerge unfortunately only blocks on a few required checks

PR #1365, opened 2023-10-19
CI: force release for automated libsbp releases

adrian-kong:
# Description

@swift-nav/devinfra

Add a force release CI to update the versioning files.
We will trigger this ONLY after necessary CI stages passes.

This will create a PR with the changes necessary.
The release tags will occur in a different CI by triggering on the created PR merging event.

# JIRA Reference

https://swift-nav.atlassian.net/browse/DIP-346


sokhealy:
Outdated

PR #1370, opened 2023-10-20
Bon voyage!

pcrumley:
# Description

@swift-nav/devinfra

remove @silverjam from code_owners. I think it would be a good idea to add an additional owner to the `c/` directory but don't know who would be a good choice.


isaactorz (approved):
This is a sad "ship it"!

I think between me and the platform team we've got plenty of coverage for the c code.

silverjam (approved):
:vulcan_salute:

Issue #1363, opened 2023-10-16
`MsgGpsTime` implementation of `gps_time` is not checking status flags


marcomaida94:
Hello, 
I have experienced some inconsistency in `MsgGpsTime`

```
{"crc":11885,"length":11,"msg_type":258,"msg_name":"MSG_GPS_TIME","payload":"AADM7g0AAAAAAAA=","preamble":85,"sender":789,"wn":0,"tow":913100,"ns_residual":0,"flags":0}
```

As you can see here, week number is zero (this data is recent) and the `flags` field is also `0`, which is supposed to communicate that the message is invalid. By reading the docs:
<img width="289" alt="Screenshot 2023-10-16 at 18 17 51" src="https://github.com/swift-nav/libsbp/assets/120410627/0977069c-6b88-441d-8510-06b8b99b2f0b">

However, the [`gps_time`](https://github.com/swift-nav/libsbp/blob/0bfe6c61bc22e370e50259d1e2e7a4fe1541e277/rust/sbp/src/messages/navigation.rs#L1890) function does not check for this flag, and when called it returns an invalid timestamp for the above case.

Should the function be fixed and return `None` if the status flag is 0?

Thank you!

silverjam:
Hello @marcomaida94 -- we're discussing this internally, we should have a response for you soon.  Thanks for raising the issue!

marcomaida94:
Thank you @silverjam for the quick reply !

pcrumley:
Hi @marcomaida94 -- Thanks for bringing this issue to our attention. It should be closed by PR: #1375

marcomaida94:
Thank you!

PR #1375, opened 2023-11-13
rust: check time source for MsgGpsTime

notoriaga:
# Description

@swift-nav/devinfra

# API compatibility

## API compatibility plan

# JIRA Reference


generator/sbpg/targets/rust.py
```diff
@@ -60,6 +60,11 @@
 BASE_TIME_MSGS = ["MSG_OBS", "MSG_OSR", "MSG_SSR"]
 
 CUSTOM_GPS_TIME_MSGS = {
+    "MSG_GPS_TIME": """
+if !matches!(self.time_source(), Ok(TimeSource::GnssSolution) | Ok(TimeSource::Propagated)) {
+    return None;
+}
+""".strip() + GPS_TIME,
```
pcrumley, line 67: code golf but could be:
```suggestion
    "MSG_GPS_TIME": """
self.time_source().ok()?
""".strip() + GPS_TIME,
```

notoriaga: Still need to check that it's not equal to `TimeSource::None` though (which obviously is not very clear by the way I'm doing it now). Maybe

```rust
if self.time_source().ok()? == TimeSource::None {
    return None;
}
```
?

pcrumley (approved):
This seems reasonable if we want to implement a fix to: https://github.com/swift-nav/libsbp/issues/1363

I had a small fix to turn for terseness but I don't mind it as written.

generator/sbpg/targets/rust.py
pcrumley: i see, in that case i probably prefer it as written, but if we are using a rust version later than 1.70
```rust
if !self.time_source().is_ok_and(|x| x != TimeSource::None) {
     return None
}
```

but i think it is better as written.

notoriaga: `is_ok_and` TIL! But yeah agreed it's probably fine as is

PR #1376, opened 2023-11-14
rust: check more messages for time validity 

notoriaga:
# Description

Same as https://github.com/swift-nav/libsbp/pull/1375 but accounts for the other 3 time messages and msg_odometry

@swift-nav/devinfra

# API compatibility

## API compatibility plan

# JIRA Reference

### DIFF
--- a/generator/sbpg/targets/rust.py
+++ b/generator/sbpg/targets/rust.py
@@ -59,12 +59,13 @@ let gps_time = match time::GpsTime::new(0, tow_s) {
 
 BASE_TIME_MSGS = ["MSG_OBS", "MSG_OSR", "MSG_SSR"]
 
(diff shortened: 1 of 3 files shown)


generator/sbpg/targets/rust.py
```diff
@@ -59,6 +59,12 @@
 
 BASE_TIME_MSGS = ["MSG_OBS", "MSG_OSR", "MSG_SSR"]
 
+CHECK_FLAGS = """
+if self.time_source().ok()? == TimeSource::None {
```
notoriaga, line 63: Actually like this one more because it checks against a specific variant. So if a message gets added that has a `TimeSource` but no `None` state for some reason we'll get a compile error

pcrumley (approved):
Thanks for following up on this and getting the remaining messages

PR #1377, opened 2023-11-15
Increase test coverage of C code [AP-949]

woodfell:
# Description

@swift-nav/devinfra

Increase the test coverage of C code by rewriting the template for the V4 C++ tests

This is a complete rewrite of the existing template which covers the following test cases:
- Behaviour of `*_encoded_len()`
- Encoding payload directly to a user-provided buffer
  - Underflow errors and alternative uses
- Decoding payload from a user provided buffer
  - Underflow errors and alternative uses
- Sending and receiving complete frame through C API (`sbp_state_t`)
- Sending and receiving complete frame through C++ API (`sbp::State` and `sbp::MessageHandler<>`)
- Encoding/decoding payload through C++ API
- Comparison functions and operators with all potential mismatches covered automatically
- C++ type traits
- All string functions generated as part of the API for each message which contains string fields

This comprehensive template covers 100% of generated V4 code. No attempt was made to increase coverage of the legacy API, although there is little code to test anyway.

This rewrite by itself increases code coverage from ~20% to ~55% just based on the existing test cases.

Coverage can be further increased by introducing new test cases for the 120 or so messages which are currently uncovered. This is done on a separate PR https://github.com/swift-nav/libsbp/pull/1382 and brings coverage up to ~99% (generated locally with lcov, sonar cloud seems to have a different way of calculating it)

To keep things a little smaller this PR will not introduce the new test cases, they will be kept separate and merged in to this branch before going to master

To aid with reviewing the commits are broken up to logical steps. The first couple deal with some supporting changes such as altering the generate to pass through some extra meta information (the generated companion fields for variable length arrays were already present to some extent in the previous template, but the information provided was not sufficient to get complete test coverage)

Next there are some minor alterations to some existing test cases. These all follow the same pattern of only affecting messages which have variable length arrays or string. For variable length arrays the existing specification for the companion "count" fields is extended out with 2 extra bits of information, for encoded strings there is one new fields. Both of these are required in order to get 100% coverage in the new template.

Finally, all code is regenerated. The final commit makes up the vast majority of changes in this PR. I suggest concentrating on just a single test case along with the code template itself, once these are understood all the other generated files can be treated with confidence.

Sonar cloud is not reporting any code coverage for this PR because there are no changes to actual code in libsbp, only the test case source code which isn't included in the calculation. You can see the effect of this PR by following the sonar cloud link an noting the "estimated after merge" number.

*For Reviewers*
Suggested places to concentrate on:
- `generator/sbpg/targets/resources/c/test/v4/sbp_cpp_test.cc.j2` - The main template which was rewritten for this PR
- `spec/tests/yaml/swiftnav/sbp/logging/test_MsgFwd.yaml` - for an example of how the meta information for variable length arrays has changed
  - `c/test/cpp/auto_check_sbp_logging_MsgFwd.cc` - generated output of above test case
- `spec/tests/yaml/swiftnav/sbp/settings/test_MsgSettingsReadByIndexResp.yaml` - for an example of how the meta information for encoded strings has changed
  - `c/test/cpp/auto_check_sbp_settings_MsgSettingsReadByIndexDone.cc` - generated output of above test case

# API compatibility

No

## API compatibility plan

N/A

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-949


generator/sbpg/test_structs.py
RReichert, line 1: looks like the python sonarcloud checks need tweaking (out of scope for PR). its assuming this file is production code and needs test cases.

RReichert (approved):
Had a glance through (since these are test cases update and have no production impact), looks good.

isaactorz (approved):

PR #1378, opened 2023-11-20
Fix decoding fixed length arrays of struct in python sbp2json [AP-946]

woodfell:
# Description

@swift-nav/devinfra

Alter the python decoding generated code so that fixed length arrays which contain structs (as opposed to primitive types) are properly decoded with the correct parser for the filled type.

Variable length arrays are already decoded properly in all cases, it's only fixed length arrays which are broken. Only the python binding are affected by this bug.

# API compatibility

Sort of.

The output of the python version of `sbp2json` changes.

Currently messages which contain a fixed length array of structs are decoded as byte arrays in the python version of sbp2json. This is misleading.

The rust version of sbp2json is able to decode these constructs properly. Therefore the output of the python and rust version of `sbp2json` can differ given identical inputs.

In the current definition there are only 3 message types which suffer from this behaviour - `MsgTrackingIq`, `MsgTrackingIqDepA`, `MsgTrackingIqDepB` - all very rare and not really seen in the wild.

Therefore compatibility risk is practically zero.

## API compatibility plan

None required

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-946


howardfernandez (approved):

python/sbp/RELEASE-VERSION
```diff
@@ -1 +1 @@
-5.0.3
\ No newline at end of file
+5.0.4-alpha
```
pcrumley, line 1: I'm not sure you should just bump this version, probabl yneed to do a full new release

ljbade (approved):

PR #1379, opened 2023-11-20
Fix incorrect specification of directory contents in read dir resp [AP-945]

woodfell:
# Description

@swift-nav/devinfra

MSG_FILEIO_READ_DIR_RESP has a string field which indicates the contents of the directory. In the SBP spec the encoding of the string is set properly (multipart) but the type of the field is incorrectly set to array. This means that the field is not actually treated as a string in the generated bindings but rather as a byte array.

This generally isn’t causing problems because this message is only used by piksi and related tools which are all using older versions of libsbp. They generally haven’t been updated to use the V4 API so this change will not actually cause any problems. But it should be fixed to enable improvements to unit tests

# API compatibility

Technically this does introduce an API change, but the message is unused except in legacy contexts so there should be no issue at all

## API compatibility plan

Should any issues come up they are easy to resolve, simply use the newly generated functions in the V4 API (`sbp_msg_fileio_read_dir_resp_contents_*`) instead of directly interacting with the byte array

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-945


howardfernandez (approved):

pcrumley (commented):
we should take this into master then bump the version to
`5.1.0` using the relase process

woodfell:
> we should take this into master then bump the version to `5.1.0` using the relase process

I'm about to queue up some improvements to tests which should bump up the code coverage stats and stop the sonar cloud stage from failing on each new PR. After that I'll learn how to do new releases and get everything in

RReichert (approved):
API change is reasonable and desirable :+1:

PR #1381, opened 2023-11-21
Fix decoding of truncated messages containing variable length arrays [AP-948]

woodfell:
# Description

@swift-nav/devinfra

If an SBP message has a variable length array of some type, when decoding a input wire encoded buffer which has been truncated the decoder will not produce an error when it should.

 

Imagine a message containing some amount of overhead, then a variable length array of some type.
```
+---------------------------------------+
| header | elem0 | elem1 | .... | elemN |
+---------------------------------------+
```

In this message the array is not fixed length, rather the number of elements is variable and is calculated from the size of the input buffer:
```
num_elems = (sizeof(input_buffer) - sizeof(header)) / sizeof(elem)
```

Decoding of each element is then performed in a loop:
```
for (int i = 0; i < num_elems; i++) {
  result = decode_elem(buffer, &output[i]);
}
```

If the size of the header is 3 bytes and the size of each element is 4 bytes an example representation on the wire could be:
offset
```
      0   1   2   3   4   5   6   7   8   9   A   
      +-----------+---------------+---------------+
      | header    | elem0         | elem1         |
      +-----------+---------------+---------------+
```

And so long as this is what’s fed in to the decoder then `num_elems` will be correctly calculated as 2 and all input data will be consumed.

But if the input buffer is somehow truncated:
```
offset
      0   1   2   3   4   5   6   7   8   
      +-----------+---------------+-------|
      | header    | elem0         | elem1 |
      +-----------+---------------+-------|
                                          ^
                                          abnormally truncated at this point
```

then due to integer rounding `num_elems` will be calculated as 1. Only the first element of the array will be decoded (and it will succeed), but the truncated data at the end of the input buffer will not be considered at all.

The correct behaviour in this situation should be for the decoder to notice that the input buffer does not contain an integer number of elements in the variable length array and return an error

# API compatibility

No

## API compatibility plan

N/A

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-948

### DIFF
--- a/generator/sbpg/targets/resources/c/src/sbp_messages_template.c
+++ b/generator/sbpg/targets/resources/c/src/sbp_messages_template.c
@@ -218,6 +218,9 @@ bool (((m.internal_decode_fn)))(sbp_decode_ctx_t *ctx, (((m.type_name))) *msg)
     if (!(((f.decode_fn)))(ctx, &(((field)))[i])) { return false; }
   }
   ((*- elif f.packing == "variable-array" *))
+    if ( ((ctx->buf_len - ctx->offset) % (((f.encoded_len_macro)))) != 0) {
+      return false;
+    }
     msg->(((f.size_fn))) = (uint8_t)((ctx->buf_len - ctx->offset) / (((f.encoded_len_macro))));
   for (uint8_t i = 0; i < msg->(((f.size_fn))); i++) {
     if (!(((f.decode_fn)))(ctx, &(((field)))[i])) { return false; }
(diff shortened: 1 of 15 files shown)


generator/sbpg/targets/resources/c/src/sbp_messages_template.c
```diff
@@ -218,6 +218,9 @@ bool (((m.internal_decode_fn)))(sbp_decode_ctx_t *ctx, (((m.type_name))) *msg)
     if (!(((f.decode_fn)))(ctx, &(((field)))[i])) { return false; }
   }
   ((*- elif f.packing == "variable-array" *))
+    if ( ((ctx->buf_len - ctx->offset) % (((f.encoded_len_macro)))) != 0) {
```
woodfell, line 221: This is the only change in need of review, everything else is generated

woodfell:
Note code coverage is going to fail for the moment, I'm working on increasing it on separate PRs

RReichert (approved):
Nice catch

ghufrankhan7 (approved):
LGTM!

PR #1388, opened 2023-12-08
Add MsgMeasurementPoint [AP-653]

woodfell:
# Description

@swift-nav/devinfra

Create a new message to output profiling information from measurement points.

Current version of some product will output measurement point profiling information in the LinuxCPUState message making use of the `cmdline` field to serialise various bits of information. Some other bits of information are placed in to the other fields in LinuxCpuState but since the fields are already defined the names are misleading.

This PR introduces a dedicated message which contains the same information as is currently sent but broken out in to fields with the correct types. This removes the need to serialise/deserialise this information once products are updated to use this message instead.

# API compatibility

No

## API compatibility plan

No

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-653


spec/tests/yaml/swiftnav/sbp/system/test_MsgMeasurementPoint.yaml
```diff
@@ -0,0 +1,43 @@
...
+      module: sbp.system
+      name: MsgMeasurementPoint
+    msg_type: '0xFF0B'
+    raw_json: '{"crc": 18333, "preamble": 85, "sender": 4096, "length": 48, "msg_type": 65291,
+      "total_time": 2042, "num_executions": 180, "min": 2, "max": 40,
+      "return_addr": 93877475527042, "id": 2496234002,
+      "slice_time": 261963842, "line": 18, "func": "route()\u0000",
+      "payload": "+gcAALQAAgAAACgAAACCyZSNYVUAABKCyZQAAAAAQkCdDwAAAAASAHJvdXRlKCkA"
+      }'
+
```
richarddeurloo, line 34: Nit: this new line seems unnecessary.

richarddeurloo (approved):
LGTM.

pcrumley (approved):

PR #1392, opened 2023-12-14
Bump node.js version to 18.17.0 [AP-1081]

woodfell:
# Description

@swift-nav/devinfra

Bump node.js version in docker image to 18.17.0

This resolves the docker image build failures seen on #1373 and allows a freshly built docker image to build the javascript bindings

# API compatibility

No

## API compatibility plan

N/A

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-1081


pcrumley (approved):

pcrumley (changes requested):
the patch numbers got all goofed. I will try and fix them from you but i don't want to merge it as is.

woodfell:
> the patch numbers got all goofed. I will try and fix them from you but i don't want to merge it as is.

Please do, I spent most of last night trying to do this and gave up. The generator seems to come up with different values when run in CI versus locally. The state of this PR is just what gets through the check

pcrumley:
> > the patch numbers got all goofed. I will try and fix them from you but i don't want to merge it as is.
> 
> Please do, I spent most of last night trying to do this and gave up. The generator seems to come up with different values when run in CI versus locally. The state of this PR is just what gets through the check

the issue is probably that the 5.0.4 tag is on an detached branch... maybe it's better to just have a commit that we take before the release which gets the version numbers aligned.

I also recommend you add a `#no_auto_pr` tag to the PR description so it won't bump a bunch of submodules in our codebase

pcrumley (approved):
Confirming plan here, since v5.0.4 is on a non master branch we will take this commit back to 5.0.3 and then jump ahead to 5.0.5 next release so no tags are clobbered

PR #1393, opened 2023-12-18
Use docker in GA [AP-1085]

woodfell:
# Description

@swift-nav/devinfra

Make use of `libsbp-build` docker image in the "check generated artefacts" stage. This bring consistency to the artefacts generated in the CI check since the canonical way of generating libsbp is via the docker image. This PR doesn't actually make any changes but should resolve some of the issues which have been seen recently when generating libsbp locally versus in CI.

# API compatibility

N/A

## API compatibility plan

N/A

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-1085

### DIFF
--- a/.github/workflows/generator.yaml
+++ b/.github/workflows/generator.yaml
@@ -30,6 +30,10 @@ jobs:
           fetch-depth: 0
           submodules: recursive
 
+      - name: Set safe git directory
+        run: |
+          git config --global --add safe.directory /__w/libsbp/libsbp
+
       - name: Generate tests
         run: |
           rm c/test/auto_check_*.c c/test/cpp/auto_check_*.cc java/test/auto_check_*.java rust/sbp/tests/integration/auto_check_*.rs
--- a/.github/workflows/generator.yaml
+++ b/.github/workflows/generator.yaml
@@ -19,32 +19,12 @@ jobs:
   generation:
     name: Generated artifacts
     runs-on: ubuntu-20.04
+    container: swiftnav/libsbp-build:latest
     steps:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
           submodules: recursive
-
-      - name: Installing dependencies
-        run: |
-          sudo apt-get -qq update
-          sudo apt-get install -y tox python3-sphinx clang-format-6.0 enchant
-          pip3 install tox-run-command
-          tox -e py --notest
-
-      - name: Installing Node
-        uses: actions/setup-node@v4
-        with:
-          node-version: 18
-
-      - name: Installing Quicktype
-        run: |
-          npm ci
-          npm install -g quicktype
-
-      - uses: gradle/gradle-build-action@v2
-        with:
-          gradle-version: 7.1.1
 
       - name: Generate tests
         run: |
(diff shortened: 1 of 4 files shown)


spec/yaml/swiftnav/sbp/acquisition.yaml
```diff
@@ -246,3 +246,5 @@ definitions:
             type: array
             fill: AcqSvProfileDep
             desc: SV profiles during acquisition time
+
```
woodfell, line 249: This is just to force the check generated artefacts stage to run in CI. I'll remove it before merging

.github/workflows/generator.yaml
```diff
@@ -19,32 +19,20 @@ jobs:
...
-
-      - name: Installing Quicktype
-        run: |
-          npm ci
-          npm install -g quicktype
-
-      - uses: gradle/gradle-build-action@v2
-        with:
-          gradle-version: 7.1.1
+          git config --global --add safe.directory /__w/libsbp/libsbp
```
woodfell, line 35: Not sure this is ideal but required to resolve a permissions issue which I think is caused by different UIDs being used inside and outside the container

.github/workflows/generator.yaml
```diff
@@ -19,32 +19,20 @@ jobs:
   generation:
     name: Generated artifacts
     runs-on: ubuntu-20.04
+    container: 
+      image: swiftnav/libsbp-build:latest-master
```
woodfell, line 23: Will need to retain libsbp-build image and update rather than using the master image all the time

woodfell (commented):
Overall brings generation inline with what's produced when docker is used locally. Should resolve annoying issues of late with getting this CI stage to pass

.github/workflows/generator.yaml
pcrumley: I think this is fine/I don't know a better solution here.

pcrumley (approved):
This seems reasonable but I guess I'm curious why you abandoned goal of using `latest` and switched to a tagged version. I guess my concern is that we could run into this issue where the latest build gets broken if there is a delta between the two.

This situation is a lot better than it was, and very happy to see this work

woodfell:
> This seems reasonable but I guess I'm curious why you abandoned goal of using `latest` and switched to a tagged version. I guess my concern is that we could run into this issue where the latest build gets broken if there is a delta between the two.
> 
> This situation is a lot better than it was, and very happy to see this work

It's so that we are able to build historical commit moving forwards. It we somehow bork the `latest-master` tag then older commits could be rendered unbuildable. Similarly, the situation we've just had is that some of the things which are pulled in to the image don't have their versions pinned and `latest-master` doesn't always generate exactly the same artefacts

pcrumley:
> > This seems reasonable but I guess I'm curious why you abandoned goal of using `latest` and switched to a tagged version. I guess my concern is that we could run into this issue where the latest build gets broken if there is a delta between the two.
> > This situation is a lot better than it was, and very happy to see this work
> 
> It's so that we are able to build historical commit moving forwards. It we somehow bork the `latest-master` tag then older commits could be rendered unbuildable. Similarly, the situation we've just had is that some of the things which are pulled in to the image don't have their versions pinned and `latest-master` doesn't always generate exactly the same artefacts

That seems very reasonable, maybe we should put the bumping of the build tag as part of the release process then

PR #1394, opened 2023-12-19
Parallel build all C tests from top level makefile [AP-1086]

woodfell:
# Description

@swift-nav/devinfra

A small change to speed up builds of C tests when using the top level makefile, as is done in all CI stages. 

The current makefile has 2 sections which build and run tests. They are similar (one targets a subset of tests for compatibility with big endian machines). They are both written as:

```
make -j4;
make do-all-tests
```
Where the first command builds all targets in parallel, and the second runs all the tests sequentially.

However, test targets are not built as part of `all` so what really happens is the main `libsbp` target is build in parallel in the first step, then all tests are built and run single threaded in the second stage. This can be easily fixed and sped up by altering the first command to build all test targets using the `-j4` flag:

```
make -j4 build-all-tests;
make do-all-tests
```

# API compatibility

No

## API compatibility plan

No

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-1086

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -378,7 +378,7 @@ test-c:
 	cd $(SWIFTNAV_ROOT)/c; \
 	mkdir -p build/ && cd build/; \
 	cmake $(CMAKEFLAGS) ../; \
-	$(MAKE) -j4; \
+	$(MAKE) -j4 build-all-tests; \
 	$(MAKE) do-all-tests
 	$(call announce-end,"Finished running C tests")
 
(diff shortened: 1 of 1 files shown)


ljbade (approved):

pcrumley:
@woodfell for PRs like this which only affects the libsbp repo it's a good idea to add #no_auto_pr to the title so our Jenkins doesn't do all the submodule updates. Nbd

PR #1400, opened 2024-01-10
Add --unbuffered option to sbp2json.py

dgburr:
# Description

Add a new `--unbuffered` command line option to sbp2json.py which enables a mode where it reads and parses messages one-at-a-time.  The existing behaviour (which remains the default) waits for 4096 bytes before parsing, which means that when decoding real-time streams which only output messages intermittently, there are often large pauses where no output is produced, followed by a large batch of decoded messages at once.

The rust version of sbp2json does not have this issue.  However, there are various use cases (e.g. embedded platforms) where it is difficult (if not impossible) to build the rust tools, but the python version works just fine.

# API compatibility

No.  The previous behaviour remains the default.

## API compatibility plan

N/A

# JIRA Reference


pcrumley (commented):
It would be nice to have a test of a very small message that just lived in buffer land so that it returns when --unbuffered is passed but not without that flag.

Also... I commented in line but I am unsure about this strategy of flushing the buffer by pointing it to a slice in a binary string. I don't know if the earlier memory will actually be freed that way. Do you have some explanation of python that can assuage me there or have some test to prove the buf that isn't part of the slice is garbage collected?

dgburr:
> It would be nice to have a test of a very small message that just lived in buffer land so that it returns when --unbuffered is passed but not without that flag.

To be clear: running with or without the `--unbuffered` flag does not make any difference to the final output which is produced. The only difference when running with the `--unbuffered` flag is the delay between the input and the output when the input rate is very low.

pcrumley:
> > Also... I commented in line but I am unsure about this strategy of flushing the buffer by pointing it to a slice in a binary string. I don't know if the earlier memory will actually be freed that way. Do you have some explanation of python that can assuage me there or have some test to prove the buf that isn't part of the slice is garbage collected?
> 
> From what I can find online (e.g. the first answer to https://stackoverflow.com/questions/64871329/does-string-slicing-perform-copy-in-memory and https://discuss.python.org/t/string-slicing-in-python/25651) it seems that making a slice makes a copy of the original PyObject (except for the the special case where the slice corresponds to the entire string). So the `buf` symbol will be changed to refer to the new PyObject, meaning that the reference count on the old PyObject will drop to 0, resulting in it getting garbage collected.

TIL. I suppose that is why the other implementation uses memoryview. I do think using memory view should be possible here and you can see you are allocating several times per message, do you think this affects performance in a measure way

pcrumley (approved):
Looks good. Thanks for walking me through the PR!

PR #1401, opened 2024-02-01
Add profiling messages [AP-650]

woodfell:
# Description

@swift-nav/devinfra

Add definitions and tests for 3 new messages

- `MSG_PROFILING_RESOURCE_COUNTER`
- `MSG_PROFILING_SYSTEM_INFO`
- `MSG_PROFILING_THREAD_INFO`

These messages are not intended for wide public use, the definitions and significance are defined elsewhere

As always, only the changes under the `spec` directory are interesting, everything else is generated

# API compatibility

No

## API compatibility plan

N/A

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-650


spec/yaml/swiftnav/sbp/profiling.yaml
```diff
@@ -52,3 +52,120 @@ definitions:
...
+          type: u32
+          desc: Number of bytes read from IO handles
+          
+
+ - MSG_PROFILING_RESOURCE_COUNTER:
+    id: 0xCF03
+    short_desc: Information about resource buckets
+    desc: >
+      Information about resource buckets. Refer to product documentation to understand
+      the meaning and values in this message.
```
pcrumley, line 156: is this description correct? I guess it contains buckets but not exactly what I was expecting give the message name

reimerix (approved):

PR #1402, opened 2024-02-01
Disable Java sonar stage [AP-1376]

woodfell:
# Description

@swift-nav/devinfra

There are java version incompatibility issues which are currently failing CI checks. The solution is not forthcoming so simply disable the stage for the moment until a solution is found

To be re-enabled with AP-1377

# API compatibility

No

## API compatibility plan

N/A

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-1376

### DIFF
--- a/.github/workflows/java.yaml
+++ b/.github/workflows/java.yaml
@@ -23,3 +23,21 @@ jobs:
       - name: Run tests
         run: make test-java
 
+# sonarqube:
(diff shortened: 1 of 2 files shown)


howardfernandez (approved):

pcrumley (approved):
looks good, someday we can fix this

PR #1404, opened 2024-02-15
Remove legacy C API [AP-1684]

woodfell:
# Removal Warning

This PR will be merged on **8/4/2024**. On this date the legacy C API will become unavailable, it will not be reintroduced. On or shortly after this date libsbp version v6.0.0 will be tagged and released

# Description

@swift-nav/devinfra

Remove legacy C API including all generated code, examples, templates, and sbpg module.

Remove all references to "v4" API which were lying around in a handful of places. There is no such thing as the V4 API any more, there is just the libsbp API

# API compatibility

Yes

Legacy C API completely removed including:
- Any message types in the form `struct msg_xxx_t`
- Functions relating to sending or processing encoded payloads or frames
- payload and frame callback types
- `#define`s message types - in the form `SBP_MSG_XXX`
- C++ `LegacyState` wrapper
- All files under the `libsbp/legacy` directory
- V4 compatibility headers for modern API in `libsbp/v4/<package.h>`

## API compatibility plan

The modern API was introduce to this repository in Sep 2021. The legacy API was retained but all symbols were marked as deprecated and compile time messages were output by GCC and Clang whenever a legacy header was used

In Dec 2023 the legacy was moved to a more difficult to access place. It was no longer accessible simply by include `libsbp/sbp.h` All headers were retained but even more compile time messages were added indicating that these files would be deleted in the next major version

This PR is the final stage of the deprecation plan. Any users of this repository should have already converted their code to use the moden API.

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-1684


sokhealy:
does this file also need to be removed?
https://github.com/swift-nav/libsbp/blob/woodfell/remove_legacy_api/generator/sbpg/targets/test_legacy_c.py

sokhealy:
Do we have to change any of the header guards?
`LIBSBP_V4_`

c/include/libsbp/orientation_macros.h
```diff
@@ -37,12 +36,10 @@
 #define SBP_BASELINE_HEADING_FIX_MODE_FLOAT_RTK (3)
 #define SBP_BASELINE_HEADING_FIX_MODE_FIXED_RTK (4)
 /**
- * Encoded length of sbp_msg_baseline_heading_t (V4 API) and
- * msg_baseline_heading_t (legacy API)
+ * Encoded length of sbp_msg_baseline_heading_t (V4 API)
```
sokhealy, line 39: did we also want to remove any reference to `V4 API`?

sokhealy (approved):
Besides the above comments I don't see any issue with the changes here.

PR #1405, opened 2024-03-15
[IOTE-6][IOTE-326] Add weeks behavior flag to MSG_GNSS_TIME_OFFSET

jakalm:
# Description

@swift-nav/devinfra

Adding a new flag to the reserved `flags` field in the `MSG_GNSS_TIME_OFFSET` message to flag if the timestamp is affected by local timestamp rollover or not. For more information, see https://swift-nav.atlassian.net/wiki/spaces/~622a65696a4c4c0070b2567a/pages/3060662918/Amazon+INS+timestamp+issue+way+forward#Conclusion-and-decision.

# API compatibility

No, this has been a reserved field so should not be incompatible with current implementations of the protocol.

## API compatibility plan

# JIRA Reference

https://swift-nav.atlassian.net/browse/IOTE-325


reimerix (approved):

spec/yaml/swiftnav/sbp/system.yaml
```diff
@@ -443,7 +443,15 @@ definitions:
...
-            desc: Status flags (reserved)
+            desc: Status flags
+            fields:
+              - 1-7:
+                  desc: Reserved, set to zero
+              - 0:
+                  desc: Weeks behavior
+                  values:
+                    - 0: Not affected on local timestamp rollover
+                    - 1: Incremented on local timestamp rollover
```
sokhealy, line 454: Apologies if I am confused, but the description seems wrong? A value of 1 should have Starling ignore `MSG_GNSS_TIME_OFFSET` completely? In this case, we are sending both `SBP_MSG_PPS_TIME` and `MSG_GNSS_TIME_OFFSET`, can both co-exist without us ignoring one of them?

c/include/libsbp/system_macros.h
```diff
@@ -856,6 +856,38 @@
...
+    (flags) =                                                              \
+        (u8)((flags & (~(SBP_GNSS_TIME_OFFSET_WEEKS_BEHAVIOR_MASK          \
+                         << SBP_GNSS_TIME_OFFSET_WEEKS_BEHAVIOR_SHIFT))) | \
+             (((val) & (SBP_GNSS_TIME_OFFSET_WEEKS_BEHAVIOR_MASK))         \
+              << (SBP_GNSS_TIME_OFFSET_WEEKS_BEHAVIOR_SHIFT)));            \
+  } while (0)
+
+#define SBP_GNSS_TIME_OFFSET_WEEKS_BEHAVIOR_NOT_AFFECTED_ON_LOCAL_TIMESTAMP_ROLLOVER \
+  (0)
+#define SBP_GNSS_TIME_OFFSET_WEEKS_BEHAVIOR_INCREMENTED_ON_LOCAL_TIMESTAMP_ROLLOVER \
```
sokhealy, line 889: I think we should change the names of these defines because it seems contrary to how they function in starling.

spec/yaml/swiftnav/sbp/system.yaml
jakalm: If the flag is set to 1, Starling will ignore the message. But the description of the flag describes how the weeks field will work. Do you mean that we should explicitly state that the message will be ignored by Starling?

I’m not sure if they can co-exist without that flag being set, but that should not have changed with these changes. It will work the same way as before.

sokhealy: On second thought, this could be independent from our PE and better describe the message itself.  Our PE can choose to either ignore the message or use it, but that would be independent from the description.

c/include/libsbp/system_macros.h
jakalm: Sorry, maybe I misunderstand your question.

They are describing how the weeks field works. If set to one, the field will be incremented on local timestamp rollover, and therefore ignored by starling. If set to zero, the weeks field will not be affected by local timestamp rollover, and will be used by Starling.

Or do you mean that we should explicitly state that it will be ignored by Starling?

sokhealy: Ignore the above comment.

sokhealy (approved):
Looks good, apologies for my confusion.

isaactorz (approved):

ljbade (approved):

PR #1421, opened 2024-08-12
Fix license

RReichert:
# Description

While reading over libsbp for another task, realized that pretty much all the license files had the grammar mistake of an extra `be`. To avoid further copy/paste errors, I've tried to fix them all.

Also realized that when launching cmake I kept getting warnings due to not prefixing the `sbp` library reference with `swiftnav::` (ex: `c/test/string/CMakeLists.txt`), fix that as well.

# API compatibility

No compatability issues

## API compatibility plan

N/A

# JIRA Reference

None


pcrumley (approved):
:shipit:

richarddeurloo (approved):
Thanks for the clean-up. This is not the first repo that needed this fix unfortunately.

PR #1424, opened 2024-08-15
fix rust lints

pcrumley:
# Description

@swift-nav/devinfra

A simple whitespace fix to get rust lints passing

# API compatibility

No

## API compatibility plan

N/A

# JIRA Reference

N/A

### DIFF
--- a/javascript/sbp/observation.js
+++ b/javascript/sbp/observation.js
@@ -2495,7 +2495,7 @@ MsgGroupDelay.prototype.fieldSpec.push(['isc_l2c', 'writeInt16LE', 2]);
  * @field health_bits number (unsigned 8-bit int, 1 byte) Satellite health status for GPS:   - bits 5-7: NAV data health status. See IS-
  *   GPS-200H     Table 20-VII: NAV Data Health Indications.   - bits 0-4: Signal
  *   health status. See IS-GPS-200H     Table 20-VIII. Codes for Health of SV Signal
- *   Components. Satellite health status for GLO (see GLO ICD 5.1 table 5.1 for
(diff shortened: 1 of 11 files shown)


ljbade (approved):

javascript/sbp/observation.js
```diff
@@ -2495,7 +2495,7 @@ MsgGroupDelay.prototype.fieldSpec.push(['isc_l2c', 'writeInt16LE', 2]);
  * @field health_bits number (unsigned 8-bit int, 1 byte) Satellite health status for GPS:   - bits 5-7: NAV data health status. See IS-
  *   GPS-200H     Table 20-VII: NAV Data Health Indications.   - bits 0-4: Signal
  *   health status. See IS-GPS-200H     Table 20-VIII. Codes for Health of SV Signal
- *   Components. Satellite health status for GLO (see GLO ICD 5.1 table 5.1 for
+ *   Components.  Satellite health status for GLO (see GLO ICD 5.1 table 5.1 for
```
RReichert, line 2498: strange that extra spacing is necissary

RReichert (approved):
I would have assumed that by running `make gen-all` I would have given me this result if I were to have checked out the `master` branch. Did you have to run something separate outside of what `make` offers to get it to format for you?

sokhealy (approved):

javascript/sbp/observation.js
pcrumley: A new clippy lint was added to rust which made the code-gen fail. I had two choices:
1. update the yaml describing the messages so it passed the lint by putting in a empty line in the description
2. make it so clippy lints (or just that new lint) aren't applied to generated code.

I chose option one hence the addition of whitespace in all the languages. Weird that newline because a single space there.

PR #1426, opened 2024-08-15
Remove haskell from bench

pcrumley:
# Description

@swift-nav/devinfra

the Haskell benchmarking pipeline is failing. This PR removes that check.

# API compatibility

No

## API compatibility plan

None

# JIRA Reference

N/A


pcrumley:
this sadly doesn't work

RReichert:
can now close this since https://github.com/swift-nav/libsbp/pull/1454 solves the problem which I believe this was trying to solve

PR #1434, opened 2024-09-05
Revert "Pcrumley/update actions"

pcrumley:
Reverts swift-nav/libsbp#1428

Auto-merge may have been a little bit premature

PR #1435, opened 2024-09-05
Pcrumley/update actions

pcrumley:
# Description

@swift-nav/algint-team

actions/checkout@v4 fails on ubuntu 18:04 because of an old Node version, reverting to v2.

Doing this instead of: https://github.com/swift-nav/libsbp/pull/1434

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+---
+version: 2
+updates:
+# Set update schedule for GitHub Actions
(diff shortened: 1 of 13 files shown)


pcrumley:
@jbangelo I'm just going to downgrade that 1 action back to v2 for now. If you can see the version of Node used in that step is quite old.

jbangelo (approved):
🚀

PR #1439, opened 2024-09-14
Documentation fixes after the removal of the legacy C API

dgburr:
# Description

@swift-nav/algint-team

* `sbp_register_callback()` and `sbp_register_frame_callback()` no longer exist.  They have been replaced with `sbp_callback_register()`.
* `sbp_msg_callback_t` has changed to `sbp_callback_t` and it has a different function prototype
* `sbp_frame_callback_t` has been removed

# API compatibility

No - it only changes documentation.

## API compatibility plan

N/A

# JIRA Reference

The removal of the legacy API is tracked in https://swift-nav.atlassian.net/browse/AP-1684

### DIFF
--- a/c/src/sbp.c
+++ b/c/src/sbp.c
@@ -29,19 +29,10 @@
  * ---------
  *
  * First setup a callback for the message you will be receiving. Our callback
- * function must have type #sbp_msg_callback_t or #sbp_frame_callback_t,
- * i.e. it must be of the form:
+ * function must have type #sbp_callback_t i.e. it must be of the form:
  *
  * ~~~
- * void my_callback(u16 sender_id, u8 len, u8 msg[], void *context)
- * {
- *   // Process msg.
- * }
- * ~~~
- * or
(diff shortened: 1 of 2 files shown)


c/src/sbp.c
```diff
@@ -29,27 +29,18 @@
  * ---------
  *
  * First setup a callback for the message you will be receiving. Our callback
- * function must have type #sbp_msg_callback_t or #sbp_frame_callback_t,
- * i.e. it must be of the form:
+ * function must have type #sbp_callback_t i.e. it must be of the form:
```
Madhunika-atmakuri, line 32: I'm not too aware of the API changes , in the description of this PR it is written that `sbp_msg_callback_t` has been changed to `sbp_callback_t`, what about `sbp_frame_callback_t`? Has that been removed too?

Madhunika-atmakuri (approved):
Just left a quick question for clarification, rest all looks good :+1:

PR #1441, opened 2024-09-25
Add compensated IMU message [AP-1475]

ismolyakov:
# Description

@swift-nav/algint-team

# API compatibility

## API compatibility plan

# JIRA Reference


spec/yaml/swiftnav/sbp/imu.yaml
```diff
@@ -115,3 +115,53 @@ definitions:
...
+        - time:
+            type: u64
+            units: microseconds
+            desc: Microseconds since reference epoch
+        - flags:
+            type: u16
+            desc: Contains the applied compensation parameters and time synchronization mode
+            fields:
+              - 2-15: 
+                  desc: Reserved
```
reimerix, line 136: I'd add flags for the following:

- gravity compensated
- earth rate compensated
- coriolis effect compensated
- accelerometer errors compensated
- gyroscope errors compensated
- maybe: transformed to VRP

spec/yaml/swiftnav/sbp/imu.yaml
```diff
@@ -115,3 +115,53 @@ definitions:
...
+              - 0-1:
+                  desc: Time status
+                  values:
+                    - 0: Reference epoch is start of current GPS week
+                    - 1: Reference epoch is time of system startup
+                    - 2: Reference epoch is unknown
+                    - 3: Reference epoch is last PPS  
+        - acc_comp_x:
+            type: s32
+            units: 1e-2 m/s^2
```
reimerix, line 146: If using 32 bits, resolution could well be on the order of 1e-6 m/s^2

spec/yaml/swiftnav/sbp/imu.yaml
```diff
@@ -115,3 +115,53 @@ definitions:
...
+            desc: Compensated acceleration Y axis
+        - acc_comp_z:
+            type: s32
+            units: 1e-2 m/s^2
+            desc: Compensated acceleration Z axis
+        - gyr_comp_x:
+            type: s32
+            units: 1e-3 deg/s
```
reimerix, line 158: With 32 bits, much better resolution should be possible, e.g. 0.000003 deg/s in case we'd want to deal with much better IMUs

spec/yaml/swiftnav/sbp/imu.yaml
```diff
@@ -115,3 +115,63 @@ definitions:
...
                       - 3: +/- 16g
                       - 4: +/- 6g
+
+ - MSG_IMU_COMP:
+    id: 0x0905
+    short_desc: Compensated IMU data
+    desc: >
+      Data from the Inertial Measurement Unit, containing accelerometer and
+      gyroscope readings compensated for estimated errors and constant 
+      physical effects. The output is valid for the center of navigation.
```
junc52, line 125: For being unambiguous, let's specify object frame, reference frame, and resolving frame. How about this?
'The output is valid for inertially referenced center of navigation(imu body frame) represented in vehicle body frame'

junc52 (approved):
It looks good to me!
And have a question, if we change these, other dependency codes can be generated automatically? Just in case, It would be good to check and share the automation mechanism at demo session?

spec/yaml/swiftnav/sbp/imu.yaml
ismolyakov: corrected! 

If we ever need to support a transformation to VRP, we will use one of the reserved flags and adjust this description.

PR #1446, opened 2024-10-01
Add compensated IMU message [AP-1475]

ismolyakov:
# Description

Adding a new MSG_IMU_COMP according to the corresponding design doc. 

Screenshot 2024-10-01 at 8 57 33 PM

# API compatibility

No, this is a new message

## API compatibility plan

n/a

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-1475


junc52 (approved):
Thanks for resolving challenging work!
Question : so updating imu.yaml will produce all these changes automatically?

ismolyakov:
> Thanks for resolving challenging work! Question : so updating imu.yaml will produce all these changes automatically?

Yes! the intended process is:
- define the message specs
- run `make all` and all code _should_ be generated (this is not what I encountered: while the scripts of a given Docker image on a given machine may succeed, the pipeline items may fail - I am still in progress to identify the best recipe that leads to passing pipeline)

pcrumley (changes requested):
Can you remove the link to the internal confluence documentation from you PR description?

I'm also curious about this statement: "This PR doesn't include a corresponding unit test due to test generation problems. It is proposed to merge as is to unblock development and address test generation as the next step."  Do you need help/advice on how to do the unit test or are you struggling to generate the input data for the messages?

Finally, can you close all the draft PRs? 
#1441
#1442 
#1443 
#1444
#1445

pcrumley (approved):

PR #1448, opened 2024-10-03
get the automated generation of test cases working again

pcrumley:
# Description

@swift-nav/algint-team

This pr fixes the json2test script using UV, which installs dependencies on the fly. I tried hard to make it so that it could easily be called from any directory & link to the local version of SBP, but throwing in the towel under the theory that working but less than ideal is better than not working.


generator/sbpg/test_structs.py
```diff
@@ -99,7 +99,8 @@ def __init__(self, raw_packet, msg_type, raw_json, msg, sbp, test_msg_data=None)
     self.payload = base64.standard_b64decode(self.sbp['payload'])
     self.payload_as_byte_array = list(bytearray(self.payload))
     self.payload_len_for_encoding_buf = len(self.payload) if len(self.payload) > 0 else 1
-    msg_name = msg['name']
+    if msg_name is None:
+        msg_name = msg['name']
```
pcrumley, line 103: I think this is something that got broken in json2test. this should only be used in the `from_msg` function and otherwise have the same behavior as before

ismolyakov (approved):
lgtm! gave it a go and it works :)

PR #1452, opened 2024-10-15
cover CI hole which appears to be hiding a proc_macro bindgen bug

pcrumley:
# Description

@swift-nav/algint-team

Prior to this PR, we only tested `cargo build --all-targets --all-features` on Linux machines. This seems to be a gap where inside of a proc-macro in the `swiftnav-rs` which can fail on different OSes. After we take this PR we no longer will have this gap.

Note, this is possibly not the best place to put this check, it may make more sense to have the ci, lint format run on a matrix instead.

# API compatibility

None

## API compatibility plan

No

# JIRA Reference

None

### DIFF
--- a/scripts/ci_build_rust.bash
+++ b/scripts/ci_build_rust.bash
@@ -22,6 +22,7 @@ else
 fi
 
 cargo build --no-default-features -p sbp
+cargo build --no-default-features -p sbp --features swiftnav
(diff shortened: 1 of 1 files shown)


sbmueller:
Nit: If there is a ticket for this change, it could be linked in the PR title

sbmueller (approved):

PR #1453, opened 2024-10-22
Fix manual file generation and include trailing newlines in accordance with POSIX [AP-3048]

sbmueller:
# Description

@swift-nav/algint-team

# API compatibility

No, this change is only fixing non-functional issues in the project.

## API compatibility plan

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-3048


generator/sbpg/targets/templating.py
```diff
@@ -25,6 +25,7 @@
                           variable_end_string = ')))',
                           comment_start_string = '((=',
                           comment_end_string = '=))',
+                          keep_trailing_newline=True,
```
pcrumley, line 28: @RReichert ☝🏻

sbmueller:
I can't modify the Dockerfile since the "build docker image" stage always fails when installing the Python dependencies :(

pcrumley (approved):
:shipit:

Madhunika-atmakuri (approved):
LGTM. Thanks @sbmueller

PR #1455, opened 2024-10-22
Upgrade Haskel version

RReichert:


junc52 (approved):
Thanks for the fix and support!

PR #1456, opened 2024-10-22
put the cargo dep in the right place

pcrumley:
# Description

@swift-nav/algint-team

Fixes a bug i added which hand edited a auto-gen'ed file, & fixes an unrelated typo

# API compatibility

No

## API compatibility plan

Non

# JIRA Reference


junc52 (approved):
Thanks for the fix and support!

RReichert (approved):

PR #1466, opened 2024-12-04
bump macos to 13

sokhealy:
# Description

MacOs-12 had been deprecated see here:
https://github.com/actions/runner-images/issues/10721

@swift-nav/algint-team

# API compatibility

No changes to API, Just changing MacOs from 12 to 13.

## API compatibility plan

None

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-3230

### DIFF
--- a/.github/workflows/c.yaml
+++ b/.github/workflows/c.yaml
@@ -82,7 +82,7 @@ jobs:
 
   macos:
     name: macOS
-    runs-on: macos-12
+    runs-on: macos-13
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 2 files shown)


sokhealy:
> Note there is this [ticket](https://swift-nav.atlassian.net/browse/AP-3178) concerning the failures on Ubuntu 18. For the rust job, I'm not sure. Clippy dislikes a lot of the code. Was it always like this?

I will leave the ubuntu 18 issue for another day, it looks like we are compiling with GCC 6 / Clang 6. I don't think ubuntu 22.04 or 24.04 will have such an old compiler.

PR #1469, opened 2025-01-23
fix rust lints

pcrumley:
# Description

@swift-nav/algint-team

Fixes a new rust lint in a test

# API compatibility

None

## API compatibility plan

N/A


pcrumley:
when i created this branch I thought the CI was using the deprecated macos-12 runners, but it actually is using macos-13, so I left it as is, but the branch name is incorrect

sbmueller (approved):

notoriaga (approved):

PR #1470, opened 2025-02-03
Update NodeJS + Haskell version

RReichert:
# Description

Fix the docker build

See: https://github.com/swift-nav/docker-recipes/pull/439

# API compatibility

No compatibility issues

## API compatibility plan

N/A


Dockerfile
```diff
@@ -154,7 +154,7 @@ USER dockerdev
 
 RUN \
   if [ "$(ls /tmp)" ]; then ls /tmp; false; fi \
-  && stack install --resolver lts-10.10 sbp \
+  && stack install --resolver lts-23.7 sbp \
```
sbmueller, line 157: Hey, I think I tried it like this in the past, but it didn't work. I remember you did it like [this](https://github.com/swift-nav/libsbp/commit/c9df176a2c930b19d5754128e77640d8069b4440) previously. Can you double-check?

RReichert: From what I understand, the `Dockerfile` explicitly states what version it wants to use to build the SBP version, where as in the [Haskell test build](https://github.com/swift-nav/libsbp/blob/9daf121f29351f5b91684cbc1b81268b845fd351/.github/workflows/haskell.yaml#L49) it uses whatever is in that `haskell/stack.yaml` file. I don't know why there is a difference there.

RReichert: The problem strangely that I'm seeing with my PR is that Haskell and NodeJS are all working, but its python that keeps failing now. I've rolled back the Haskell from `23.7` to `14.9` and it still fails. 

image

python/tests/sbp/client/test_logger.py
```diff
@@ -75,25 +75,6 @@ def test_non_utf8_json_log():
...
-  with open(log_datafile, 'r') as infile:
-    with JSONLogIterator(infile, conventional=conventional) as log:
-      with warnings.catch_warnings(record=True) as w:
-        for _, _ in fetch_next(log):
-          pass
-        warnings.simplefilter("always")
-        # Check for warnings.
-        assert len(w) == 1
-        assert issubclass(w[0].category, RuntimeWarning)
-        assert str(w[0].message).startswith('Bad message parsing for line')
```
RReichert, line 89: This is a strange test and is dependent on the content of the python JSON library underneath. looks like the earlier versions of JSON library might have worked here, but its not anymore.

package-lock.json
RReichert, line 1: Needed to regenerated the nodejs list of packages, the old versions were unsupported with the latest LTS

package.json
```diff
@@ -14,10 +14,13 @@
...
   "license": "MIT",
   "dependencies": {
+    "assert": "^2.1.0",
     "binary-parser": "^1.7.0",
     "cuint": "^0.2.2",
     "js-yaml": "^3.13.1",
-    "node-int64": "^0.4.0"
+    "node-int64": "^0.4.0",
+    "path": "^0.12.7",
+    "stream": "^0.0.3"
```
RReichert, line 23: new packages were required to generate `javascript/sbp.bundle.js`

Dockerfile
```diff
@@ -20,7 +20,7 @@ ARG UBUNTU_RELEASE
 
 ARG DEBIAN_FRONTEND=noninteractive
 
-ENV NODE_VERSION=v18.17.0
+ENV NODE_VERSION=v22.13.1
```
RReichert, line 23: Using the latest NodeJS TLS

Dockerfile
```diff
@@ -79,10 +79,10 @@ RUN \
       imagemagick \
       enchant \
       clang-format-6.0 \
-      python3.7 python3.7-dev python3.7-distutils \
```
RReichert, line 82: Python 3.7 is no longer working, the build keep reporting:

image

RReichert: Python 3.7 and python 3.8 are now at EOL

Dockerfile
```diff
@@ -79,10 +79,10 @@ RUN \
...
       python3 python3-dev python3-distutils \
       python3.9 python3.9-dev python3.9-distutils \
       python3.10 python3.10-dev python3.10-distutils \
+      python3.11 python3.11-dev python3.11-distutils \
```
RReichert, line 85: I've included python 3.11 since its the latest python version that the Ubuntu Focal offers

javascript/sbp.bundle.js
RReichert, line 1: Needed to regenerate after updating the NodeJS version

python/tox.ini
```diff
@@ -19,31 +19,6 @@ deps =
...
-    # Don't complain if non-runnable code isn't run:
-    if 0:
-    if __name__ == .__main__.:
-ignore_errors = True
-
-[coverage:omit]
-omit =
-    tests/*
-    limbo/*
-    data/*
```
RReichert, line 49: This was the source of a lot of headaches. The python tests would randomly crash because of some internal Python code coverage error, it would normally happen with Python 3.10 builds, but not exclusively:

image

There is no need for code coverage here, we are looking to run the test cases and see if they work. If we need code coverage, we can add it later with SonarCloud integration.

Dockerfile
RReichert: I've since reverted this change and tried to unify the Haskell version to `lts-14.9`

python/tox.ini
sbmueller: Maybe someone else from @swift-nav/algint-team can confirm that we don't require coverage right now? That would be my only concern.

PR #1473, opened 2025-02-12
Update from Ubuntu 18 (gcc-6, clang-6) to Ubuntu 20 (gcc-7, clang-6) [AP-3178]

sbmueller:
# Description

@swift-nav/algint-team

This PR updates the C stages in the CI from relying on Ubuntu 18 (gcc-6, clang-6) to Ubuntu 20 (gcc-7, clang-6). Ubuntu 18 is EOL since May 2023. Recently, we observed issues, e.g. in GitHub Actions due to old versions of GLIBC in Ubuntu 18 jobs. We decided to not further invest time and energy in maintaining these old pipelines, because:
- No customer directly relies on gcc-6.
- We are planning to update our codebase to C++17, which requires at least gcc-8 (or 7 without CTAD).
- If we want to have full coverage for all customer required toolchains, that is a different topic and an additional effort. Currently, that is not done for dependency repositories like `libsbp`.
- To still have a certain confidence that gcc versions down to 7 are supported, this PR updates to Ubuntu 20, which is still maintained and offers gcc 7 in its apt repositories.

# API compatibility

No, there are no changes to the code.

## API compatibility plan

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-3178

### DIFF
--- a/.github/workflows/c.yaml
+++ b/.github/workflows/c.yaml
@@ -18,28 +18,29 @@ jobs:
         build_type: ["Debug", "Release"]
         compiler:
           [
-            { c: "gcc-14", cxx: "g++-14", package: "gcc-14 g++-14" },
-            { c: "clang-18", cxx: "clang++-18", package: "clang-18" },
+            { c: "gcc-7", cxx: "g++-7", package: "gcc-7 g++-7" },
+            { c: "clang-6.0", cxx: "clang++-6.0", package: "clang-6.0" },
           ]
         build_shared_libraries: [true, false]
 
-    name: "Ubuntu 24.04 (Build: ${{ matrix.build_type }}, Compilers: ${{ matrix.compiler.c }}/${{ matrix.compiler.cxx }}, Shared Library: ${{ matrix.build_shared_libraries }})"
+    name: "Ubuntu 20.04 (Build: ${{ matrix.build_type }}, Compilers: ${{ matrix.compiler.c }}/${{ matrix.compiler.cxx }}, Shared Library: ${{ matrix.build_shared_libraries }})"
 
-    runs-on: ubuntu-24.04
+    # We run on a docker image to avoid being affected from GitHub
+    # decommissioning certain runners
+    runs-on: ubuntu-latest
+    container: ubuntu:20.04
 
     steps:
       - name: Setup
         run: |
-          sudo bash -c '
           apt-get -qq update
           apt-get -qq install software-properties-common gpg wget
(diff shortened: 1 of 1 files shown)


.github/workflows/c.yaml
```diff
@@ -12,38 +12,34 @@ on:
...
-
     strategy:
       matrix:
-        build_type: [ "Debug", "Release" ]
-        compiler: [
-          { c: "gcc-6", cxx: "g++-6", package: "gcc-6 g++-6" },
-          { c: "clang-6.0", cxx: "clang++-6.0", package: "clang-6.0" }
-        ]
-        build_shared_libraries: [ true, false ]
-        exclude: [
```
sbmueller, line 26: I digged in the history and this combination was excluded because at some point, there was a "code coverage" job that covered this combination. This job is not existing anymore, so I'm adding the combination here.

.github/workflows/c.yaml
```diff
@@ -12,38 +12,34 @@ on:
...
+            { c: "clang-18", cxx: "clang++-18", package: "clang-18" },
+          ]
+        build_shared_libraries: [true, false]
 
-    name: "Ubuntu 18.04 (Build: ${{ matrix.build_type }}, Compilers: ${{ matrix.compiler.c }}/${{ matrix.compiler.cxx }}, Shared Library: ${{ matrix.build_shared_libraries }})"
+    name: "Ubuntu 24.04 (Build: ${{ matrix.build_type }}, Compilers: ${{ matrix.compiler.c }}/${{ matrix.compiler.cxx }}, Shared Library: ${{ matrix.build_shared_libraries }})"
 
-    runs-on: ubuntu-latest
-    container: ubuntu:18.04
+    runs-on: ubuntu-24.04
```
RReichert, line 28: I would say we shouldn't run on the Ubuntu 24 instance, we should always use a container within the runner, the runner's version can always be the latest. The reason for this is to avoid finding ourselves in a position where Github no longer supports a version of a runner and our builds breaks, which happens regularly enough. We you could argue that the container would suffer similar issues as we always calls the `apt-get` to pull in packages from a remote source but the remote source support things for many years, we are still running 18.04 without a problem (apart from the github action which recently was fixed up).

sbmueller: Roger that, corrected

RReichert (approved):

marcoamm:
@sbmueller Ubuntu 20 will EOL in April this year. Would it be worth it to try to bump it to 22 or 24?

PR #1485, opened 2025-07-09
Fix rust build

armallen:
# Description

# API compatibility

## API compatibility plan

# JIRA Reference

### DIFF
--- a/generator/sbpg/targets/resources/rust/test/sbp_tests_template.rs
+++ b/generator/sbpg/targets/resources/rust/test/sbp_tests_template.rs
@@ -51,7 +51,7 @@ fn test_(((s.suite_name|snake_case)))()
         match &sbp_msg {
             sbp::messages::Sbp::(((t.msg.name|lower_acronyms)))(msg) => {
                 let msg_type = msg.message_type().unwrap();
-                assert_eq!( msg_type, (((t.msg_type))), "Incorrect message type, expected (((t.msg_type))), is {}", msg_type);
+                assert_eq!( msg_type, (((t.msg_type))), "Incorrect message type, expected (((t.msg_type))), is {msg_type}");
                 let sender_id = msg.sender_id().unwrap();
                 assert_eq!(sender_id, (((t.sbp.sender))), "incorrect sender id, expected (((t.sbp.sender))), is {sender_id}");
                 ((*- for f in t.fieldskeys *))(((compare_value( (((f))), (((t.fields[f]))) ))))((*- endfor *))
--- a/rust/sbp/src/sbp_iter_ext.rs
+++ b/rust/sbp/src/sbp_iter_ext.rs
@@ -23,7 +23,7 @@ pub trait SbpIterExt: Iterator {
         Self: Iterator<Item = Result<Sbp, DeserializeError>> + Sized,
     {
         HandleErrorsIter::new(self, move |e| {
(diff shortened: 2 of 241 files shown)


.github/workflows/generator.yaml
```diff
@@ -18,8 +18,8 @@ on:
 jobs:
   generation:
     name: Generated artifacts
-    runs-on: ubuntu-20.04
-    container: 
+    runs-on: ubuntu-24.04
```
armallen, line 21: I believe this workflow can run on anything since it's using a container below

generator/sbpg/targets/resources/rust/test/sbp_tests_template.rs
```diff
@@ -51,7 +51,7 @@ fn test_(((s.suite_name|snake_case)))()
         match &sbp_msg {
             sbp::messages::Sbp::(((t.msg.name|lower_acronyms)))(msg) => {
                 let msg_type = msg.message_type().unwrap();
-                assert_eq!( msg_type, (((t.msg_type))), "Incorrect message type, expected (((t.msg_type))), is {}", msg_type);
+                assert_eq!( msg_type, (((t.msg_type))), "Incorrect message type, expected (((t.msg_type))), is {msg_type}");
```
armallen, line 54: msg_type moved in between the curly braces to fix all lints in automatically generated tests

armallen:
Just noticed benchmarks were running on an old `xlarge` runner, trying it with `xlarge-24.04`?

.github/workflows/python.yaml
```diff
@@ -18,7 +18,7 @@ on:
 jobs:
   coverage:
     name: Test
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
armallen, line 21: Updated to ubuntu-22.04 because the python dependencies build procedures are not portable and no wheels are available for newer OS versions (pip install fails on ubuntu-24.04 with default python version).

Wheels are available for ubuntu-22.04.

.github/workflows/rust.yaml
```diff
@@ -64,17 +64,22 @@ jobs:
...
           - windows-2022
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
 
+      - name: Setup cmake
+        uses: jwlawson/actions-setup-cmake@v2
+        with:
+          cmake-version: "3.31.2"
+
```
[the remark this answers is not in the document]
sbmueller: I think the minimum CMake version used anywhere right now should be 3.13. So maybe the solution is rather to update this repo's demands?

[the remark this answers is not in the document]
armallen: The issue occurs [here](https://github.com/swift-nav/libsbp/actions/runs/16174054830/job/45654941886#:~:text=error%3A%20failed%20to%20run%20custom%20build%20command%20for%20%60swiftnav%2Dsys%20v0.10.0%60)
which is in the rust build of `swiftnav-rs`, which seems to be pulling/building `libswiftnav` using a [generated cmake file ](https://github.com/swift-nav/swiftnav-rs/blob/b6e22fc7b5197b499ffd241eee1dadf6ea8e4ebb/swiftnav-sys/build.rs#L9)

I don't have enough motivation to investigate/fix that today 😆

PR #1486, opened 2025-07-17
Update rust Docker image to rust:1.88.0-slim

dgburr:
# Description

Update rust Docker image to the latest version (rust:1.88.0-slim).  This fixes the following build error when trying to compile sbp2json:

```
error: failed to compile `sbp2json v6.2.2-alpha (https://github.com/swift-nav/libsbp.git#635fc94f)`, intermediate artifacts can be found at `/tmp/cargo-installwOcSKv`

Caused by:
  package `textwrap v0.16.2` cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.67.1
  Either upgrade to rustc 1.70 or newer, or use
  cargo update -p textwrap@0.16.2 --precise ver
  where `ver` is the latest version of `textwrap` supporting rustc 1.67.1
```

Also bump the haskell and python benchmarking Docker images to debian:bookworm-slim to fix the benchmark step in CI.

# API compatibility

No

## API compatibility plan

N/A

# JIRA Reference

### DIFF
--- a/.github/workflows/haskell.yaml
+++ b/.github/workflows/haskell.yaml
@@ -14,7 +14,7 @@ on:
 jobs:
   build:
     name: Build
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
 
--- a/haskell/Dockerfile
+++ b/haskell/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:buster-slim
+FROM debian:bookworm-slim
 
 RUN \
   apt-get update && \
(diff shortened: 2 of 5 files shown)


Madhunika-atmakuri (approved):

armallen:
@dgburr  The Haskell step can't run because it's using ubuntu-20.04 runner, which are not available anymore.
Updating the workflow to run on `ubuntu-24.04` fixes the issue (see change [here](https://github.com/swift-nav/libsbp/pull/1487/files#diff-149989f541cb9b71ff7f7a9059f9e484b535d575192f4dc440aa03adab5c1e23) and workflow result [here](https://github.com/swift-nav/libsbp/actions/runs/16367126387/job/46246812767)

armallen (approved):

pcrumley (approved):

PR #1489, opened 2025-07-25
Implement MODULE [AP-4083]

martin4861:
# Description

@swift-nav/algint-team

Implement a Bazel 7 compatible `MODULE` file.
Uses https://github.com/swift-nav/rules_swiftnav/pull/167

The `MODULE` file is sufficient to consume `libsbp`. However, building `libsbp` using the `MODULE` is not yet supported on all platforms (e.g. darwin aarch64 is not yet working). The local development in this repo is still on Bazel 6/`WORKSPACE`, therefore nothing changes.

Needed `rules_swiftnav` changes: https://github.com/swift-nav/rules_swiftnav/pull/167

# API compatibility

No!

## API compatibility plan

Not needed.

# JIRA Reference

https://swift-nav.atlassian.net/browse/AP-4083


martin4861:
- [ ] Add comment about MODULE/WORKSPACE compatibility

.bazelignore
```diff
@@ -0,0 +1 @@
+c/third_party
```
armallen, line 1: Why is this needed?

armallen (commented):
LGTM overall, I only wonder why we need the .bazelignore change?

.bazelignore
martin4861: This avoids, that `BUILD` files in `c/third_party` are directly used in the build graph. However, those folder shall be only used as `workspace`s or `module`s. `.bazelignore` files exist also in most other repos.

MODULE.bazel
```diff
@@ -0,0 +1,44 @@
+# This MODULE file can not be used for local development inside libsbp, due to problems
```
woodfell, line 1: I'm not so hot on bazel, a lot of this stuff is still in `WORKSPACE.bazel` as well. Does it need to be in both places?

martin4861: Yes, for now we keep the `WORKSPACE.bazel` file untouched to still fully support Bazel 6.
In case Bazel 7+bzlmod is selected, then the `MODULE.bazel` + `WORKSPACE.bzlmod` files are used.

PR #1521, opened 2025-10-25
Fix "Generate artifacts" stage in CI

dgburr:
# Description

Fix the "Generate artifacts" stage in CI by updating the templates used to generate https://github.com/swift-nav/libsbp/blob/master/rust/sbp/Cargo.toml and https://github.com/swift-nav/libsbp/blob/master/rust/sbp2json/Cargo.toml so that they will produce files which are equivalent to those already committed to git (in https://github.com/swift-nav/libsbp/pull/1505).

This PR reverts ddf63632ae6de2790fc03840828f1befa47e8ef5 (which was a previous attempt to fix this issue) as well as part of 0814cbce5509c2ae26c93186da34067e2d70cabc (which "updated" the .toml files to the incorrect versions).

# API compatibility

No

## API compatibility plan

N/A

# JIRA Reference


tzilist (approved):

Madhunika-atmakuri (approved):
Thanks @dgburr 
I'm wondering about the formatting differences in rust during the release process, do you know if there is a version mismatch in what we are using during release? I don't have much knowledge about it and wondering if you can help me out on this.

pcrumley:
@dgburr i don't understand how my commit could have broken the build? I don't care if you revert it but all it does it some cosmetic formatting in the rust code reverting a previous lint fix i had written and does a minor bump of a rust dependency.

There is something fishy about how this project is being built etc vs ci. That said. whatever gets it working is a-ok by me.

dgburr:
> @dgburr i don't understand how my commit could have broken the build? I don't care if you revert it but all it does it some cosmetic formatting in the rust code reverting a previous lint fix i had written and does a minor bump of a rust dependency.
> 
> There is something fishy about how this project is being built etc vs ci. That said. whatever gets it working is a-ok by me.

@pcrumley: My previous statement was incorrect - it wasn't your commit which broke the build, but it was 0814cbce5509c2ae26c93186da34067e2d70cabc, which resulted in the .toml files and rust formatting to get overwritten with versions derived from the "bad" template.

PR #1582, opened 2026-05-13
Fix spellings to unblock libsbp release

Madhunika-atmakuri:
# Description

@swift-nav/algint-team

This PR adds two words to the spellings list to make `make docs` pass

# API compatibility

No

## API compatibility plan

# JIRA Reference

### DIFF
--- a/python/docs/source/spelling_wordlist.txt
+++ b/python/docs/source/spelling_wordlist.txt
@@ -48,6 +48,7 @@ cdc
 cn
 correctionPointSetID
 correlators
+cov
 covariances
 cpu
 dataset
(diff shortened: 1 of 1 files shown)


python/docs/source/spelling_wordlist.txt
```diff
@@ -117,6 +118,8 @@ oc
 odometry
 oe
 orthometric
+parameterization
+parameterizations
```
Madhunika-atmakuri, line 122: Adding the required spellings from newly added docs:

Image

sokhealy (approved):

PR #1584, opened 2026-06-16
Migrate to rules_rs, modernize Rust toolchain/deps, and Bazel 9.1.1

pcrumley:
Ports the Rust+Bazel setup from `rules_rust` to [`rules_rs`](https://github.com/hermeticbuild/rules_rs) and modernizes the Rust toolchain, dependencies, and Bazel version. Bottom of the `libsbp → bird_sweater → gnss-converters → starling → orion-engine → orion` rollout.

## Bazel rules
- `rules_rust` → **rules_rs 0.0.86** (llvm 0.8.3, platforms 1.1.0).
- Dependencies via `crate.from_cargo` into the existing `crates` hub, using explicit `@crates//:foo` labels (forward-compatible with a future `use_hub` switch — no `all_crate_deps()`). BUILD loads repointed to `@rules_rs//rs:*`; clippy/doc/rustfmt stay on `@rules_rust`.
- Removed `Cargo.bazel.lock`; **`Cargo.lock` is now committed** (and un-ignored) as the `from_cargo` input.
- `platform_triples`: `{x86_64,aarch64}-unknown-linux-gnu`, `{x86_64,aarch64}-apple-darwin` (matches the registered cc toolchains).

## Toolchain
- Rust edition unified to **2024**; rustc pin **1.91.1 → 1.96.0**.
- Bazel **→ 9.1.1** (`.bazeliskrc`). Bazel-9 compatibility required: `rules_go` → 0.61.1 (dev-only; 0.42.0 used the removed `CcInfo`), `rules_swiftnav` → 0.20.0, `googletest` → 1.17.0.bcr.2.

## Dependencies
- **env_logger** 0.8 → 0.11; logger configured via the builder instead of mutating `RUST_LOG` (no `unsafe set_var` under edition 2024). Drops unmaintained `atty`.
- **clap** 3 → 4 (also drops `atty`).
- **vergen** 3 → **vergen-gitcl 10**: `build.rs` rewrite; version env var is now `VERGEN_GIT_DESCRIBE` (build.rs honours a pre-set value so the hermetic Bazel build needs no git).
- **sha2** 0.8 → 0.11, **assert_cmd** 1 → 2, **serialport** 2 → 4 (example).
- Dropped unused direct `predicates`/`-core`/`-tree` dev-deps (transitive via assert_cmd).
- Generated `Cargo.toml` templates updated in lockstep.

## Deferred to a follow-up
- **swiftnav 0.10 → 0.11 and building the `link` feature.** swiftnav 0.11 removed the `ephemeris`/`navmeas` modules, so 12 of the 14 conversions in `swiftnav_conversions.rs` have no target types and can't be ported as-is — needs a dedicated migration. swiftnav stays at 0.10.

## Verification
- `cargo build --all-targets`, `cargo test`, `cargo clippy` — all green.
- Bazel **9.1.1**: `bazel build //...` and `bazel test //rust/...` (3/3 pass).

> Note: `cargo build --all-features` fails on macOS with CMake 4.x (the swiftnav-0.10 `link` C lib uses an old `cmake_minimum_required`); CI pins cmake 3.31.2 so it's unaffected. That path is the deferred, unchanged `link` code.


MODULE.bazel
```diff
@@ -33,125 +37,50 @@ register_toolchains(
     "@rules_swiftnav//cc/toolchains/llvm20/x86_64-darwin:cc-toolchain-x86_64-darwin",
 )
 
-# Rust toolchain
-bazel_dep(name = "rules_rust", version = "0.68.1")
+# Rust toolchain (rules_rs is built on top of rules_rust; it reimplements the
+# compile path + dependency resolver but still defers to rules_rust for the rules
+# it does not wrap). platforms 1.1.0 is declared above.
+bazel_dep(name = "rules_rs", version = "0.0.86")
+bazel_dep(name = "llvm", version = "0.8.3")
```
martin4861, line 42: I am unsure, whether this dependency is needed and if it has an impact downstream

martin4861 (approved):
lgtm and thanks for the cleanup! Maybe check before merging, if it can be consumed downstream.

MODULE.bazel
pcrumley: Downstream as in rust libraries/bins which consume this? Any internal libs will be bumped when pulling this in. As this is an F/OSS repo and we should be good stewards of that, we may also need to put a major revision in the SemVer.

pcrumley:
this is ready for review but pointed at a personal dev branch of a fork of `rules_rs` where I have some fixes i hope the maintainers will take. Let's wait to see how hat plays out before doing this

MODULE.bazel
pcrumley: @llvm is the hermetic C linker/cc toolchain that rules_rs uses to link Rust. It's required: we register @llvm//toolchain:all directly (line 59), and bzlmod strict-deps requires a direct bazel_dep for any repo we reference. rules_rs registers that toolchain only as a dev_dependency, so as a consumer we have to register it ourselves. Version 0.8.3 matches rules_rs's own pin.

PR #1590, opened 2026-06-24
Modernize python environment

### LINKED ISSUE #1583: Python 3.12 support
Hello, thanks for your work on this project.

It looks like this library is not supported in Python 3.12 because it pulls in `pyftdi~=0.13.4`. 

Would it be possible to use a more recent version of `pyftdi`?

Thanks

martin4861:
# Description

 - Update python environment and extend support up to python 3.14.
 - Drop unnecessary dependency `pyftdi` and just use `six`.
 - Modernize python CI workflow

# API compatibility

Yes

# Reference

Fixes https://github.com/swift-nav/libsbp/issues/1583


sbmueller (approved):
LGTM! Though I'm not too familiar with the python env used here (tox etc.). Requested @armallen as review backup.

armallen (approved):
Thanks for handling the legacy case!

PR #1595, opened 2026-07-08
Sync Cargo.lock during releases

Madhunika-atmakuri:
# Description

@swift-nav/algint-team

Since this [PR](https://github.com/swift-nav/libsbp/pull/1584) the root Cargo.lock is tracked, but the release tooling bumped rust/*/Cargo.toml without updating it, breaking make `dist-rust` and `bazel` builds of the tagged release.

- Makefile (gen-rust): run `cargo update --workspace` after generating the manifests so the lock stays in sync automatically.
- HOWTO.md: add an optional `bazel test //...` reminder before pushing.

# API compatibility

NO

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -307,6 +307,12 @@ gen-rust:
 
 	$(call announce-end,"Finished formatting Rust code")
 
+	$(call announce-begin,"Syncing Cargo.lock to generated manifest versions")
+
+	cd $(SWIFTNAV_ROOT) && cargo update --workspace
+
+	$(call announce-end,"Finished syncing Cargo.lock")
+
 gen-protobuf:
 	$(call announce-begin,"Generating Protocol Buffers bindings")
 	cd $(SWIFTNAV_ROOT)/generator; \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -307,6 +307,12 @@ gen-rust:
 
 	$(call announce-end,"Finished formatting Rust code")
 
+	$(call announce-begin,"Syncing Cargo.lock to generated manifest versions")
+
+	cd $(SWIFTNAV_ROOT) && cargo update --workspace
+
+	$(call announce-end,"Finished syncing Cargo.lock")
```
pcrumley, line 314: I think this actually updates it & any dependencies in it. you might be better off with `cargo generate-lockfile`

Madhunika-atmakuri: I had similar concern while making this change. This is what Claude had to say:
image

pcrumley: huh, ok. i was just going off of the names & not looking what they actually did. TIL.

pcrumley (approved):

howardfernandez (approved):

sokhealy (approved):

PR #1601, opened 2026-07-28
fix(lint): clippy::useless_borrows_in_formatting

jbangelo:
# Description

@swift-nav/algint-team

Since 386dee066e86aa4f3e6d8bf322fc0570bd460590 the Rust Format & Lint CI stage has failed with a new lint error. [This lint](https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#useless_borrows_in_formatting) warns about unnecessary borrows in formatting expressions, which can have a minor impact on performance but no correctness issues. This lint is new in Rust v1.97 which was released earlier in July.

This lint was only present in the test code, so the impact is relatively minor. This PR updates the Rust template to adhere to the new lint rather than ignore it.

# API compatibility

This change does not change any API.


pcrumley (approved):

martin4861 (approved):

sbmueller (approved):

sbmueller:
Would it make sense to add `Makefile` to the triggers for the `Rust` pipeline? That was the original reason the issue went undetected in the PR, and later failed all the builds on master.

jbangelo:
@sbmueller Running the Rust pipeline might have caught this earlier, but none of the changes in 386dee066e86aa4f3e6d8bf322fc0570bd460590 directly caused the breakage so changing the workflow triggers as you suggest wouldn't reliably catch similar issues. The root cause is that CI uses the latest stable build of the Rust toolchain, so even if we always built everything in CI a branch and master might use different versions depending on when the PR is built and the PR is merged.

PR #1614, opened 2026-09-16
Fix Haskell release asset upload on tags

sokhealy:
# Description

@swift-nav/algint-team

The tag-only `Publish github release` job downloads the build artifact with `actions/download-artifact` (no `name`, i.e. download all) and globs `artifacts/*/*.zip`. Since `actions/download-artifact` v5 (bumped in #1523) a download-all that resolves to exactly one artifact is extracted straight into `path` with no per-artifact subdirectory, so the zip lands at `artifacts/<name>.zip`, the glob matches nothing, and the job fails with `No files matching the glob pattern found`. It has failed on every tag since v6.3.2; the last release with a `sbp_tools_haskell-*.zip` asset is v6.3.1 (the v6.5.2 asset was uploaded by hand).

Changes:
- expose `artifact_name` as a `build` job output and download the artifact by `name`, so it always extracts directly into `artifacts/` regardless of action version or artifact count
- add a `Check artifact` step that lists the directory and fails if the zip is missing, so a layout change surfaces there instead of as an empty glob
- glob `artifacts/*.zip`
- replace the deprecated `::set-output` with `$GITHUB_OUTPUT`

Example failing run: https://github.com/swift-nav/libsbp/actions/runs/35133229724

# Testing

- This PR's `Build` job exercises the first half: the `artifact_name` job output and the `upload-artifact` step.
- The `publish` job only runs on `v*` tags, so the download, check, and release upload run for real on the next release tag. On that run, `Check artifact` should print a single `sbp_tools_haskell-<tag>-x86_64-linux-gnu.zip` under `artifacts/`, and the zip should appear under the release assets next to the `sbp_tools-<tag>-*.zip` files from the Rust workflow.
- If the layout is ever different, `Check artifact` fails with the directory listing instead of the upload step failing on an empty glob.

# API compatibility

No. CI-only change.

## API compatibility plan

N/A

# JIRA Reference

N/A


.github/workflows/haskell.yaml
```diff
@@ -72,13 +74,17 @@ jobs:
...
 
+      - name: Check artifact
+        run: ls -l artifacts && test -s artifacts/${{ needs.build.outputs.artifact_name }}
+
       - name: Upload Release
         uses: svenstaro/upload-release-action@2.11.5
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
-          file: "artifacts/*/*.zip"
+          file: "artifacts/*.zip"
```
sokhealy, line 87: This is the main fix for the issue. All of the other stuff is just nice to have

martin4861 (approved):
lgtm!

